PR #222, opened 2020-05-14
Improve Handle Properties typings

andys8:
* id is used in an example
* isValidConnection has to take the connection argument to validate


moklick:
thanks 👍

PR #298, opened 2020-06-17
README: fitView needs an argument

andys8:
## Working

```
  fitView({ padding: 10 })
  fitView({})
```

## Not working

image

## Fix

* Change readme (this PR)
* Change the type signature

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -376,7 +376,8 @@ export const storeModel: StoreModel = {
     state.isInteractive = isInteractive;
   }),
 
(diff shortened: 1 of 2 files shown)


moklick:
Hey @andys8 

thanks for the PR. I would prefer to make this parameter optional but I don't know how to do this for the [fitView action](https://github.com/wbkd/react-flow/blob/master/src/store/index.ts#L124). Can you help with this?

andys8:
I force pushed a change that seems to work. Please have a second look at it ;)

moklick:
This looks good, thanks.

PR #373, opened 2020-07-27
Bugfix: Update onConnect

### LINKED ISSUE #372: Bug: onConnect won't be updated
`onConnect` is set here:

https://github.com/wbkd/react-flow/blob/991863f2ae0a04a903baa0fe2dd3ab82924db832/src/container/GraphView/index.tsx#L145-L147

It's inside of:

https://github.com/wbkd/react-flow/blob/991863f2ae0a04a903baa0fe2dd3ab82924db832/src/container/GraphView/index.tsx#L139-L166

I think because the code is using `useEffect(() => { }, [ ])` it's only run once. `onConnect` will never be updated.

A fix could be:

```javascript
useEffect(() => { }, [onConnect])
```

or to move calling the `setOnConnect` function outside.

This issue is missing a concrete example. Let me know if you have trouble reconstructing the bug and I'll try to come up with an example.

andys8:
Will call `setOnConnect` if function passed as `onConnect` changes.

There are other ways to fix #372. Feel free to close this proposal to fix the issue and use another solution instead :)

Closes #372

### DIFF
--- a/src/container/GraphView/index.tsx
+++ b/src/container/GraphView/index.tsx
@@ -142,10 +142,6 @@ const GraphView = memo(
       updateDimensions();
       window.onresize = updateDimensions;
 
-      if (onConnect) {
-        setOnConnect(onConnect);
-      }
-
       if (rendererNode.current) {
         resizeObserver = new ResizeObserver((entries) => {
           for (let _ of entries) {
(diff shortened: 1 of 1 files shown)


moklick:
Hey could you put it into a new useEffect hook like `setSnapGrid`. Then I will merge it :)

andys8:
Sure, updated the PR

PR #381, opened 2020-07-31
feat(props): add onZoomPaneClick

namack:
First off - thanks for creating and maintaining this library. I have found it to be extremely powerful, especially how the internal state and actions are exposed.

I have a use case where I need to know when the zoom pane has been clicked. This PR would allow developers to optionally hook into this event.

Feel free to let me know if there are other ways you think this can/should be accomplished.


moklick:
Thanks for the PR :) Good idea! I renamed `onZoomPaneClick` to `onPaneClick` because we also say `paneMoveable` so it's more consistent.

moklick:
It's released in v4.9.0

namack:
Great! Thanks for the quick response!

Issue #399, opened 2020-08-04
Exposing Mouse Events


namack:
Following the lead of `onNodeMouse*` and `onNodeContextMenu`, it would be helpful to expose the `MouseEvent` to the developer when something happens on the canvas. I can see this being useful for `onElementClick`, `onNodeDrag*`, `onMove*`, and `onPaneClick`. 

All of these functions (except for `onPaneClick`) already return values, however if we were to follow the pattern set by `onNodeMouse*`, the mouse event argument would come first. Adding this argument consistent with `onNodeMouse*` would cause a breaking change in the application.

I have an immediate need for returning the mouse event for `onPaneClick`. This would not require a breaking change, but I am also willing to add the mouse events to the other actions as well. 

What would be the best way to move forward with this?
Do you agree that exposing these events would be useful to the developer?

I can open a PR that adds the event to `onPaneClick` to allow for a minor release (I should have just exposed this initially 🤦) and then another PR for the other actions, which would require a major release. Or do them all at once! 😄

moklick:
Hey @namack 

I thought about this too and there is another issue #384 regarding this one. I would prefer to do a major update with all handlers updated.

moklick:
released in v5.0.0

pitkes22:
Its abbreviation of `event` to `evt` really needed? It's subjective thing, but I think that word event is short enough to not be abbreviated and also everywhere in DOM and React API it is called `event`.

moklick:
I don't have a strong opinion about that. But you are right, React also uses `event`. I think  I will refactor it I order to use `event` instead of `evt`.

PR #400, opened 2020-08-04
Add event handlers

namack:
Addresses #399

Expose `MouseEvent` in the following functions:
- `onElementClick`
- `onNodeDragStart`
- `onNodeDragStop`
- `onConnectStart`
- `onConnectStop`
- `onPaneClick`


namack:
Looking at the source code more, I didn't think `onMoveStart` and `onMoveEnd` really needed the event to be exposed since it is more closely related to Draggable.

moklick:
Thanks a lot :) Everything looked fine and I just released v5.0.0. 🙌

PR #440, opened 2020-08-18
Always run `onElementClick` if provided, regardless of `isSelectable` or `isDraggable`

namack:
It would be useful to have the `onElementClick` event run, even if elements are not selectable or draggable.


moklick:
that's a good idea! thanks.

PR #441, opened 2020-08-18
Add `id` to EdgeProps Interface

namack:


namack:
Would it be better to instead have the Edge components extend `EdgeCompProps` instead of `EdgeProps` since that is what is passed to `EdgeComponent` in `wrapEdge.tsx`? https://github.com/wbkd/react-flow/blob/81eb7fa695eb8b94264fe178e9b3a7aae7e03c09/src/components/Edges/wrapEdge.tsx#L78-L94

moklick:
You are right. I will open an issue that we should refactor the types. There are some places where we could extend other types.

PR #443, opened 2020-08-19
Improve the way that zoom level is calculated in the fitView function

alexdawes:


moklick:
This is a very good enhancement :) Thanks!

PR #464, opened 2020-08-28
Expose getConnectedEdge + Set className of minimap node

rgoupil:
Hi,

We exposed `getConnectedEdge` as it met exactly our needs but it wasn't exported, even though it is very easy to use.

We also added a `nodeClassName?: StringFunc | string` prop to the `MiniMap` component to pass a custom class name to the nodes, with the possibility to customize it for each individual node.
In our case we needed this to display each node in the same color in the pane and in the minimap directly from the SCSS.
Before this change we could only do this through JS which forced us to duplicate the color palette in SCSS and JS.

I have to warn you that I was not able to run all tests locally, as it first complained about missing a very specific version of eslint (which I had to install manually), then that it could not find React in the compiled files, even though it was obviously present in the node_modules. I'm not sure what I did wrong, `npm i` and `npm test` did not work.
Please bear with me if I broke any tests 😅

### DIFF
--- a/src/additional-components/MiniMap/MiniMapNode.tsx
+++ b/src/additional-components/MiniMap/MiniMapNode.tsx
@@ -6,17 +6,18 @@ interface MiniMapNodeProps {
   width: number;
   height: number;
   color: string;
+  className: string;
   borderRadius: number;
   style?: CSSProperties;
 }
 
-const MiniMapNode = ({ x, y, width, height, style, color, borderRadius }: MiniMapNodeProps) => {
+const MiniMapNode = ({ x, y, width, height, style, color, className, borderRadius }: MiniMapNodeProps) => {
   const { background, backgroundColor } = style || {};
   const fill = (color || background || backgroundColor) as string;
 
   return (
     <rect
-      className="react-flow__minimap-node"
+      className={`react-flow__minimap-node ${className}`}
       x={x}
       y={y}
       rx={borderRadius}
--- a/src/additional-components/MiniMap/MiniMapNode.tsx
+++ b/src/additional-components/MiniMap/MiniMapNode.tsx
@@ -1,4 +1,5 @@
 import React, { memo, CSSProperties } from 'react';
+import cc from 'classcat';
 
 interface MiniMapNodeProps {
   x: number;
@@ -17,7 +18,7 @@ const MiniMapNode = ({ x, y, width, height, style, color, className, borderRadiu
 
   return (
     <rect
-      className={`react-flow__minimap-node ${className}`}
+      className={cc(['react-flow__minimap-node', className])}
       x={x}
       y={y}
       rx={borderRadius}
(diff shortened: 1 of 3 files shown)


src/additional-components/MiniMap/MiniMapNode.tsx
```diff
@@ -6,17 +6,18 @@ interface MiniMapNodeProps {
...
 
-const MiniMapNode = ({ x, y, width, height, style, color, borderRadius }: MiniMapNodeProps) => {
+const MiniMapNode = ({ x, y, width, height, style, color, className, borderRadius }: MiniMapNodeProps) => {
   const { background, backgroundColor } = style || {};
   const fill = (color || background || backgroundColor) as string;
 
   return (
     <rect
-      className="react-flow__minimap-node"
+      className={`react-flow__minimap-node ${className}`}
```
moklick, line 20: Could you create `classcat` to create the class name?

src/additional-components/MiniMap/index.tsx
```diff
@@ -12,6 +12,7 @@ type StringFunc = (node: Node) => string;
 
 interface MiniMapProps extends React.HTMLAttributes<SVGSVGElement> {
   nodeColor?: string | StringFunc;
+  nodeClassName?: string | StringFunc;
```
moklick, line 15: Do we really need the option to pass a function here or would it be enough to pass a string?

moklick (commented):
Did you run `npm install` inside the example folder?

src/index.ts
```diff
@@ -8,7 +8,7 @@ export { getBezierPath } from './components/Edges/BezierEdge';
 export { getSmoothStepPath } from './components/Edges/SmoothStepEdge';
 export { getMarkerEnd, getCenter as getEdgeCenter } from './components/Edges/utils';
 
-export { isNode, isEdge, removeElements, addEdge, getOutgoers } from './utils/graph';
```
moklick, line 11: Good idea 👍

src/index.ts
```diff
@@ -8,7 +8,7 @@ export { getBezierPath } from './components/Edges/BezierEdge';
...
+export { isNode, isEdge, removeElements, addEdge, getOutgoers, getConnectedEdges } from './utils/graph';
```
moklick, line 11: Good idea 👍

src/additional-components/MiniMap/index.tsx
rgoupil: We need a different class based on the node type. This can only be achieved with a StringFunc isn't ?

rgoupil:
I did not, I will give it a try, thanks !

src/additional-components/MiniMap/MiniMapNode.tsx
rgoupil: Done, I didn't know about `classcat`, thanks !

rgoupil:
Running `npm install` in the example folder did the trick, although I had to switch to firefox for the test as it couldn't run all the tests before losing connection to the browser 🤷‍♂️ All tests passed though.

src/additional-components/MiniMap/index.tsx
moklick: Yep, that's correct. Then it's fine.

moklick (approved):

Issue #480, opened 2020-09-01
onPaneScroll event handler


emielwit:
Hey,

What do you think of an `onPaneScroll` function like the `onPaneClick` to implement custom logic behind the scroll wheel? For example, you can move the canvas up and down with the `setInitTransform` with a custom function or any other logic.

Let me know how you think about this. If you want I can create a PR for this.

moklick:
Hey @emielwit 

we could also think about adding a `useScrollToPan` option as it is suggested in #359. What do you think about that?

emielwit:
I think `useScrollToPan` is a good solution. I don't think there is any other case for scrolling then zooming and panning. Do you already have a solution in mind to achieve this feature?

moklick:
Unfortunately d3-zoom (which is used for zooming and panning) doesn't support this so we need to implement it by ourselves. Maybe this issue could help https://github.com/d3/d3-zoom/issues/46.

EmielW:
I created a PR for the `onPaneScroll` i will take a look at the `useScrollToPan` if I have some time left. I think for now the `onPaneScroll` can help people to create a custom solution.

PR #481, opened 2020-09-02
feat: add onPaneScroll event handler

EmielW:
- add the possibility to listen to the onWheel event
  - add event handler to interaction demo

#480


moklick:
Thanks @EmielW. I added a note to the docs that this only works when `zoomOnScroll` is set to false.

PR #490, opened 2020-09-04
feat: add scrollToPan Handler

EmielW:
- add event handler to pan canvas when scrolling

### DIFF
--- a/example/src/Interaction/index.js
+++ b/example/src/Interaction/index.js
@@ -26,6 +26,7 @@ const InteractionFlow = () => {
   const [isDraggable, setIsDraggable] = useState(false);
   const [isConnectable, setIsConnectable] = useState(false);
   const [zoomOnScroll, setZoomOnScroll] = useState(false);
+  const [useScrollToPan, setUseScrollToPan] = useState(false);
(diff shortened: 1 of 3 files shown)


EmielW:
Hey @moklick 

I made a start on the `useScrollToPan` handler. What do you think of the solution so far? I'm not sure if the `GraphView` is the right location for the `scrollToPanHandler`. Let me know what you think and if there are any improvements I can make.

moklick:
Thanks! I would not offer the possibility to adjust the scroll speed but I think we must consider the current zoom level. Maybe this could help https://stackoverflow.com/questions/55778403/d3-zoom-pan-axis-on-scroll-pan-axis-on-drag-zoom-axis-on-pinch

EmielW:
Oke, I will take a look at that solution

moklick:
In the meantime we introduced a `translateExtent` option which gives you the possibility to set the size of the pane. The new scrolling behaviour also needs to pay attention to that. Thanks for your efforts :)

moklick:
I renamed the `master` branch to `main`. You need to open a new PR.

Issue #462, opened 2020-08-26
Cant refresh type of node


piiok:
I  have some nodes, and i change type of a element but not change.

moklick:
There is no specific reason I guess but the `type` option is not dynamic. I will look into it.

100le:
I also tried to update the **animated** option for the edges but its not reflected/re-rendered on the graph automatically.
Example: 
`
const edges = useStoreState((store) => store.edges);

if (animateEdge) {
  const matchedEdge = edges.find((value) => value.id === 'edgeId');
  matchedEdge.animated = true;
}
`

piiok:
I solved it with a component inside the ReactFlow component like this:

`
const Example= (props) => {
  const [elements, setElements] = useState([]);
  const StateUpdater = () => {
    const nodes = useStoreState(state => state.nodes);
    useEffect(() => {
      console.log(elements)
      nodes.forEach((n)=>{
         const e = elements.find((item) => ( item.id === n.id))
         if(e && e.type){
           n.type = e.type
         }
      })
    }, [elements])

    return null
  }
 return (
      <ReactFlow
        elements={elements}
      >    
        <StateUpdater/>
      </ReactFlow>
  )
}
`

I do not know if it is correct, any suggestion is welcome.

100le:
I did something similar as the example above just with the Edges as i want to dynamically control the **animated** property of the edges but it doesn't work as well..

moklick:
The `animated` property is dynamic. It's important that you update your `elements` array and pass the updated array to the ReactFlow component (like we are doing it [in this example](https://github.com/wbkd/react-flow/blob/master/example/src/Stress/index.js#L18-L34)). If you just change a reference, ReactFlow can't detect the change immediately.

The `type` property is not dynamic but we will change that. i was not aware that it could be a use case to change the type of a node.

100le:
> There is no specific reason for this behavior but the `type` option is not dynamic. I will look into it.

Can you also please enable this feature for the edges. I really need to option to dynamically change the edge `type` as well as the `node`.

moklick:
released in v5.7.1

PR #493, opened 2020-09-04
Added support for Node and Edge type updates

100le:
Fix for issue https://github.com/wbkd/react-flow/issues/462


moklick:
Thanks :)

PR #531, opened 2020-09-22
fix click after dragging

elenaclare:
When user click on node after dragging it is still not handled click. Some fixes on it.


moklick:
thanks for pointing out this bug! This fix was a bit different but it should work now.

PR #532, opened 2020-09-22
feat(control): add optional callbacks

is343:
This PR adds optional callbacks to the `Control` component for when the buttons are clicked. 
I also added the callback descriptions to the README.


moklick:
thanks!

PR #543, opened 2020-10-01
Connection validation

jasonpul:
Hello Moritz,
I started using your react-flow library recently. It's great! thanks for releasing it for people like me to use. I was playing around with the validation example you created and had a few ideas for improvements. I think this is a complete list of what my pull request changes:

- restricts connections to target-to-source or source-to-target only. source-to-source and target-to-target connection attempts are ignored. I'm not sure when there's a use case for those types of connections.

- disallows non-unique edges. addEdge now checks to see if the requested connection (source node & handle to target node & handle) already exists.

- I broke out handle-Id into it's own attribute (data-handleid). I needed to do this to implement the unique-edges-only functionality and keep the self-connection check working. Also doing this allows users to have double underscores in their node ID if they want. 

- with the handle-id mod, I also changed the connection type. Now when you implement a isValidConnection check, you will get the involved nodes and handles, not just the nodes. This will allow more complex validation schemes

### DIFF
--- a/src/utils/graph.ts
+++ b/src/utils/graph.ts
@@ -65,10 +65,14 @@ export const addEdge = (edgeParams: Edge | Connection, elements: Elements): Elem
       throw new Error(`Can't create edge. Node with id=${id} does not exist.`);
     }
   });
-  // need to check if handles exist
-  // 
-  // 
-  // 
+  
+  // make sure that the handles exists in each node
+  const handleElements = Array.from(document.getElementsByClassName("react-flow__handle")) as HTMLDivElement[]
+  [[edgeParams.source, edgeParams.sourceHandle], [edgeParams.target, edgeParams.targetHandle]].forEach(([nodeId, handleId]) => {
+    if (!handleElements.find((he) => he.getAttribute('data-nodeid') === nodeId && he.getAttribute('data-handleid') === handleId)) {
+      throw new Error(`Can't create edge. Handle with id=${handleId} does not exist within Node with id=${nodeId}.`);
+    }
+  })
 
   let edge: Edge
   if (isEdge(edgeParams)) {
--- a/src/utils/graph.ts
+++ b/src/utils/graph.ts
@@ -40,7 +40,19 @@ export const removeElements = (elementsToRemove: Elements, elements: Elements): 
   });
 };
 
-const getEdgeId = ({ source, target }: Connection): ElementId => `reactflow__edge-${source}-${target}`;
+const getEdgeId = ({ source, sourceHandle, target, targetHandle }: Connection): ElementId => `reactflow__edge-${source}${sourceHandle}-${target}${targetHandle}`;
+
+const existingConnection = (edge: Edge, elements: Elements) => {
+  for (const element of elements) {
+    if (isEdge(element)) {
+      if (element.source === edge.source && element.sourceHandle === edge.sourceHandle && element.target === edge.target && element.targetHandle === edge.targetHandle) {
+        return true
+      }
+    }
+  }
+  return false
+}
+
 
 export const addEdge = (edgeParams: Edge | Connection, elements: Elements): Elements => {
   if (!edgeParams.source || !edgeParams.target) {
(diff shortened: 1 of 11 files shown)


moklick:
This looks pretty good. Thanks for your work :) I will check it next week and let you know what I think about it.

moklick:
Hey @jasonpul 

Can you check the comments? Thanks :)

jasonpul:
hello @moklick 
Are you talking about the comments above? I've only been using GitHub for 6 months. I looked around and couldn't find any other related comments. Thanks!

moklick:
No worries :) I meant the review comments. I had some questions about your PR. I just mentioned you in the review comments. Maybe you can see them now.

src/utils/graph.ts
```diff
@@ -49,22 +61,27 @@ export const addEdge = (edgeParams: Edge | Connection, elements: Elements): Elem
...
   // make sure that there is node with the target and one with the source id
   [edgeParams.source, edgeParams.target].forEach((id) => {
-    const nodeId = id.includes('__') ? id.split('__')[0] : id;
-    if (!elements.find((e) => isNode(e) && e.id === nodeId)) {
-      throw new Error(`Can't create edge. Node with id=${nodeId} does not exist.`);
+    if (!elements.find((e) => isNode(e) && e.id === id)) {
+      throw new Error(`Can't create edge. Node with id=${id} does not exist.`);
     }
   });
+  // need to check if handles exist
```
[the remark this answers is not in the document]
jasonpul: hello @moklick , I just made the changes you requested. The fork I'm working on right now is 6.1.3 with both my pull requests integrated. Should I remove my current pull requests and submit a new one? #543 is from version 6.0.0 and #568 is from version 6.1.3 but does not have #543 changes. It's confusing and you might have a hard time merging changes.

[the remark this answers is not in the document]
jasonpul: For clarification, my current fork has:
- The handles update associated with #543 with the fixes you requested. I also modified the Validation example to showcase the usefulness of breaking out handles by itself.
- The drag and drop changes associated with #569 along with a new example to show how it can be used.

[link to my current fork](https://github.com/jasonpul/react-flow)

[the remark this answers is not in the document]
moklick: The problem right now is that #568 also contains #543 when I see it right. It would be necessary that you divide the changes so that we can merge them separately.

src/utils/graph.ts
```diff
@@ -65,10 +65,14 @@ export const addEdge = (edgeParams: Edge | Connection, elements: Elements): Elem
       throw new Error(`Can't create edge. Node with id=${id} does not exist.`);
     }
   });
-  // need to check if handles exist
-  // 
-  // 
-  // 
+  
+  // make sure that the handles exists in each node
+  const handleElements = Array.from(document.getElementsByClassName("react-flow__handle")) as HTMLDivElement[]
```
moklick, line 70: We are storing some informations about the handles in the node's `__rf` object. Couldn't we use this in order to check the handles here?

moklick (approved):

src/utils/graph.ts
jasonpul: I also thought that was the case, but upon further investigation `__rf` is not always available. `addEdge` is passed elements which can have nodes in it. But the nodes are not necessarily the node objects provided by the store. For example, in Validation, `addEdge` is passed the state array `elements` not the store array `elements`. And `__rf` is only available from nodes provided by the store. At least that's my understanding.

src/utils/graph.ts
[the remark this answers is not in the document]
jasonpul: ok i'll break them out and resubmit. Thanks!

src/utils/graph.ts
moklick: good point ;) But can't we check wether the node id exists (in the passed element array) or the node with the corresponding handleId (if one is passed via `sourceHandle ` or `targetHandle`).

jasonpul: I'm sorry I'm not following. The nodeID existence check is happening in the baseline code already. And there's not enough information being passed in by `elements` to check if the handleID exists within the nodeIDs specified. 

If there is an assumption that `addEdge` will only ever be called as a callback function by `onConnect` I guess there's no need to do the check because you're guaranteed that handleID exist within the nodeIDs specified.

moklick: Ok, I think we can remove this check. `addEdge` should only be used as a helper for the `onConnect` handler.

nadenf:
Just wondering what the outstanding issues here are ? I am happy to spend time to get this over the line.

moklick:
hey @nadenf 

I am not sure if @jasonpul still wants to split the PRs or what the status is.

jasonpul:
> hey @nadenf
> 
> I am not sure if @jasonpul still wants to split the PRs or what the status is.

Hello, sorry stuff has been keeping me busy. I'll split the pull request and submit an update by 10/22. This makes the most sense because drag and drop is already split out and in it's own pull request.

nadenf:
@jasonpul @moklick  .. Any updates on this. We can merge this one though right since it's independent of the drag/drop ?

PR #556, opened 2020-10-06
made example programmer friendlier

michaeltheobald:
it cost me an hour to fix this in my simple test. ;)


moklick:
good point :)

PR #557, opened 2020-10-06
fixed delete

michaeltheobald:
implemented key code so example does what it should


moklick:
Thanks! We will correct the wording from "delete key" to "backspace".

PR #568, opened 2020-10-10
updates to allow for drag and drop onto FlowRenderer

jasonpul:
Had to make a few changes to allow drag and dropping to work on FlowRenderer. Also added an example in the examples folder/page for it.


TheOnlyArtz:
Is there an ETA?

MadSciencist:
> Is there an ETA?

**bump**

ahmetbicer:
Hi is there an ETA? I would like to use this feature in a project. Thanks!

moklick:
Hey @jasonpul 

thanks for the PR. 

We will add an example of how you can implement a drag and drop behaviour. We will not merge this since you don't need to pass down the `onDrop` event. You can already add the event handlers by passing these event handlers to the React Flow component:

```js 
<ReactFlow
  elements={[]}
  onDrop={(event) => console.log(event)}
  onDragOver={(event) => event.preventDefault()}
/>
```

PR #622, opened 2020-10-22
Update index.js

michaeltheobald:
added height, so the example shows something :)

### DIFF
--- a/example/src/CustomNode/index.js
+++ b/example/src/CustomNode/index.js
@@ -72,7 +72,7 @@ const CustomNodeFlow = () => {
       onElementsRemove={onElementsRemove}
       onConnect={onConnect}
       onNodeDragStop={onNodeDragStop}
-      style={{ background: bgColor }}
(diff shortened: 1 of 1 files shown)


moklick:
Which browser are you using? The examples should be visible without setting a height (they already have 100% height).

michaeltheobald:
> Which browser are you using? The examples should be visible without setting a height (they already have 100% height).

Hello Moritz,

Thanks for the quick answer. I am using Chrome at Version 86.0.4240.75. 
I could send you my sample project if you want.

BTW you're doing a great job with this framework in matter of documentation and examples!

Greetings,

Michael

moklick:
The examples inside the "example" folder are meant for testing. They should work when you run `npm run dev`. 

Did you have problems because you copy pasted it and your parent container has no height?

michaeltheobald:
Yes, I copy pasted. Sry to have bothered you.

moklick:
no worries!

PR #637, opened 2020-10-27
Add elementsRemovable/removable option

nadenf:
This is to allow the ability for elements to be removed to be made an option.


moklick:
Thanks for your work but I think we won't need it. When you don't want to allow removable elements in general you just don't pass the `onElementsRemove` handler.  When you don't want a specific element to be removable you can check it in the handler like:

```javascript 
const onElementsRemove = (elementsToRemove) => {
  const removableElements =  elementsToRemove.filter(e => e.removable);
  setElements((els) => removeElements(removableElements, els));
}
```

PR #638, opened 2020-10-27
Added support for multiple handles

nadenf:
Cleanup of the work done by @jasonpul.


moklick:
thanks for your work here! We will try will find some time this week to look into it :)

PR #660, opened 2020-11-04
fix: Set border-box explicitly on Controls

### LINKED ISSUE #659: Controls sizing very small when used with a framework which sets global `box-sizing: border-box`
This is not an issue with react-flow _per-se_; however, fixing it will help with compatibility.

If a framework is used which sets `box-sizing: border-box` globally (e.g. Material UI), then the `<Controls />` end up unusably small.

See demo: https://codesandbox.io/s/immutable-breeze-0c3op 

It's an easy fix, I'll have a PR incoming shortly.

justrhysism:
Set border-box explicitly so upstream frameworks don't upset the layout.

### DIFF
--- a/src/additional-components/Controls/style.css
+++ b/src/additional-components/Controls/style.css
@@ -8,6 +8,7 @@
   &-button {
     background: #fefefe;
     border-bottom: 1px solid #eee;
+    box-sizing: content-box;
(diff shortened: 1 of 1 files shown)


moklick:
thanks :)

PR #662, opened 2020-11-04
Forward position to NodeComponent

### LINKED ISSUE #661: [Feature] Provide position to Node props
It would be useful to pass a Node's position to the rendering component.

I have seen the [example](https://reactflow.dev/examples/provider) for getting posiitonal data, and this _works_, however it's _slow_ when there's many nodes because we need to loop through the entire list on every render.

There's a few use cases for this:
- Collision detection
- Node grouping
- Change behaviour based on position (e.g. dragging a node within a "zone")
- Debugging

justrhysism:
Forward positional data to Node Component.


moklick:
good idea. Thanks for the PR!

PR #668, opened 2020-11-06
chore(edge): Output edge id with source/target warnings

justrhysism:
I've had trouble tracking down erroneously configured edges (e.g. incorrectly configured source handle). 

This small change just outputs the edge id in the warning to assist in tracking down the issue, as getting just:

> ⚠️ couldn't create edge for source handle id: null

doesn't really help in tracking down which edge I need to fix.


moklick:
Hey! Thanks :) Could you re-create the PR into the `next` branch?

justrhysism:
@moklick changes rebased onto `next` 👍🏼

PR #726, opened 2020-11-26
Feat/dragedge

moklick:


moklick:
Hey @zcmgyu 

can you have another look at it before we merge it?

zcmgyu:
I will review it on the next working day.

moklick:
We are only displaying the move cursor when you pass the `onUpdateEdge` function.

zcmgyu:
@moklick I confused when having multiple edges starting from one source. I think we should bring the selected edge to the front for easy dragging. I've tried setting z-index into the selected edge but it still doesn't work.

Screen Shot 2020-11-30 at 14 57 07

moklick:
Unfortunately there is no z-index for SVGs. We could order the edges in a way that the selected ones are on top. Do you see any problems besides that?

zcmgyu:
@moklick Besides that, I think everything is OK.

moklick:
Ok, I will keep that in mind but I think we can already merge this feature. I added a hover effect, that helps to recognize which edge updater is hovered.

zcmgyu:
@moklick Thanks <3

moklick:
Thank you for your work here @zcmgyu

moklick:
Released in v8.1.0. Example: https://reactflow.dev/examples/updatable-edge/

PR #730, opened 2020-11-29
Updated d3-zoom instance to include touchable to fix mobile interaction

balbatross:
Fixes the issue of mobile browsers not being able to interact with the background (at least on chrome dev tools)

### DIFF
--- a/src/container/ZoomPane/index.tsx
+++ b/src/container/ZoomPane/index.tsx
@@ -71,7 +71,12 @@ const ZoomPane = ({
     if (zoomPane.current) {
       const state = store.getState();
       const currentTranslateExtent = typeof translateExtent !== 'undefined' ? translateExtent : state.translateExtent;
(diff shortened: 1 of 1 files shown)


moklick:
Hey @balbatross 

we would like to keep the d3-zoom default. Your change would always register the touch event listeners. We could think about passing this as an option.

meatflavourdev:
> Hey @balbatross
> 
> we would like to keep the d3-zoom default. Your change would always register the touch event listeners. We could think about passing this as an option.

Is that to say that mobile is unsupported by default? Just curious.. haven't done much testing on mobile.

balbatross:
@meatflavourdev there's a few missing functions for mobile, I'll have some free time to make a better PR sometime in the next week

moklick:
No this only affects the dev tools mobile view. Real touch screens get recognized.

moklick:
I you make this an option I would merge it but it's not a good default. Reopen if you are still interested in this feature.

PR #813, opened 2021-01-01
[Types]: Add type inference for data property

ShaMan123:
This is very minor, but very convinient.
Not breaking (due to default type set to any `T=any`).
Now this is possible:
```tsx
const CustomNodeComponent = (props: NodeProps<{label:JSX.Element}>) => {
return (
    <div>{props.data.label}</div>    // inferred currectly
  );
}
```

### DIFF
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -34,12 +34,12 @@ export interface Box extends XYPosition {
 
 export type SnapGrid = [number, number];
 
-export interface Node {
+export interface Node<T = any> {
   id: ElementId;
   position: XYPosition;
   type?: string;
   __rf?: any;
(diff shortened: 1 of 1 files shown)


moklick:
Hey @ShaMan123 

I don't understand why the deploy fails. Everything looks good to me. Could you explain the advantage of this implementation?

ShaMan123:
I don't understand why the build fails either.
It compiled on my machine and I was able to run it.
This PR makes the `data` attribute known to typescript, better dev experience.

moklick:
Thanks! Released in v8.3.3

ShaMan123:
Thanks!

Issue #817, opened 2021-01-04
Compiling examples exhausting heap memory


Joe8Bit:
**Steps to reproduce**

1. Clone a fresh copy of `master` branch
2. `cd react-flow`
3. `npm install`
4. `cd example`
5. `npm install`
6. `npm run start`

**Outcome**

A new browser window will open (at `http://localhost:3000/`) after a few seconds with a blank page and all of the resources 404'ing. The terminal will stick at `Compiling...` for several minutes. Eventually it will bork with the following stack trace and exit with a non-zero code:

```
<--- Last few GCs --->
[44628:0x1041d9000]   451315 ms: Scavenge (reduce) 3519.7 (4100.7) -> 3518.8 (4100.7) MB, 3.7 / 0.0 ms  (average mu = 0.229, current mu = 0.262) allocation failure
[44628:0x1041d9000]   451323 ms: Scavenge (reduce) 3519.8 (4100.7) -> 3518.9 (4100.7) MB, 3.3 / 0.0 ms  (average mu = 0.229, current mu = 0.262) allocation failure
[44628:0x1041d9000]   451330 ms: Scavenge (reduce) 3519.9 (4100.7) -> 3519.0 (4101.0) MB, 3.7 / 0.0 ms  (average mu = 0.229, current mu = 0.262) allocation failure


<--- JS stacktrace --->

FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
 1: 0x100bfea43 node::Abort() (.cold.1) [/usr/local/bin/node]
 2: 0x10008609d node::FatalError(char const*, char const*) [/usr/local/bin/node]
 3: 0x100086206 node::OnFatalError(char const*, char const*) [/usr/local/bin/node]
 4: 0x10018cc95 v8::Utils::ReportOOMFailure(v8::internal::Isolate*, char const*, bool) [/usr/local/bin/node]
 5: 0x10018cc3f v8::internal::V8::FatalProcessOutOfMemory(v8::internal::Isolate*, char const*, bool) [/usr/local/bin/node]
 6: 0x1002b2ff1 v8::internal::Heap::FatalProcessOutOfMemory(char const*) [/usr/local/bin/node]
 7: 0x1002b4354 v8::internal::Heap::MarkCompactPrologue() [/usr/local/bin/node]
[...]

PR #818, opened 2021-01-04
Update example README.md

Joe8Bit:
Add note about needing to NPM install in the top-level project before running the examples. If this isn't run some deps (like react) are missing and the `npm start` command will fail.

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -3,6 +3,7 @@
 The examples are deployed here: https://reactflow.dev/
 
 ### Installation
+After running `npm install` first in the top-level project directory:
 
 ```
 npm install


moklick:
Hey @Joe8Bit 

maybe we should rename the Development section to "Development & Examples"? The information about the installation is already there: https://github.com/wbkd/react-flow#development

Joe8Bit:
Yeah, I think that might be clearer. I'll close this PR. It wasn't a tricky thing to work out, but it was a little confusing.

It might even be even simpler if you merge the `./example/package.json` into the top level `./package.json` as `devDependenices`. Then you have one install and one set of commands for new contributors.

moklick:
I like that idea :)

Joe8Bit:
I'd create a PR but it's blocked by #817 at the moment, so I can't test it's working.

Issue #577, opened 2020-10-13
Grid lines or dots not rendering


dPacc:
While moving the pane, the grid lines or dots do not render at certain points. This was working earlier in version 3.2.4, but the bug appears to be in the latest version.

![grid](https://user-images.githubusercontent.com/36117641/95840423-29d2e380-0d55-11eb-89a8-3b2dcb677179.png)

moklick:
I already tried to fix this bug. It is really strange what is going here. We are dynamically creating a svg which is used as a repeated background image. The svg only includes a `<circle />` element. I can't explain why it is not always rendered and I don't see a pattern. Maybe these are rounding errors?

Shaquu:
What is ETA to fix/merge PR it?

dPacc:
Wondering the same, waiting for it to be merged, if the fix works.

moklick:
Unfortunately the solution is not working. The PR is still open, because @AndyLnd is still working on it.

ryanwr:
I've had a go at fixing this and appears to be working for me, would be good to get opinions/confirmation.

https://github.com/wbkd/react-flow/pull/835

moklick:
Thanks @ryanwr! Works perfectly fine :) 

released in v8.3.6

PR #835, opened 2021-01-12
Use svg pattern for background instead of css tiling

ryanwr:
I had a go at changing the rendering of the svg background to use svg patterns instead of css tiling, from my testing it looks like it resolves issue #577.

I haven't looked at the performance implications of this so would be good to get your opinion. Also would be good to confirm if you no longer see the missing dots/lines issue.


moklick:
wow :) Finally this bug is solved 🙏 thanks a lot!

PR #844, opened 2021-01-18
Add a smooth transition example & fit-view optional callback

dev-mish-mash:
Hey @moklick 

I made some improvements in the smooth transitioning of fit-view

please check this PR first :) [docs#10](https://github.com/wbkd/react-flow-docs/pull/10)

The callback function parameter 'transformFunction'  added in fit-view receives calculated  x, y, and clampedZoom, then it can do whatever it wants.

I tried to make a minimal changes of core source code, as it should be

fit-view example works fine


dev-mish-mash:
image
test failed because of this

Issue #834, opened 2021-01-11
Distribute package without default styles


ryanwr:
Could you distribute a version of the package without the default styles?

We would like to customise a lot of the graph and would be good if we could exclude the default styles. Especially if using a css library such as tailwind where it is styled with css classes it becomes tricky to deal with the specificity of the default styles applying over our custom styles.

moklick:
I think the next major version will not include the default styles anymore. Maybe we can already deliver a version without styles with an environment variable for example(without breaking changes)?

ryanwr:
Should be okay to wait for the next major version if that is your plan already. I'm not familiar enough with the packaging but perhaps having a different build/import path for the styleless version could work too?

moklick:
Thats also an idea, yep.

ambroseus:
@moklick  how about  using [tsdx](https://tsdx.io/) as package builder ? ;)

moklick:
Thanks @ambroseus I will check if tsdx will work for us.

moklick:
We checked tsdx but decided to keep our rollup-only packaging setup. We want to be able to use the latest version and be as flexible as possible.  

I just pushed a new branch `feat/nocss-entry`. It works fine but it's ugly because you would need to do:

`import ReactFlow from 'react-flow-renderer/dist/nocss/ReactFlow.esm.js'`

Could anyone help me here on how to create another clean entry point? Do we need a `nocss` folder in the root folder with it's own `package.json` in order to have another `main` and `module` entry?

At the end I would like to do `import ReactFlow from 'react-flow-renderer/nocss'`.

ambroseus:
@moklick  plz review https://github.com/wbkd/react-flow/pull/846

moklick:
Thanks for your help here @ambroseus  :)

With v8.4.1 you can do:


```js
import ReactFlow from 'react-flow-renderer/nocss';

// you need these styles for React Flow to work properly
import 'react-flow-renderer/dist/style.css';

// additionally you can load the default theme
import 'react-flow-renderer/dist/theme-default.css';
```

https://reactflow.dev/docs/theming/

ambroseus:
@moklick  awesome! 🥇

PR #846, opened 2021-01-19
feat(npm): nocss package entry point

ambroseus:
Fixes Distribute package without default styles #834


moklick:
👌  thanks @ambroseus.

PR #850, opened 2021-01-20
Feat/nocss entry

moklick:


moklick:
Separate bundle without styles was released in v8.4.1 https://reactflow.dev/docs/theming/

Issue #805, opened 2020-12-27
Library breaks when trying to create an Edge from a Handle that is generated programmatically


StefanoSega:
Hi guys!
First of all, thanks for the amazing library that React Flow is.

I've a Custom Node where I need to programmatically generate the Source Handles based on a list stored in `data`; the list is named `groups` and contains strings like '1', '2', ...
`id` is the Node Id.

```
const outputHandles = groups.map((group: string, outputIdx: number) => {
    const handleId = `${id}|${outputIdx}`;
    const leftPos = `${((100 / (groups.length + 1)) * (outputIdx + 1))}%`;

    return <Handle
      type="source"
      position={Position.Bottom}
      id={handleId}
      key={handleId}
      style={{ left: leftPos, background: getGroupColor(group) }}
    />;
  });
```

The Node is generated correctly showing an handle for each group, but when I try to drag a Edge out of the Handle I get this error:

```
scheduler.development.js:178 Uncaught TypeError: Cannot read property 'find' of null
    at ConnectionLine (cjs.js:12)
    at renderWithHooks (react-dom.development.js:14803)
    at updateFunctionComponent (react-dom.development.js:17034)
    at beginWork (react-dom.development.js:18610)
    at HTMLUnknownElement.callCallback (react-dom.development.js:188)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:237)
    at invokeGuardedCallback (react-dom.development.js:292)
    at beginWork$1 (react-dom.development.js:23203)
    at performUnitOfWork (react-dom.development.js:22154)
    at workLoopSync (react-dom.development.js:22130)
    at performSyncWorkOnRoot (react-dom.development.js:21756)
    at react-dom.development.js:11089
    at unstable_runWithPriority (scheduler.development.js:653)
    at runWithPriority$1 (react-dom.development.js:11039)
    at flushSyncCallbackQueueImpl (react-dom.development.js:11084)
    at flushSyncCallbackQueue (react-dom.development.js:11072)
    at flushPassiveEffectsImpl (react-dom.development.js:22883)
    at unstable_runWithPriority (scheduler.development.js:653)
    at runWithPriority$1 (react-dom.development.js:11039)
    at flushPassiveEffects (react-dom.development.js:22820)
    at react-dom.development.js:22699
    at workLoop (scheduler.development.js:597)
    at flushWork (scheduler.development.js:552)
    at MessagePort.performWorkUntilDeadline (scheduler.development.js:164)
```

[...]

PR #859, opened 2021-01-28
Modify node wrapper to update pin state if the child list mutates, or…

burberger:
… if the subchildren mutate. Only update pin state on this occurrence as resize is already handled elsewhere.

Fixes an issue discussed on #805 where updating the ids for handles under a node would not propagate the new handle layout unless a resize event occurred.  This is handled by adding a MutationObserver to the nodes root element, and setting handleBounds each time a mutation event within the nodes element tree is observed.

### DIFF
--- a/src/components/Nodes/wrapNode.tsx
+++ b/src/components/Nodes/wrapNode.tsx
@@ -37,6 +37,7 @@ export default (NodeComponent: ComponentType<NodeComponentProps>) => {
     snapGrid,
     isDragging,
     resizeObserver,
+    mutationObserver,
   }: WrapNodeProps) => {
     const updateNodeDimensions = useStoreActions((actions) => actions.updateNodeDimensions);
     const addSelectedElements = useStoreActions((actions) => actions.addSelectedElements);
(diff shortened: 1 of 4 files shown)


moklick:
Thanks for your work here! I am a little afraid that this could have a negative impact on our performance. Maybe we could make this optional and implement a `updateNodeInternals` action that could be called when necessary. What do you think about this?

burberger:
It could be a performance issue, but the mutate observer is configured to only fire when child elements are added or removed.  It should be relatively low overhead unless the node components are being regularly mutated, it strikes me as no greater of an impact than the resize observer, which is essentially operating the same job.  There's also the option of not enabling it by default, and allowing users to pass in event filters used by the mutation observer.  

As a library user, this seemed like an obviously incorrect behavior where changing the contents of a dynamic component didn't actually change all aspects of the component, despite react properly updating the rendered object.  I don't particularly care though - if you just want to expose the update internals action fine by me, I can move the logic into my app.  Thanks for looking at the PR.

burberger:
I could also make it so the observer can be enabled per node - I only have certain types of nodes which require it, and I imagine other similar use cases behave the same.

burberger:
I did some profiling on both mobile and desktop targets and while it was acceptable on most dekstop targets, the mutation observer was way, way too expensive on mobile.  I'll rework it to be an action, thanks for the feedback.

moklick:
Ok! Maybe it could be implemented as a hook that exports a function to update a certain node/ or all nodes like the panZoomHelper hook. What do you think about that?

burberger:
Close in favor of #937

Issue #875, opened 2021-02-06
Improvement: fitView - option to exclude hidden nodes


ambroseus:
Have workaround with a lot of hidden nodes. It will be nice to have an option to exclude hidden nodes from rect bounds calculation in fitView helper
```js
fitView: (options: FitViewParams = { padding: 0.1, excludeHidden: false }) => {
          const { nodes, width, height, minZoom, maxZoom } = store.getState();

          if (!nodes.length) {
            return;
          }

          const bounds = getRectOfNodes(options.excludeHidden ? nodes.filter(node => !node.isHidden) : nodes);
```

moklick:
Released as `includeHiddenNodes` option in v8.8.0

PR #876, opened 2021-02-06
feat(fitView): added option to exclude hidden nodes

ambroseus:
implementation for #875

### DIFF
--- a/src/additional-components/Controls/index.tsx
+++ b/src/additional-components/Controls/index.tsx
@@ -10,11 +10,13 @@ import LockIcon from '../../../assets/icons/lock.svg';
 import UnlockIcon from '../../../assets/icons/unlock.svg';
 
 import useZoomPanHelper from '../../hooks/useZoomPanHelper';
+import { FitViewParams } from '../../types';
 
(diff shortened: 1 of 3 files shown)


ambroseus:
fit view examples with hidden nodes:
Selection_011
Selection_012

ambroseus:
added fitViewParams prop to Control component 
```jsx
const FlowWithControls = () => (
  <ReactFlow elements={elements}>
    <Controls fitViewParams={ excludeHidden: true } />
  </ReactFlow>
);
```

ambroseus:
@moklick  or maybe no need to add `excludeHidden` option.. just force to exclude.. what do you think? ;)

moklick:
Hey @ambroseus 

thanks for your support :) I really like the idea. What do you think about renaming it to `includeHidden` which is set to `false` by default?

ambroseus:
> renaming it to `includeHiddenNodes` which is set to `false` by default?
awesome! :)

ambroseus:
@moklick  done, please review

moklick:
Looks good! Thanks :)

moklick:
Released in v8.8.0

PR #907, opened 2021-02-15
Next 9.0.0

moklick:
Next version 9.0.0. This version doesn't use easy-peasy but redux as a state management lib. 

We decided to use redux, because easy-peasy is built on redux and we could implement this version without breaking changes. We might switch to recoil or `useSelectedContext` in the future but that is not sure. This version doesn't come with new features but it is more performant, smaller and easier to adapt for older browsers.


alevosia:
Thank you for your hard work! Great tool!

Issue #912, opened 2021-02-19
[v9.0.0]  onNodeDragStop not fired


ambroseus:
After upgrading to v9 found strange behaivour in `wrapNode`:
`isDragging` prop is always undefined, so `onNodeDragStop` handler  never call

![Selection_002](https://user-images.githubusercontent.com/380645/108513323-1b7c7180-72cb-11eb-808f-6db87fb46744.png)

ambroseus:
@moklick  I continue investigation..

ambroseus:
guess root cause is missing isDragging default `true` 
like in prev. version
```js
updateNodePosDiff: action((state, { id = null, diff = null, isDragging = true }) 
```

ambroseus:
fix:
```diff
diff --git a/src/components/Nodes/wrapNode.tsx b/src/components/Nodes/wrapNode.tsx
index a55ef38..28a695a 100644
--- a/src/components/Nodes/wrapNode.tsx
+++ b/src/components/Nodes/wrapNode.tsx
@@ -139,6 +139,7 @@ export default (NodeComponent: ComponentType<NodeComponentProps>) => {
             x: draggableData.deltaX,
             y: draggableData.deltaY,
           },
+          isDragging: true,
         });
       },
       [id, node, onNodeDrag]
diff --git a/src/components/NodesSelection/index.tsx b/src/components/NodesSelection/index.tsx
index 79cd517..0378ac6 100644
--- a/src/components/NodesSelection/index.tsx
+++ b/src/components/NodesSelection/index.tsx
@@ -87,6 +87,7 @@ export default ({
           x: data.deltaX,
           y: data.deltaY,
         },
+        isDragging: true,
       });
     },
     [onSelectionDrag, selectedNodes, updateNodePosDiff]

```

moklick:
thanks for your help :) released in v9.0.1

ambroseus:
@moklick  thanks for v9 - my bundle lost 10kb! :)

![Selection_001](https://user-images.githubusercontent.com/380645/108589915-82e8fe80-7369-11eb-8b87-4abb350660a6.png)

PR #914, opened 2021-02-19
fix(updateNodePosDiff): added missing isDragging: true in action payload

ambroseus:
fix for https://github.com/wbkd/react-flow/issues/912


moklick:
thanks a lot 💯

PR #937, opened 2021-02-24
Add a layout mutation hook for updating user blocks on change

### LINKED ISSUE #805: Library breaks when trying to create an Edge from a Handle that is generated programmatically

burberger:
This change adds an effect hook which triggers an update to the blocks pin positions every time the layout of component it is registered to is mutated.  When used within a block that mutates its own state, fixes #805.  Supersedes #859, where it was discovered that diagram global mutation observation is detrimental to performance.

### DIFF
--- a/src/hooks/useNodeLayoutEffect.ts
+++ b/src/hooks/useNodeLayoutEffect.ts
@@ -1,0 +1,13 @@
+import {MutableRefObject, useLayoutEffect} from "react";
+import {useStoreActions} from "../store/hooks";
+import {ElementId} from "../types";
+
+export default (id: ElementId, nodeElement: MutableRefObject<HTMLDivElement | null>) => {
+    const updateNodeDimensions = useStoreActions((actions) => actions.updateNodeDimensions);
+
+    useLayoutEffect(() => {
+        if (nodeElement.current) {
+            updateNodeDimensions([{id, nodeElement: nodeElement.current, forceUpdate: true}]);
+        }
+    });
+}
--- a/src/index.ts
+++ b/src/index.ts
@@ -19,6 +19,7 @@ export {
   updateEdge,
 } from './utils/graph';
 export { default as useZoomPanHelper } from './hooks/useZoomPanHelper';
+export { default as useNodeLayoutEffect } from './hooks/useNodeLayoutEffect';
 
 export * from './additional-components';
 export * from './store/hooks';


moklick:
thanks for your work! Wouldn't it be more helpful to have a function we could call if we want to trigger a re-initialization of a node? Something like:

```js 
const updateNodeDimensions = useUpdateNodeDimensions(nodeId);

// then we could call that function inside an event handler or a useEffect for example
updateNodeDimensions() 
```

It would be also helpful if we don't need to pass a ref to the hook but let the hook search for the ref vie querySelector. What do you think?

moklick:
As suggested in my comment above I added a `useUpdateNodeInternals` hook in v9.1.0. Does it solve your issue and can we close this PR?

burberger:
Yep, provided hook works fine. Closing.

PR #942, opened 2021-02-25
refactor(rollup): use babel only for cjs build

moklick:


Codesleuth:
This resolved #934 for me. Note that you can also remove `.babelrc`, it has no impact on the output.

rollup.config.js
```diff
@@ -27,6 +28,21 @@ export const baseConfig = ({ mainFile = pkg.main, moduleFile = pkg.module, injec
       format: 'cjs',
       sourcemap: true,
       exports: 'named',
+      plugins: [
+        getBabelOutputPlugin({
+          extensions: [...DEFAULT_BABEL_EXTENSIONS, 'ts', 'tsx'],
```
Codesleuth, line 33: > (!) Plugin babel: The "include", "exclude" and "extensions" options are ignored when transforming the output.

Makes no difference to the build if removed

PR #1003, opened 2021-03-15
Support react-flow inside shadow root

TheUnderScorer:
Hi!

Currently I am working on React project that is contained inside shadow root, and I've noticed that connection handlers were not working correctly, mostly due to `document.elementFromPoint` used in src/components/Handle/handler.ts because they couldn't detect elements stored inside shadow root. I've added new prop "document" that gives us ability to pass either document or shadow root that will be later used in connection handle. 

Let me know what you think! I know that it's kinda edge case, but it might be useful for other ppl as well 😄


moklick:
This is interesting. What is the use case for putting it inside shadow root?

TheUnderScorer:
In my case it's a chrome extension, thanks to putting the whole content script inside shadow root I make sure that all styles and scripts are encapsulated.

moklick:
Thanks for your work here. I think this would crash in a nextjs context for example because there is window is undefined there. Could you add some checks when using window?

TheUnderScorer:
Done!

moklick:
Looks good. One more thing: What about the key handlers? Right now they are added to the window. Is this an issue for you?

TheUnderScorer:
They've been working fine for me, so I guess that's it from my side! :)

PR #1008, opened 2021-03-15
Support element double click handler.

kid-icarus:
We have a custom use case for node double click events, wondering if you're open to supporting it. No worries if not :)

### DIFF
--- a/src/components/Nodes/wrapNode.tsx
+++ b/src/components/Nodes/wrapNode.tsx
@@ -20,6 +20,7 @@ export default (NodeComponent: ComponentType<NodeComponentProps>) => {
     onMouseMove,
     onMouseLeave,
     onContextMenu,
+    onNodeDoubleClick,
     onNodeDragStart,
     onNodeDrag,
     onNodeDragStop,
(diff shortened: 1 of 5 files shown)


moklick:
Good idea :) Thanks! Could you change it to `onNodeDoubleClick`. I don't want to combine the handlers for nodes and edges anymore. The `onElementsClick` api was a bad decision ;) When you update your PR you also need to call the `onDoubleClick` function inside the `wrapNode` component somewhere around here: https://github.com/wbkd/react-flow/blob/main/src/components/Nodes/wrapNode.tsx#L222

moklick:
Awesome! Released in v9.3.0.

kid-icarus:
Thanks @moklick! Also thanks for passing the handler down into the GraphView here: https://github.com/wbkd/react-flow/commit/afe69887e2413c211d25f31cadc4754579d976dc#diff-e32742ae7ddd6e4702c55a7c20dad28db3292a33db31b1df9bfc5672cd00ffc4R207

Much appreciated, tested and it works like a charm!

moklick:
That was the missing part ;) Thanks for your help!

PR #1046, opened 2021-03-29
refactor(handles): check if inside a shadow root

moklick:


moklick:
Hey @TheUnderScorer 

sorry for the back and forth here :D I reverted your changes because I think I found a solution that solves the problem without having the user to pass the document. For this I created a helper that returns the current host:

```ts 
export const getHostForElement = (element: HTMLElement): Document | ShadowRoot =>
  (element.getRootNode?.() as Document | ShadowRoot) || window?.document;
``` 

Can you check if this branch works for your use case?

TheUnderScorer:
It does, nice, completely forgot that something like "getRootNode" exists :D! Thanks @moklick :)

moklick:
Ok, great! Released in v9.4.1

PR #1099, opened 2021-04-15
Fixed minimap node borders on Safari, Chrome, etc

verluci:
Borders from the nodes on the minimap would clip terribly on non-Chrome browsers. This was caused by them having the css `shape-rendering: crispedges;`. Changing this to `shape-rendering: geometricPrecision;` fixes the issue on non-Chrome browsers but breaks it on Chrome.
Example:
Chrome:
Chrome
Firfox:
Firefox
Removing the CSS and instead moving the `shapeRendering` directly into the SVG-element allows us to first check if a user is using Chrome or not. If a user is using Chrome we still use `crispEdges`, else we can use `geometricPrecision`.

### DIFF
--- a/src/additional-components/MiniMap/MiniMapNode.tsx
+++ b/src/additional-components/MiniMap/MiniMapNode.tsx
@@ -9,6 +9,7 @@ interface MiniMapNodeProps {
   borderRadius: number;
   className: string;
   color: string;
+  shapeRendering: string;
   strokeColor: string;
   strokeWidth: number;
   style?: CSSProperties;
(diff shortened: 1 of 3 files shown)


moklick:
This is a great fix :) Thanks! Could you check if `window` is defined before you use it? Otherwise this would crash in a server side rendering context like next js.

verluci:
@moklick thanks :) 
I've made it so that if `typeof window === "undefined"` it defaults to `crispEdges`, does that suffice?

moklick:
that should work, thanks.

moklick:
Released in 9.5.1 :)

PR #1116, opened 2021-04-22
fix(edge): change order of the first EdgeAnchor disturbed by the EdgeComponent

clementloridan:
Capture d’écran 2021-04-22 à 10 01 01

### Problem

The first EdgeAnchor is disturbed by the EdgeComponent (the line). To demonstrate this, you can see on the screenshot, if we fill the line path and the circle of the edge anchor, the red one cannot be dragged because the line path goes under it.

### Solution

On the `wrapEdge` file, I just change the order of the svg elements. Like this anchors are always on the top.

### DIFF
--- a/src/components/Edges/wrapEdge.tsx
+++ b/src/components/Edges/wrapEdge.tsx
@@ -184,15 +184,6 @@ export default (EdgeComponent: ComponentType<EdgeProps>) => {
         onMouseMove={onEdgeMouseMove}
         onMouseLeave={onEdgeMouseLeave}
       >
-        {handleEdgeUpdate && (
-          <g
-            onMouseDown={onEdgeUpdaterSourceMouseDown}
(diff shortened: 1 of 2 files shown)


meatflavourdev:
@clementloridan 
I think it might be best to also set the following property in `src/style.css`
```css
.react-flow__edge {
  pointer-events: visibleStroke;
}
  ```
This way, the fill of the edge will not be clickable at all. This can be a huge and weird issue when edges are eg. smoothstep-- that whole right angle area of each step becomes clickable and that just doesn't make any sense...

clementloridan:
Ok, I update my PR with two updates, allow only click on the Edge's stroke with `pointer-events: visibleStroke ` Thks @meatflavourdev and add property: `pointer-events: all` of updaters.

moklick:
thanks :) released in v9.5.2.

Issue #1001, opened 2021-03-15
selectionKeyCode to null?


mikkmartin:
I want to implement box-selection without pressing additional keys. There doesen't seem to be away to start box selection without any meta/additional key. `selectionKeyCode={null}` or `selectionKeyCode="none"` would be nice.

moklick:
So in this case mouse is only used for drawing selections?

mikkmartin:
Yes, panning is done with scroll events like the UX with Facebook Origami

moklick:
Ok, this is not possible at the moment. We will see how we can make this possible.

moklick:
in v10 you will be able to pass `null` in order to remove the handler.

tajammul1996:
@moklick  - In v10, with selectionKeyCode={null}, is it possible to do a selection on drag without using any keypress?

moklick:
@tajammul1996 nope. I'll think about it..

tajammul1996:
Thank you @moklick

Issue #964, opened 2021-03-03
Feature: multiple node selection with mouse without selectionKeyCode


alexkuz:
If `paneMoveable` is disabled, it would be nice to make `selectionKeyCode` optional, so that nodes could be selected without using additional keys.

moklick:
What do you think about `selectionKeyCode="none"`? I think this is related #1001

alexkuz:
Yeah, it’s the same issue.
I don’t have a preference here, both `null` and `"none"` would be fine for me.

alxpez:
Hey guys, are there any updates/progress around this feature?

#1156 seems like a nice approach and easy enough.
Could you guys shed some light on what's stopping the merge from happening?
I'm looking forward to this little improvement and would prefer to keep using the core repo rather than branching out 😕 

However, nice job with `react-flow` awesome library, keep it up 💪

moklick:
Hey @alxpez - Thanks for the reminder. Calling a hook conditionally is not valid https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level . 

If you want to help here, please do a PR against the v10 branch. I will close #1156.

moklick:
closed in favour of #2167

PR #1156, opened 2021-05-02
Feature: poc for boolean selectionKeyCode

r2dliu:
Not necessarily the way maintainers of this library might want to do it, but I have a similar use case as #1001 and #964 and whipped up a proof of concept. 

My use case is I'm basically trying to create a toolbar where instead of holding down a key, you would click a button to activate the "box drag select" state (think photoshop toolbarish sort of thing)
 
In this situation, I needed to be able to pass true and false instead of just simply something like null. I guess I could have just made the keycode something obscure but it feels clearer to have something that specifically disables the box select when desired to be disabled.

### DIFF
--- a/src/container/FlowRenderer/index.tsx
+++ b/src/container/FlowRenderer/index.tsx
@@ -58,7 +58,9 @@ const FlowRenderer = ({
   const resetSelectedElements = useStoreActions((actions) => actions.resetSelectedElements);
   const nodesSelectionActive = useStoreState((state) => state.nodesSelectionActive);
 
-  const selectionKeyPressed = useKeyPress(selectionKeyCode);
+  const selectionKeyPressed = (typeof selectionKeyCode === 'boolean')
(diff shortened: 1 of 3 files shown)


kuldeepkeshwar:
hey, @moklick, any chance of merging it? I am using react-flow in one of our products and had a similar requirement.

alxpez:
@r2dliu pretty nice solution I wonder what's stopping the merge of this PR

moklick:
Hey @r2dliu 

Thanks for your help here! I think it would be a great addition to be able have some different kind of controls. When I am right you want to be able to drag while a certain key is pressed and the selection is the default dragging action (like in Figma for example). Unfortunately it's not valid to [call a hook conditionally](https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level). This breaks React. We would need to find another way. If you still want to contribute to this, please do a PR against the `v10` branch.

PR #1225, opened 2021-05-31
Update Controls/index.tsx: Updating ControlButton element from div to button.

itstushargarg:
Updating ControlButton element from div to button.


itstushargarg:
Hi, this is for accessibility. This change would enable focus to Controls via Keyboard Navigation

moklick:
hey @itstushargarg 

this is a good point. Why did you close the PR?

PR #1272, opened 2021-06-15
fixed smoothStepEdge behaviour for sourceX > targetX case

flamewow:
fixed smoothStepEdge behaviour for sourceX > targetX case

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,4 @@ dist
 .env
 stats.html
 .eslintcache
+.idea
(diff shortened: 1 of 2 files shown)


moklick:
The current behaviour is intended. I think the current behaviour makes it clearer where the edge starts and ends. 

your version: 
Screenshot 2021-06-16 at 11 16 19
here it's not possible to see that the edge starts on the right side of the "Source 98" node.

current version:
Screenshot 2021-06-16 at 11 17 55

flamewow:
In a situation like:
image

where I have a source on the left and target on the right (B -> C) current smoothStepEdge implementation looks ugly.

How about if I add an additional check that the target is on the right and the source is on the left?
In that case, we won't break the current version and we still going to support the new version for the case I've described?

moklick:
sounds good. I think the edge logic could benefit from a complete rewrite :D It's possible to write this way simpler and more generic.

Issue #1269, opened 2021-06-15
Feature request: Add node to selection using Ctrl+click


Ark-kun:
This is a very common mechanism for adding items to selection.

Ark-kun:
Err... How do I say it... The idea to use Win+click for multi-selection is pretty radical.
It would be interesting t hear the backstory.

moklick:
You can change it by using the `multiSelectionKeyCode` prop https://reactflow.dev/docs/api/component-props/

Ark-kun:
I'm not sure that the issue is fixed.

The default multi-selection key is unusable on most machines.

Just because there is a way to work around the issue does not mean that the issue does not exist.

Issue #1268, opened 2021-06-15
Feature request: Support deleting elements using the `Del` key


Ark-kun:
This key is widely used on the majority of machines in the world.

moklick:
You can specify the keyCode via the `deleteKeyCode` prop https://reactflow.dev/docs/api/component-props/

Ark-kun:
>You can specify the keyCode via the deleteKeyCode prop https://reactflow.dev/docs/api/component-props/

Will it work on Mac?

avichai-PhoneDo:
deleteKeyCode={['Backspace', 'Delete']}

Issue #1054, opened 2021-03-31
Improve deleteKeyCode to accept array


clementloridan:
I want to delete my elements with Backspace (keyCode: 8) and Delete (keyCode: 46)

So possibility to accept this:
`
deleteKeyCode={[8, 46]}
`

OR

`
deleteKeyCode={['Backspace', 'Delete']}
`

moklick:
Maybe it's possible to release this feature without a breaking change when we don't support multiple `keyCodes` and only make `key` and `code` available for key combinations.

clementloridan:
deprecated keyCode anymore?

I don't understand why it will be a breaking change?

If needed, I can help :)

moklick:
Right now you can pass a single `event.key` or `event.keyCode` ('Backspace' / 8) as a key code. Since using [`keyCode` is deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) and it's more complicated to build, I implemented a new version where you can pass a single `event.key` or `event.code` or multiple ones but no `event.keyCodes` anymore.

clementloridan:
Ok so what is the next step to have this on npm? Maybe on canary version?

clementloridan:
@moklick Any update? :)

moklick:
This has a very low priority. If you need it now, you can implement it with https://github.com/JohannesKlauss/react-hotkeys-hook for example.

clementloridan:
Hey @moklick I test the branch refactor/keys. For me all is good 👍

NikolMarg:
Also interested in this one. I've already done an implementation with a hook, but since I have an input inside my nodes, it was mistaking the Delete/Backspace for node deletion. Fix was easy but there are more issues (eg. when deleting multiple nodes with Shift) so I'd prefer a cleaner solution like proposed above.

Ark-kun:
I've created a PR to allow multiple keys to be specified: https://github.com/wbkd/react-flow/pull/1281

moklick:
I will close this. It's already built into the next version which can be tested with `npm install react-flow-renderer@next`

Issue #1002, opened 2021-03-15
multiple values for multiSelectionKeyCode


mikkmartin:
I want to be able to both use Shift and Cmd for multi-selection. Something like `multiSelectionKeyCode([19, 62])` would be nice. Any other way to implement it?

moklick:
This is not possible with the current version. We will put it on our list.

moklick:
Can be tested on the [refactor/keys branch](https://github.com/wbkd/react-flow/tree/refactor/keys). Unfortunately this is an ABC and will only be available in the next major release v10.

PR #1281, opened 2021-06-17
Allow specifying multiple keys for selection and deletion

Ark-kun:
Fixes https://github.com/wbkd/react-flow/issues/1269
Fixes https://github.com/wbkd/react-flow/issues/1268
Fixes https://github.com/wbkd/react-flow/issues/1054
Fixes https://github.com/wbkd/react-flow/issues/1002

### DIFF
--- a/cypress/integration/flow/basic.spec.js
+++ b/cypress/integration/flow/basic.spec.js
@@ -96,14 +96,6 @@ describe('Basic Flow Rendering', () => {
     });
   });
 
-  it('removes a node', () => {
-    cy.get('.react-flow__node').contains('Node 2').click();
-    cy.get('body').type('{backspace}');
-
-    cy.get('.react-flow__node').should('have.length', 3);
-    cy.get('.react-flow__edge').should('have.length', 1);
-  });
-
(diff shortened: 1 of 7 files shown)


saswatds:
Hi, I wanted to understand the prospects of this PR getting merged? Without this, the UX in windows and Linux environment is abysmal.

moklick:
I already worked on this feature on the `refactor/keys` branch.. Maybe it would be better to use a library for this. I would like to support key combinations like `Meta+s` and multiple possibilities like `['Del', 'Backspace']`.

In my branch I changed the API to support combinations by passing an array. I think your approach to use an array for multiple possibilities is more common so I would go for this API instead. I will see If I can find a good tiny library that fits our needs here.

Ark-kun:
> I didn't merged it because it's a breaking change

Perhaps we can keep supporting the old parameter names as well.

Ark-kun:
We've released our project and our Mac users are really confused and cannot delete the nodes =(

moklick:
This feature will be available in the next major release. You can check the API changes and installation instructions for the bete in this PR #1555.

PR #1282, opened 2021-06-17
Examples - Drag and Drop - Passing node properties when dragging from sidebar

Ark-kun:
This allows instantiating parametrized custom nodes.


moklick:
I want to keep the examples as simple as possible. This PR also only affects the lib examples not the ones on the reactflow.dev website. These are the internal examples.

PR #1344, opened 2021-07-13
feat(MiniMap): support MiniMapNode handle click

Fiv5:
see the discussion in https://github.com/wbkd/react-flow/discussions/1343


moklick:
Hey @Fiv5 

sorry for the very late response here and thanks for your work. I would prefer a bit more advanced control for the mini map. It would be nice if we could drag and zoom on the mini map. A click should always work and return the x and y of the point. I am actually not sure if we really need a node click. I think I would just use the mini map for controlling the viewport.

moklick:
Added in v11.2.0

PR #1372, opened 2021-07-20
Extend Connection Lines

RickeyWard:
Include Node and Handle in props passed to custom connection lines for advanced usage.

I created advanced label handles for a project and realized that the center of handle source tuple was not sufficient for my use case, rather than modify this behavior I could easily adjust the source position if I had access to the handle which the default connection line already finds. this change passes that power onto custom implementations. This allows me to add a `sourceX+(sourceHandle.width/2 - 5)` offset for `Position.Right` source handles. Power than I needed to satisfy application requirements by our design team.

### DIFF
--- a/src/components/ConnectionLine/index.tsx
+++ b/src/components/ConnectionLine/index.tsx
@@ -79,6 +79,8 @@ export default ({
           targetPosition={targetPosition}
           connectionLineType={connectionLineType}
           connectionLineStyle={connectionLineStyle}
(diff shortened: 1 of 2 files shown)


moklick:
Hey Ricky! Thanks for all your help here :)

moklick:
released in 9.6.6

RickeyWard:
Happy to help!

Issue #1444, opened 2021-08-14
Node drag causes React-Flow's internal state to get out of sync


datoslabs:
I encountered a problem where after node is dragged, its position becomes stale and cannot be updated via changing node.position property; the root cause of this seems to be related to the following code in the reducer:

https://github.com/wbkd/react-flow/blob/4b043bceccf41ae1edbd52e2919e1bf5a65aef99/src/store/reducer.ts#L44-L46

In my testing, the condition is never `true` and therefore allows `node.__rf.position` to get out of sync with `node.position`; I believe the correct logic should be 

```
if (storeNode.__rf.position.x !== propElement.position.x || storeNode.__rf.position.y !== propElement.position.y) { 
   updatedNode.__rf.position = propElement.position; 
 } 
```
Please note, this bug only surfaces **after** node position has been updated by node drag and subsequently updating the node's position via `node.position` will cease to have an effect on the node's position (ie `node.__rf.position` becomes out of sync with `node.position`).

jossh88:
If u are saving nodes in local state than u update same inside onNodeDragStop with new position.
Else have a look here https://github.com/wbkd/react-flow/issues/916#issuecomment-788095320

datoslabs:
Thank you @jossh88, however, this issue is different from #916.  The inconsistencies I mentioned are not between my local state and the internal react-flow state; rather, they are observed in internal react-flow state.  As a result, the API behaves inconsistently concerning node position updates via `node.position` property.  

Before node drag, programmatically updating `node.position` will move the node to the specified position; after node drag, programmatically updating `node.position` has no effect on the node position.  In other words, suppose a node has a starting position of `{x:0,y:0}`:

**Before node drag**
Updating local state's `node.position = {x:10, y:10}` will trigger the API to move the node to (10,10).  Using `useStoreState`, I can see the internal state where both `node.position` and `node.__rf.position` have been updated to `{x:10,y:10}`.

**After node drag**
Suppose we drag the node to a new position {x:50, y:50} via mouse then update local state's `node.position = {x:20, y:20}`, using `useStoreState` the internal state now shows `node.position={x:20,y:20}` and `node.__rf.position={x:50, y:50}`.  Since `NodeComponent`'s `xPos` and `yPos` properties are based on `node.__rf.position` as shown from the code excerpt below, the node is not moved to (20,20):

[...]

Issue #1425, opened 2021-08-08
Multiple Restoring after drag is not working


theashishmaurya:
Save and Restore is working fine if you save the elements and then drag things around and click on Restore everything is restored but after I change things further and then click on restore elements are not getting restored to their saved state.

for recreating this issue 
1) go to https://reactflow.dev/examples/save-and-restore/
2) Drag things around and save and again drag things around and restore. (It is working fine till then)
3) Drag things further and try restoring them to the saved state. It is not working.

tshana:
Yeah, I think I have a similar issue where the nodes are not being restored properly. I don't know if this will be helpful, but I think it might have to do with how the state is only causing React to re-render if an element is deleted or added into the flow since it's just checking the Elements array indexes and not the x/y positions of each node/edge to detect change. 

(Please delete my comment if this is not true.) I am currently trying to resolve this "issue" as well.

tshana:
I might have found a workaround. In the onRestore callback I set the elements to an empty array while waiting to get a response from the saved data. Seems like it restores properly now (haven't tested all cases yet).

```
  const onRestore = useCallback(() => {
    const restoreFlow = async () => {
      setElements([])
      const flow = await localforage.getItem(flowKey);
      
      if (flow) {
        setElements(flow.elements || []);
             ...
      }
    };
      restoreFlow();
  }, [setElements, transform]);
```

Ashumaurya:
Hello  @tshana 
I too actually think the problem is that the page is not rerendering for some reason and I think you found a good workaround to the problem, but do let me know if you find the root cause. Just in case if you wanna work with me feel free to email me.

jossh88:
https://github.com/wbkd/react-flow/issues/916#issuecomment-788095320

datoslabs:
This problem is related to #1440.  I created the following scenarios and output the `node.position` as well as `node.__rf.position` for each scenario to help demonstrate how node drag causes the API's internal states (ie `node.position` and `node.__rf.position`) to get out of sync resulting in node position not getting updated in accordance to `node.position` property.

1.  a)Drag Node 2 to far-right, b)save, c)drag Node 2 to far-left, d)restore.  This appears to be working because Node 2's `node.position` is different from internal state's `node.position`:
```
onNodeDragStart: id:2 node.position:{x:100, y:200}, node.__rf.position:{x:100, y:200}
onNodeDragStop: id:2 node.position:{x:100, y:200}, node.__rf.position:{x:709, y:242}

[...]

PR #1449, opened 2021-08-16
Ensure node.position and node.__rf.position are in sync when node position is updated via drag and node.position property updates.

datoslabs:
Fix for issues #1425 and #1444.

### DIFF
--- a/src/store/reducer.ts
+++ b/src/store/reducer.ts
@@ -41,7 +41,7 @@ export default function reactFlowReducer(state = initialState, action: ReactFlow
               ...propElement,
             };
 
-            if (storeNode.position.x !== propElement.position.x || storeNode.position.y !== propElement.position.y) {
+            if (storeNode.__rf.position.x !== propElement.position.x || storeNode.__rf.position.y !== propElement.position.y) {
               updatedNode.__rf.position = propElement.position;
             }
 
@@ -121,6 +121,7 @@ export default function reactFlowReducer(state = initialState, action: ReactFlow
         if (node.id === id) {
           return {
             ...node,
+            position,
             __rf: {
               ...node.__rf,
               position,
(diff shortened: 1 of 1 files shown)


rafaell-lycan (approved):

moklick:
Hey @datoslabs.

With your changes the states are still out of sync. The ReactFlow user state position would not match with the internal one. If you would update another node data attribute the node would jump back to the old position. This can be seen in the update-node example.

datoslabs:
@moklick, Could you please provide steps to reproduce it and I will resubmit the PR?  I have been using this fix in my dev environment without problems, maybe there's a scenario that I missed.

moklick:
Sure! Go to http://localhost:3000/update-node , drag the upper node, change the label (with the input on the upper right) - The node will jump back to its old position.

datoslabs:
Thank you @moklick, after more digging, I think I finally got it; the root cause is the use of shallow copy in `SET_ELEMENTS` reducer causing both `storeNode.position` and `propElement.position` to reference the same object instance when the user (I) also used shallow copy to update the position.  Based on this, I have made changes on my end to perform a deep clone to avoid further problems, I am not sure if you want to also make changes to the library.  On one hand, according to react best practices, HOC should ensure data objects are immutable, on the other hand, it appears that the logic in `SET_ELEMENTS` reducer intends to track element position between updates to selectively update __rf.position values which suggests the library should take steps to ensure it's not referencing the position object provided by the user.  Let me know what you think, I can make the simple fix in the reducer and resubmit the PR or you can update the documentation to clarify this behavior.

A more detailed explanation, for documentation purposes, is as follows:

The reason react-flow keeps 2 internal states of the node positions (ie `storeNode.position` vs `storeNode.__rf.position`) is to keep track of the on-screen position as well as user specified position via `setElements`.  This way if user altered the node position via node drag it can detect in the next 'setElements' update if the node position is updated programmatically by checking `if (storeNode.position.x !== propElement.position.x || storeNode.position.y !== propElement.position.y)` in the `SET_ELEMENTS` reducer thus avoiding the node jumping back to its old position after non-position updates such as label update in your scenario.  

This logic is sound except because position is a nested object and since the reducer uses a shallow copy, the position object inside the store (both `storeNode.position` as well as `storeNode.__rf.position`) is actually the same instance as the position object provided to `setElements` (ie `propElement`) after the 2nd time `SET_ELEMENTS` reducer is executed.  Because I included logic to auto layout the elements using dagre.js, when my graph is finally initialized, it would have called `SET_ELEMENTS` reducer twice.  Therefore, **before any node drag**, `propElement.position === storeNode.position === storeNode.__rf.position`; **after node drag**, `UPDATE_NODE_POS_DIFF` reducer is called and assigned a new instance to `storeNode.__rf.position` at which point `propElement.position === storeNode.position !== storeNode.__rf.position`.  Because This gives the illusion that node position is being updated by `setElements` before node drag even though `if (storeNode.position.x !== propElement.position.x || storeNode.position.y !== propElement.position.y)` always evaluates to false after being initialized.

moklick:
I will close this because this is fixed in v10. Thanks for your input and your effort here @datoslabs 🙌

PR #1512, opened 2021-09-13
Fix when it disappears when connecting an edge to update to a new node

SeonghoJin:
I am using the drag-and-drop function and Edge Update function by applying it to my project, but EdgeUpdate sometimes does not work well.

When I checked, it was not included in the dependencies of the useCallback functions that use onEdgeUpdate.

The onEdgeUpdate function code I used is   
```typescript
const [elements, setElements] = useState<Elements>(initialElements);
const onEdgeUpdate = (oldEdge: Edge, newConnection: Connection) =>
    setElements(updateEdge(oldEdge, newConnection, elements));
```

I think it's better to use the onEdgeUpdate Code from Example, but I hope this code works as well.

Below is a picture of edgeUpdate not working

Example

### DIFF
--- a/src/components/Edges/wrapEdge.tsx
+++ b/src/components/Edges/wrapEdge.tsx
@@ -160,7 +160,7 @@ export default (EdgeComponent: ComponentType<EdgeProps>) => {
           _onEdgeUpdate
         );
       },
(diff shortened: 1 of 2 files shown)


moklick:
Thanks @SeonghoJin :)

PR #1525, opened 2021-09-22
fix(connectionLine): sourceHandle with connectionMode = loose

clementloridan:
Fix `TypeError: Cannot read property 'find' of null` if the connectionMode is "loose" and all the handles are "source".
-> Crash from ConnectionLine because in `handleBounds` of a sourceNode: `target` or `source` can be null.


moklick:
Hey @clementloridan. Thanks for the PR :) Could you remove the example adjustments and the react downgrade? Then I will merge it.

moklick:
@li4man0v This is strange I don't get an error when do that.

li4man0v:
https://user-images.githubusercontent.com/5625684/139734176-599ba30c-1831-424b-a5be-b47e8075acd2.mov

This is what I see after fresh repo setup.

IOR88:
@moklick when thix fix will be merged ? we have the same problem in our app, really looking forward to thix fix :)

Issue #1314, opened 2021-06-27
Drag stop listeners aren't working if multiple nodes are moved


andys8:
### Issue

If more than one node is selected with the "click to multi select" feature (https://github.com/wbkd/react-flow/issues/609, https://github.com/wbkd/react-flow/pull/663), then there seems to be no handler that will work, if more than one node is dragged.

### Reproducing the issue

* Open the [overview example](https://github.com/wbkd/react-flow/blob/e1c34d3cd276d5acfef0f2233e83598c578dfdae/example/src/Overview/index.tsx)
* Enable or set `multiSelectionKeyCode` (default is `Meta`)
* `Meta + Click` two nodes
* Drag to change both node positions

### Observation

`onNodeDragStop(event, node)` will fire, but only contain a single node instead of a list of all nodes. 

![image](https://user-images.githubusercontent.com/13085980/123558971-a26db900-d799-11eb-9a54-e09eb43c32bb.png)

`onSelectionDragStop(event, nodes)` doesn't fire. Based on [this comment](https://github.com/wbkd/react-flow/issues/858#issuecomment-768888429) it seems to be the expected behavior, and "selection" is the blue box, but not multiple elements that are selected per click.

### Potential solutions

It might be that I'm missing something here. So, helping me out, could be a solution :)
If that's not the case:
* `onSelectionDragStop` could fire with a list of moved nodes regardless if the multiselection was done with `multiSelectionKeyCode` or with a rectangle
* `onNodeDragStop` could be called with a list of nodes instead of a single node, if more than one node was moved

### Info

* Using `react-flow` in version `9.6.1`
* Similar to https://github.com/wbkd/react-flow/issues/858
* Relevant comment https://github.com/wbkd/react-flow/issues/1006#issuecomment-799758872

pudgereyem:
Thanks for creating this issue @andys8, I have the exact same problem. Let me know if you find something out :)

andys8:
@moklick Hey, please don't feel pressured or anything, but I'd be very much interested in your thoughts about this issue. Is it likely to be tackled in the near future? How would the solution look like (referring to the options mentioned)?

It would help me a lot in being able to better decide how to move forward with this :)

moklick:
Hey @andys8 

[...]

PR #1555, opened 2021-10-05
WIP: next major release v10 

### LINKED ISSUE #425: Issue Edge Stacking Order
Edges will stack in the order they are laid out in the page HTML. However, the edge labels will also follow this order. Therefore, edge lines will be placed on top of edge labels for edges with a lower stacking order.

I think the default behavior should be that edge labels should always be placed above edge lines regardless of their order in the HTML.

I tried changing the z-index of the edge labels, but this doesn't solve the issue.

### LINKED ISSUE #1001: selectionKeyCode to null?

### LINKED ISSUE #1002: multiple values for multiSelectionKeyCode

### LINKED ISSUE #1153: Layouting Callback Hook?
Love the project! One small headache I'm dealing with is using the `fitView()` function after I update the layout of some nodes. These make their way into the react-flow via the `elements` arg. I believe you've mentioned in the past that the layouting is asynchronous, so the `fitView()` would have to be called after that for it to have any effect. 

Is there a way to hook into a callback as part of the layouting? Even whenever the flow updates would also work for my use case. I could use that to then fit the view whenever I wanted based on some component state.

### LINKED ISSUE #1353: `node.__rf.width` and `node.__rf.height` is not ready in `onLoad`.
I want to do manually layouting in `onLoad` callback. However size is not ready at that time (However DOM is ready).

It's hard to find the proper timing for first layouting.

### LINKED ISSUE #1370: Key event handling
support for combinations and multiple keys for one action.

With the new api it should be possible to pass a single value, an array for multiple possibilities and a combination of keys marked with a plus sign.

Old api:

```
deleteKeyCode="Backspace"
deleteKeyCode={8}
```

new api:

```
deleteKeyCode="Backspace" 
deleteKeyCode={['Backspace', 'Delete']}
deleteKeyCode="Meta+Delete" 
```

Related: 
#1269
#1268
#1054
#1002
#1281

### LINKED ISSUE #1373: useNode, useEdge hooks
I think it would be great, if there were a `useNode` and `useEdge` hook. Just like a `useState` hook, both should allow to update the state of a node and edge. 

A possible API could look like this:

```
const [{ data, position}, setData, setPosition] = useNode<NodeType>(id);
```

I think it would improve the developer experience and would prevent cluttering the data property of a node with an `onChange` listener, like on the front page of the documentation.

### LINKED ISSUE #1377: Nested flows
The most wanted feature is nested flows. It would be great to have it in the next major release :)

### LINKED ISSUE #1378: State management
This is a big topic. We started with easy-peasy, because we made good experience with that library. After some time we realized that it's a bit of an overhead so we dropped it. Since it's built on redux, we switched to redux. Now we would really like to get rid of the reducer, action creator boilerplate stuff and find a state management library that helps us to optimize renderings and get a better performance. Zustand, Recoil and Mobx are on our list but we are open for suggestions.

### LINKED ISSUE #1448: Pan on Space + Move
Hi,

I would like to restrict the panning to when the space key is pressed (it's common in tools to use Space + Move to pan). I thought this would be straightforward but I could not find a way to do it. Is there something like:
- a property (eg: panActivationKeyCode)?
- a way to enable / disable panning (so that I can enable it only when Space is pressed)?
- an event that can be cancelled (eg: in onMoveStart)?

Else, do you have any suggestion or workaround?

### LINKED ISSUE #1565: Feature Request: Add setCenter to OnLoadParams
`OnLoadParams` currently has `setTransform` to allow for zooming to a certain viewport. However, due to way transform handles scaling, the passed in scale distorts the `XYPosition`, requiring a bit of math to calculate the corrected coordinates to use in setTransform.

Adding `setCenter` to `OnLoadParams` will allow for zooming in on a node without the need for additional math.

### LINKED ISSUE #1611: unify API for boolean props and options
Sometimes we prefix an option with `is...` (`isDragging`) and sometimes we don't (`animated`, `selected`, .. ). I would like to unify this. Since `animated` and `selected` is already used in user land we should always omit the `is` prefix.

### LINKED ISSUE #1676: Please pass the sourceEvent to onMoveStart, onMove and onMoveEnd
d3-drag stops the even propagation at https://github.com/d3/d3-drag/blob/main/src/drag.js#L58
This makes client code cannot get on `mousedown` event.
One possible fix is to pass the sourceEvent to onMoveStart, onMove and onMoveEnd, same as other handlers.

### LINKED ISSUE #1927: The library doesn't support tree shaking
### Describe the Bug

When including the library, even just one function, the whole size of the library is added to the bundle

1. install the latest `create-react-app`
2. run a build and write down the size of the bundle (in my case 132k)
3. install the latest version of react-flow-renderer `npm i react-flow-renderer@next`
4. import one function to the `app.js` file, for example:
```js
import {addEdge} from "react-flow-renderer";
function App() {
  return (
    <div className="App">

    </div>
  );
}

export default App;
```
5. build again and compare the size of the bundle (in my case it grow to 308k)

### Expected behavior

The bundle size should grow only with the size of the imported peace of code

### Platform

- OS: Linux - Ubuntu
- Browser: Not relevant
- Version: Not relevant

moklick:
Hey! 

This is an ongoing PR where we will track the progress of the next version v10.  

### 👉 Release date is March 14th 2022 👈 

- [x] Feat: Possibility to add multiple keys or key combinations
- [x] Feat: Nested Nodes
- [x] Feat: Touch device support
- [x] Feat: Unify helper functions in `useReactFlow` hook
- [x] Feat/ Refactor: Split up `elements` into `nodes` and `edges` that are fully controlled
- [x] Feat: Add option to set markerStart for edges
- [x] Feat: Improve Marker API (set individual colors, rotation, scaling) 
- [x] Feat: Introduce `defaultNodes` and `defaultEdges` for uncontrolled usage of the component
- [x] Feat: Allow control of z-index for edges
- [x] Refactor: Replace Redux with Zustand

This new version of React Flow comes with a lot of changes (unfortunately also a lot of breaking API changes). We will use this PR to keep track of the changes and let you know how to use the new API. The API is still under heavy development and you need to expect further changes.

**Don't use this in production. The API is very likely to change. As soon as the API is stable we will remove this message and update the PR accordingly.**

If you want to try the current version you can install it via:

```sh
npm install --save react-flow-renderer@next
```

The following points are a rough overview which we will fill with more information and details over time:

# New Features

## Sub Flows

This was the most requested feature and we talked about it a lot #1024. It's now possible to create a node with child nodes by passing a `parentNode` id:

#### API

```js
const initialNodes = [
  { id: '1', data: { label: 'Node 1' }, position: { x: 50, y: 0 } },
  { id: '2', data: { label: 'Group node 2' }, position: { x: 50, y: 100 }, style: { width: 400, height: 200 } },
  // this position is relative to the position of the parent node '2' 
  { id: '2a', data: { label: 'Child node 2a' }, position: { x: 0, y: 0 }, parentNode: '2' }
];
```

All nodes have the same z-index so it's important that you always pass children after their parents (since that's also the rendering order). When a node has a parent node its position is relative to the position of the parent node. When you drag a parent all children and sub children get dragged too. You can nest your flows indefinitely. 

#### Drag Extent / Node Extent

By default you can drag a child out of its parent without any consequences. If you want a node to stay inside its parent you can use the new node option `extent='parent'`.  

#### Edges in Subflows or between different levels

By default all edges are below the nodes. In order to be able to render edges on top of a group (or just a normal node) we introduced the new `z-index` option for edges. By default all nodes have z-index: 0. When you pass `z-index:1`, the edge is on top.

## Touch Device Support

With the new version it is possible to connect nodes on touch devices. By default you can now connect nodes by clicking on handles. You can disable this feature with the new `connectOnClick` prop.

## Key Handling

You can now not only pass single keys but also multiple keys and key combinations: 

For a single key the API hasn't changed but we dropped support for the deprecated `keyCode` option.  

```jsx
// single key
const deleteKeyCode = 'Backspace';
// multiple keys 
const deleteKeyCode= ['Backspace', 'd'];
// combination of keys 
const deleteKeyCode= 'Meta+d';
// multiple combinations
const deleteKeyCode = ['Meta+d', 'ControlLeft+d'];

When you are passing an array as the deleteKeyCode, you need to define it outside of the component or memoize it. 
```

## Fit View on Init

From now on you can use the new prop `fitView` to fit the initial view. Additionally, you can pass `fitViewOptions` to configure the options passed to the initial fitView call.

```jsx
<ReactFlow
 ...
  fitView
  fitViewOptions={{ padding: 0.2 }}
/>
```

## `useReactFlow` hook

Returns a React Flow instance that exposes functions to manipulate the graph after initialization. It's the old `useZoomPanHelper` hook and the ReactFlow instance combined.

```js 
import { useReactFlow } from 'react-flow-renderer'

...
const { 
  getNodes,
  getNode,
  getEdges,
  getEdge,
  setNodes,
  setEdges,
  addNodes,
  addEdges,
  toObject,
  viewportInitialized,
  zoomIn,
  zoomOut,
  zoomTo,
  getZoom,
  setViewport,
  getViewport,
  fitView,
  setCenter,
  fitBounds,
  project,
 } = useReactFlow();
```

## `useNodes`, `useEdges` and `useViewport` hooks

Hooks for receiving nodes, edges and the viewport.

```js 
import { useNodes, useEdges, useViewport } from 'react-flow-renderer'

...
const nodes = useNodes();
const edges = useEdges();
const viewport = useViewport();
```

## Node Type: `'group'`

We added a new node type without handles that can be used for a group node.

# Breaking Changes

## ~~Elements~~ - Nodes and Edges

We saw that a lot of people struggle with the semi controlled `elements`. It was always a bit messy to sync the local user state with the internal state of React Flow. Some of you used the internal store that was never documented and always a kind of hacky solution. For the new version we offer two ways to use React Flow - uncontrolled and controlled.

### Uncontrolled `defaultNodes` and `defaultEdges`

The easiest way to get started is to use the `defaultNodes and `defaultEdges` props. When you set these props, all actions are handled internally. You don't need to add any other handlers to get a fully interactive flow with dragging nodes, removing nodes and edges and connecting nodes:

#### New API
```jsx
import ReactFlow from 'react-flow-renderer';

const defaultNodes = [
  { id: '1', data: { label: 'Node 1' }, position: { x: 250, y: 0 } },
  { id: '2', data: { label: 'Node 2' }, position: { x: 150, y: 100 } }
];

const defaultEdges = [
  { id: 'e1-2', source: '1', target: '2' }
];

const BasicFlow = () => {
  return (
    <ReactFlow
     defaultNodes={defaultNodes} 
     defaultEdges={defaultEdges}
    />
  );
};

export default BasicFlow;
```

If you want to add, remove or update a node or edge you can only do this by using the ReactFlow instance that you can receive either with the new `useReactFlow` hook or by using the `onPaneReady` handler that gets the instance as a function param.

### Controlled `nodes` and `edges`

If you want to have the full control and use the nodes and edges from local state, you can use the `nodes`, `edges` props in combination with `onNodesChange` and `onEdgesChange`. You need to implement these handlers for an interactive flow (if you are fine with just pan and zoom you don't need them). You'll receive a change when a node(s) gets initialized, dragged, selected or removed. This means that you always know the exact position and dimensions of a node or if it's selected for example. We export the helper functions `applyNodeChanges` and `applyEdgeChanges` that you can/should use to apply the changes.  

#### Old API 
```jsx
import ReactFlow, {
  removeElements,
} from 'react-flow-renderer';

const initialElements = [
  { id: '1', data: { label: 'Node 1' }, position: { x: 250, y: 0 } },
  { id: '2', data: { label: 'Node 2' }, position: { x: 150, y: 100 } },
  { id: 'e1-2', source: '1', target: '2' }
];

const BasicFlow = () => {
  const [elements, setElements] = useState(initialElements);
  const onElementsRemove = useCallback((elementsToRemove) => setElements((els) => removeElements(elementsToRemove, els)),[]);

  return (
    <ReactFlow
      elements={elements}
      onElementsRemove={onElementsRemove}
    />
  );
};

export default BasicFlow;
```

#### New API
```jsx
import ReactFlow, {
  applyNodeChanges,
  applyEdgeChanges
} from 'react-flow-renderer';

const initialNodes = [
  { id: '1', data: { label: 'Node 1' }, position: { x: 250, y: 0 } },
  { id: '2', data: { label: 'Node 2' }, position: { x: 150, y: 100 } }
];

const initialEdges = [
  { id: 'e1-2', source: '1', target: '2' }
];

const BasicFlow = () => {
  const [nodes, setNodes] = useState(initialNodes);
  const [edges, setEdges] = useState(initialEdges);

  const onNodesChange = useCallback((changes) => setNodes((ns) => applyNodeChanges(changes, ns)), []);
  const onEdgesChange = useCallback((changes) => setEdges((es) => applyEdgeChanges(changes, es)), []);

  return (
    <ReactFlow
      nodes={nodes} 
      edges={edges}
      onNodesChange={onNodesChange}
      onEdgesChange={onEdgesChange}
    />
  );
};

export default BasicFlow;
```

You can also use the new hooks `useNodesState` and `useEdgesState` for a quick start:

```js 
const [nodes, setNodes, onNodesChange] = useNodesState(initialNodes);
const [edges, setEdges, onEdgesChange] = useEdgesState(initialEdges);
```

related changes:

* `onElementsClick` ->`onNodeClick` and `onEdgeClick`
* `onElementsRemove` -> replaced by the `onNodesChange` and `onEdgesChange` handler

## ~~Redux~~  - Zustand

We switched our state management library from Redux to Zustand. With this change we removed ~300LOC from our state related code :) The internal store and actions were not documented, but a lot of you used them. If you still need access, you can use the `useStore` hook:

#### Old API
```jsx 
import { useStoreState, useStoreActions } from 'react-flow-renderer';

...

const transform = useStoreState((store) => store.transform);
const setSelectedElements = useStoreActions((actions) => actions.setSelectedElements);
```

#### New API
```jsx 
import { useStore } from 'react-flow-renderer';

...
const transform = useStore((store) => store.transform);
const setSelectedElements = useStore((store) => store.setSelectedElements);
```

You still need to wrap your component with a `ReactFlowProvider` if you want to access the internal store.

We are also exporting `useStoreApi` if you need to get the store in an eventhandler for example without triggering re-renders. 

```js
import { useStoreApi } from 'react-flow-renderer';

...

const store = useStoreApi();

...
// in an event handler
const [x, y, zoom] = store.getState().transform;
```

## ~~onLoad~~ - onInit

The `onLoad` callback has been renamed to `onInit` and now fires when the nodes are initialized.

#### Old API
```js 
const onLoad = (reactFlowInstance: OnLoadParams) => reactFlowInstance.zoomTo(2);
...
<ReactFlow
   ...
  onLoad={onLoad}
/>
```

#### New API
```js 
const onInit = (reactFlowInstance: ReactFlowInstance) => reactFlowInstance.zoomTo(2);
...
<ReactFlow
   ...
  onInit={onInit}
/>
```

## ~~paneMoveable~~ - panOnDrag

This is more consistent with the rest of the API (`panOnScroll`, `zoomOnScroll`, etc.)

## ~~useZoomPanHelper transform~~ - unified in `useReactFlow`

Since "transform" is also the variable name of the transform in the store and it's not clear that `transform` is a setter we renamed it to `setViewport`. This is also more consistent with the other functions. Also, all `useZoomPanHelper` functions have been moved to the React Flow instance that you get from the `useReactFlow` hook or the `onPaneReady` handler.

#### Old API
```js 
const { transform } = useZoomPanHelper();
...
transform({ x: 100, y: 100, zoom: 2 });
```

#### New API
```js 
const { setViewport } = useReactFlow();
...
setViewport({ x: 100, y: 100, zoom: 2 });
```

New viewport functions:

* `getZoom`
* `getViewport`

## ~~isHidden~~ - hidden

We mixed prefixed (`is...`) and non-prefixed boolean options. All node and edge options are not prefixed anymore. So it's `hidden`, `animated` and `selected`, `draggable`, `selectable` and `connectable`.

#### Old API
```js 
const hiddenNode = { id: '1', isHidden: true, position: { x: 50, y: 50 } } ; 
```

#### New API
```js 
const hiddenNode = { id: '1', hidden: true, position: { x: 50, y: 50 } } ; 
```
    
## ~~arrowHeadType~~ ~~markerEndId~~ - markerStart / markerEnd

We improved the API for customizing the markers for an edge. With the new api you are able to set individual markers at the start and the end of an edge as well as customizing them with colors, strokeWidth etc. You still have the ability to set a markerEndId but instead of using different properties, the `markerStart` and `markerEnd` property accepts either a string (id for the svg marker that you need to define yourself) or a configuration object for using the built in arrowclosed or arrow markers.

#### Old API
```js
const markerEdge = { source: '1', target: '2', arrowHeadType: 'arrow' }
```

#### New API
```js
const markerEdge = {
     source: '1',
     target: '2',
     markerStart: 'myCustomSvgMarker',
     markerEnd: { type: 'arrow', color: '#f00' }
}
```

## ~~ArrowHeadType~~ - MarkerType

This is just a wording change for making the marker api more consistent. As we are now able to set markers for the start of the edge, the name type ArrowHeadType has been renamed to MarkerType. In the future, this can also not only contain arrow shapes but others like circles, diamonds etc.

## onMoveStart, onMove and onMoveEnd handlers

Then signature has slightly changed. We are now also passing the source event:

```ts
type OnMove = (event: MouseEvent | TouchEvent, viewport: Viewport) => void;
type OnMoveStart = (event: MouseEvent | TouchEvent, viewport: Viewport) => void;
type OnMoveEnd = (event: MouseEvent | TouchEvent, viewport: Viewport) => void;
```

## nodeTypes and edgeTypes

We are not memoizing `nodeTypes` and `edgeTypes` anymore. In v9 we introduced a way to update the node/edge types via `nodeTypesId`/`edgeTypesId`. This seems unnecessary and we want the these props to behave like the other props too. This means that it is important that you are memoizing your node and edge types or simply define them outside of your component:

```js
const nodeTypes = useMemo(() => ({ custom: CustomNode }), []);
```

## Attribution

This is not really a breaking change to the API but a little change in the general appearance of React Flow. We added a tiny "React Flow" attribution to the bottom left (the position is configurable). We want to focus on the development of React Flow and for this we will add the possibility to create a paid account in order to remove the attribution. When everything is ready we will publish a blog post with further details. 

closes #1676, closes #1611, closes #1448, closes #1378, closes #1377, closes #1370, closes #1373, closes #1565, closes #1001, closes #1002, closes #425, closes #1353, closes #1153, closes #1927


GeVic:
Hey @moklick,

This PR is turning out be really helpful in my project. I appreciate your work.
Also, can you tell me how to actually make use of `nodeExtent='parent'` option? reason being it still expects `coordinates` and not `string`

moklick:
Hi @GeVic,

you need to set the `extent` option for a node https://github.com/wbkd/react-flow/blob/v10/src/types/nodes.ts#L28. I'll update the information above

rithvikvibhu:
The example code (`New API` under **Nodes and Edges**) seems to break when `nodeTypes` is passed (even if empty).
Wasn't sure if it was okay to open issues for an unreleased version, so posting it here, will create a new one if needed.

Have a basic example to reproduce here (Next.js project, but only `index.js` matters): https://stackblitz.com/edit/nextjs-vkswq1?file=pages%2Findex.js

**Without** `nodeTypes={nodeTypes}`: everything works fine.

**With** `nodeTypes={nodeTypes}` and **without** `onNodesChange={onNodesChange}`: it loads fine, but dragging nodes don't work (which is expected I guess?)

**With** `nodeTypes={nodeTypes}` and **with** `onNodesChange={onNodesChange}`: the node component doesn't load fully and if you Inspect Element, then the div `__container` and all nodes infinitely update very quickly (the element tag keeps flashing). Looks like a loop of updates - the order of the elements also changes sometimes. And of course, dragging does not work.

moklick:
Hey @rithvikvibhu thanks for your feedback! Some one already opened an issue for this #1774. We are not memoizing node types anymore and you are creating a new object on every render.

cliffordfajardo:
@moklick - I've heard great things about Zustand & I'm excited that we can use a simpler state management solution for the internals; I think this will help with people who want to dig in and contribute 😄 

I'm new to the internals of the codebase, for context, learning and curiosity:
- what does moving from Redux to Zustand bring to the library that maybe we couldn't do?

moklick:
@cliffordfajardo Our codebase is way smaller and a lot clearer. If you compare the [v9/redux store code](https://github.com/wbkd/react-flow/tree/main/src/store) with the [v10/zustand one](https://github.com/wbkd/react-flow/tree/v10/src/store) you'll see. With Zustand [transient updates](https://github.com/pmndrs/zustand#transient-updates-for-often-occuring-state-changes) are easy to implement for example. For library consumers a benefit is a smaller bundle size :)

cliffordfajardo:
@moklick - when using the new `useStoreApi` and calling `store.getState()`, `edges` exists as a property but `nodes` is represented as `nodeInternals`.


I found it quite confusing seeing the word `internals` from the `nodeInternals` property. Is it possible to have it renamed to `nodes` so it has consistency with its counterpart `edges`?

- https://codesandbox.io/s/ecstatic-gwen-2ox7z?file=/src/examples/Example1/index.tsx

cliffordfajardo:
I observed a different behaviors when upgrading from v9 to v10.
I lost the ability to drag elements on the canvas


- V9: nodes on in the canvas are draggable ✅
   - https://codesandbox.io/s/inspiring-tree-vplyu?file=/src/examples/Example1/index.tsx
- V10: nodes not draggable (same setup as v9^) ❌
   - https://codesandbox.io/s/competent-bassi-d1hjv?file=/src/examples/Example1/index.tsx

cliffordfajardo:
I observed a different behaviors when upgrading from v9 to v10.
I lost the ability to drag elements on the canvas

V9: nodes on in the canvas are draggable ✅
- https://codesandbox.io/s/competent-bassi-d1hjv?file=/src/examples/Example1/index.tsx

V10: nodes not draggable (same setup as v9^) ❌
- https://codesandbox.io/s/ecstatic-gwen-2ox7z?file=/src/examples/Example1/index.tsx

https://user-images.githubusercontent.com/6743796/147396416-7fb9fece-9385-470f-ae64-dc13cf3a9b6a.mp4

rithvikvibhu:
@cliffordfajardo that's a breaking change mentioned under **Nodes and Edges**. Check out how `onNodesChange` (and edges) is used under **New API**. Has come across the same issue and then realised :p

moklick:
hey @cliffordfajardo 

as @rithvikvibhu said you need to implement the `onNodesChange` and `onEdgesChange` handler. 
With the new structure it should be obsolete to get the `nodes` object from the store because you already have the up to date nodes in your own state. Why do you need the nodes from the store?

cliffordfajardo:
Hi @moklick,
> With the new structure it should be obsolete to get the nodes object from the store because you already have the up to date nodes in your own state

I suppose I can store the node & edge data in a context and then I wouldn't need to grab them from the store & that will solve my use case

<br />

> With the new structure it should be obsolete to get the nodes object from the store because you already have the up to date nodes in your own state

Can the same be said about edges? Thus why expose `edges` & `nodes` to users or is the `edges` case different?
If the intent is we should store the state, would it be a good idea to make these private & not expose them to the user?

<br />

> Why do you need the nodes from the store?


Some background context to help explain my journey 😅. Currently I am building something that looks like this:

CleanShot 2021-12-26 at 05 53 10

- there is a 1 CentralDevice  (in the middle)
    - a central node will have N number of handles (_exactly the same number of handles as there are connectedDevices_)
- there multiple `ConnectedDevice`'s
   - a connected device will have 1 handle 
   - a connected device handle will connect to of the handles the `CentralDevice` has


<br />

What the `CentralDevice` and `ConnectedDevice` look like:

<details>
<summary> Code Snippet: `CentralNode` and `ConnectedNode` </summary>

**CentralDevice**
```tsx
const CentralDevice: React.FC<NodeProps> = ({ data, isConnectable, id, type }) => {
  const store = useStoreApi();
  const connectedDevices = Array.from(store.getState().nodeInternals.values()).filter((d: any) => d.type === 'connectedDevice');
  const centralDeviceID = id;

  return (
    <>
      <div className={styles.CentralDevice}>
        <div>{data.Asn}</div>
      </div>

      {connectedDevices.map((connectedDevice, i) => {
        return (
          <React.Fragment key={connectedDevice.id}>
            <Handle
              id={`source=${centralDeviceID} target=${connectedDevice.id}`}
              type="source"
              position={Position.Top}
              style={{ ...baseHandleStyle, left: `${i * 25}px` }}
            />
          </React.Fragment>
        );
      })}
    </>
  );
};
```

**ConnectedNode:** 
```tsx
const ConnectedNode: React.FC<NodeProps> = ({ data, id }) => {
  const { contextValue } = useDeviceConnectionsViewer();
  const PORT_VALUE = getPortValue(contextValue.activePortDisplayOption, data);
  const deviceID = id;

  return (
    <>
      <div data-id="connected_device" className={styles.ConnectedDevice}>
        <div className={styles.ConnectedDevice__Name}>{data.RemoteDevice.Asn}</div>
        <div className={styles.ConnectedDevice__Hostname}>{data.RemoteDevice.Hostname}</div>
        <div className={styles.ConnectedDevice__PortEntry}>{PORT_VALUE}</div>
      </div>

      <Handle id={`${deviceID}`} type="target" position={Position.Bottom} style={handleStyle} />
    </>
  );
};
```
</details>


CleanShot 2021-12-26 at 06 06 58@2x



The edge nodes that make the connection between the `centralDevice` and connected node look like this:

```ts
const edge = {
 source: `${centralDevice.id}`,
 sourceHandle: `source=${centralDevice?.id} target=${connectedDevice.id},
  target: connectedDevice.id,
}
```

Why did I need access to the nodes from the store?

```
```

- I needed access to the `nodes` in my `CentralDevice` custom node component in order to:
   -  create  a  unique`<Handle />` component for each `ConnectedDevice` inside the `CentralDevice` custom node component

moklick:
@cliffordfajardo 

I understand that it's convenient to use the React Flow store as you said with `useStoreApi` for example. When your app gets bigger I would recommend to add some state management and then you wouldn't need to access the React Flow store directly. This makes it also easier to trigger node updates on connect. For clarification: We store the edges as they are but we do some extra stuff with the nodes (adding handle bounds) for example. For the internal processes it's better for us to use a `Map` (because we often need to find a node by id). That's why we renamed it to `nodeInternals`. However if you want to access the nodes, it's fine to use `nodeInternals`. It's just a `Map` instead of an `Array` with some additional information. 

To your code: You should not use `useStoreApi` inside your render function. That's for accessing the store inside event handlers or transient updates for example. `useStore` in combination with a selector is what you are looking for: https://github.com/pmndrs/zustand#selecting-multiple-state-slices  

To your questions:
1. Just use `nodeInternals` + `Array.from(nodeInternals)`
2. Sorry but I think this is not the right place to discuss this ;)

AndrewJSchoen:
I have a question related to node movement as well, and I am not sure based on reading this thread that my use-case was covered.  Basically, I have an existing zustand store that contains a map of the data that is rendered, and some of that store gets converted into the nodes that are visualized in the flow diagram. This zustand store has a bunch of other stuff for the application in it, so I can't just rely on the zustand store from react-flow. A toy example is here: 

https://codesandbox.io/s/intelligent-shamir-jogr0?file=/src/Node.jsx

Basically, the store is in charge of handling a bunch of other stuff, and in my non-toy project, contains a bunch of nested data that gets rendered within the nodes themselves. In any case, if I just try to hook up `onNodesChange` to a handler function that makes changes to my own zustand store, I get an error: `Unable to find node on an unmounted component.` Any ideas on how to use the new API with such a setup?

moklick:
@AndrewJSchoen the issue is that you are doing `nodeTypes={{ block: Node }}` if you are define the node types outside of the component it works as expected: https://codesandbox.io/s/react-flow-renderer-react-dnd-forked-g82ue?file=/src/Canvas.jsx:212-247 . With your implementation you create a new `nodeTypes` object on every render.

FateRiddle:
Hi, don't know if It's rude to reply here(if so delete it), I'm about to write a project using react-flow, and see the breaking change is coming. 
Just wondering when do you plan for the release? A matter of days or still count in weeks?

moklick:
@FateRiddle not rude at all ;) hopefully by the end of January but no later than the end of February.

pgupta56:
Hi @moklick  when it will be available to use in production projects ?? Thanks!!

moklick:
@pgupta56 please see the comment above.

bluebill1049:
Thank you guys for this amazing new major release!

moklick:
Hey @bluebill1049 

this is indeed the biggest change. I am thinking about adding `defaultNodes` and `defaultEdges`. This would restore the v9 behaviour somehow. You can't omit the drag updates.

With v10 it's not recommended to pass down nodes or edges through components. I would use some kind of state management library or React context if you need the nodes and edges at a higher place in your component tree.

bluebill1049:
Hey @moklick 

Thanks for the reply! First of all, I love all those new features coming from the v10 update and am a huge fan of your guys' work. I am currently finding a way to work around the above use case and upgrading to v10 (beta). I certainly don't want it to skew the general direction this library is heading to and add more API/work load on your guys. I think overall the API is a huge step forward. Component wrapped with `<ReactFlow />` is going to be a more isolated component for containing all updates within the React Flow. 

What I am trying currently is 

- `useimperativehandle` which forwards some of the methods back to the parent component 
- building my own `Provider` to hoist some of the utility methods.

**Side note:** I have also played around with `useStoreAPI` and `useStore` to set nodes or edges, but they don't really act the changes for state update which I supplied (incorrect usage). 

Overall, I think `add`, `remove` or manipulate `edges` or `nodes` is one thing which I find a bit more difficult than before due to the architecture change. But this is specific to my use case which is not ideal to re-render at the parent component, Would be nice if we can leverage `useStoreAPI` within the Provider, so we can still manipulate with the `edges` and `nodes`. 

Thanks

Himself65:
I'm trying to implement a function that allows editing the view on a specific node. 

and I'm thinking to code like this
```ts
setNodes(ns => applyNodeChanges([{
  id: 'xxx',
  type: 'update',
  data: {
    label: 'new label'
  }
}], ns))
```

but the current workaround is

```ts
setNodes(ns => {
  const index = ns.findIndex((n) => n.id === menuStore.node?.id)
  const data = { ... }
  const newNodes = [...ns]
  newNodes[index].data = data
  return newNodes
})
```

but the problem is the size of nodes would be like thousands os

Is there another solution for this?

Himself65:
I see I can use `store.nodeInternals` to access the node using id. sorry for the stupid question

Himself65:
I'm trying to use `store.setNodes` to restore nodes, but the nodes immediately disappear after calling the function.

It seems like the `item` parameter is still empty, but the `changes` is not, then it returns `[].`

Kapaacius:
Hey @moklick 

First of all thank you for this library and for v10 with parent/child node relations.
I was wondering if there is a simple way to know if node is connected. I want to apply a specific style to nodes that have a connection.

moklick:
@Himself65 the `applyNodeChanges` helper is for applying changes that were triggered by React Flow internally. If you want to update a node programatically it's the same as before:

```js 
setNodes((nds) =>
  nds.map((node) => {
    if (node.id === clickedID) {
      node.data = yourUpdatedData;
    }

    return node;
  })
);
```

@Kapaacius 
You could use the [`getConnectedEdges`](https://reactflow.dev/docs/api/helper-functions/) helper function for this.

moklick:
@bluebill1049 

I think we will add `defaultNodes` and `defaultEdges`. This would make it really easy to get started and would be a great improvement in general. In your case you would probably use the controlled one but for that type we are missing a comfortable way to update the nodes and edges. An idea would be to add an `useUpdateNodes`/`useUpdateEdges` hook with an API like this:

Uncontrolled Flow (wrapped with a ReactFlowProvider): 
```js
const updateNodes = useUpdateNodes();
...
const onClick = () => {
  updateNodes((nodes, setNodes) => {
    setNodes(nodes.map((node) => {
      node.position = { x: 0, y: 0 };
      return node;
    }))
  }
}
...
return (
  <ReactFlow
    defaultNodes={defaultNodes}
    defaultEdges={defaultEdges}
```
What do you think about this?

bluebill1049:
That's awesome @moklick I think this would give users the freedom of choice on their application state update. `useUpdateNodes` / `useUpdateEdges` would be a great opt-in option for them to consume the state update. This is similar to what I have done with `react-hook-form`'s `useFormState` which users can benefit from with isolated component re-render when their applicant requires to.

moklick:
@bluebill1049 

We've added `defaultNodes` and `defaultEdges`. This is a great new feature for the next release :) You get a fully interactive flow only by defining these two props. You can alter the nodes by using the new `useReactFlow` hook that comes with `setNodes`, `addNodes` and `removeNodes` helper functions. An example can be found here: https://github.com/wbkd/react-flow/blob/v10/example/src/DefaultNodes/index.tsx#L27-L43

bluebill1049:
That's amazing! Thank you guys for looking into this. ❤️

alessandropisu:
Hi @moklick, thanks for the update! With this new uncontrolled approach how is it possibile to track nodes/edges deletion if I can't use callback methods (onNodesChanges, onEdgesChange)? I noticed also that useNodes and useEdges hooks are not exported and importable. Thanks again and amazing work :)

BrianHung:
@moklick 

Is it possible to introduce another breaking change to fix https://github.com/wbkd/react-flow/issues/1314? Specifically, change the following callback parameters from `node` to `nodes` to correctly pass back an 

https://github.com/wbkd/react-flow/blob/6de04df3396338b5a85fc189d630761f3ab6994c/src/container/NodeRenderer/index.tsx#L17-L19

Figma's selection api uses array if that's any confidence :)

https://www.figma.com/plugin-docs/api/properties/PageNode-selection/
https://www.figma.com/plugin-docs/accessing-document/#getting-the-current-selection

thadk:
I think `onPaneReady` may have been renamed to `onInit` (references in the docs above in case you're thinking of repurposing them or in case other people are already doing implementations too)

moklick:
Hey @BrianHung good point ;) I will look into it! 
@thadk yep, we renamed it to `onInit`.

sachaw:
Love the improvements, however I've seen a missive bundle size bump. Is this expected or can it be improved?

moklick:
Hey @sachaw the bundle size should be relatively similar. What are your measurements? However we have plans to reduce the bundle size but not in this release.

sachaw:
Thanks for the reply, I'll have to do a deeper look,  currently using Vite (so rebuild or roll-up). 

My app gets built, gzipped then tar'd. After installing 10.x the tar size jumped 100kb or so.
I'll double check there's not an issue on my end.

sachaw:
Just checked pnpm had cached both versions, I measured the increase at 1.2kb from v9-v10.
I would however try to allow tree-shaking/code splitting. the footprint is fairly monolithic.
Here's a breakdown of my application to put tit in perspective 
image

moklick:
@sachaw thanks for the hint. React Flow should be treeshakable.. I will try to find out what is going on here

xiyangjun:
Why not provide the setNode method?

pgupta56:
any idea when we can V10 use this in our production project?  when the PR will be merged and we can use officially?

moklick:
By end of Feb / beginning of March

megashot:
Hey @moklick is there going to be performance comparison? Because we run a heavily modified version and do not know if you should update to v10 (and apply our modifications on top)?

antoniobologna:
Just found out about this project! Was going to install it in our production env but I guess we'll wait for v10 🥳

alessandropisu:
I think it would great to have, at least in uncontrolled mode, some method to catch when a user delete a nodes/edges (like onNodesDelete/onEdgesDelete) to distinguish from the others changes (position, selection etc...). At the moment is pretty hard to track the changes in uncontrolled mode to make some business logic at some specific events :(

moklick:
@alessandropisu thanks for the feedback. I thought about this too.. we will add these handlers!

Sec-ant:
There seems to be bugs in multiple keys deletion. Except for the first node, one has to press `delete` twice to delete of the rest nodes when using arrays to bind multiple deleteion key codes.

A minimum example:

[Edit recursing-brown-d02xxw](https://codesandbox.io/s/recursing-brown-d02xxw?fontsize=14&hidenavigation=1&theme=dark)

A work around for this is to memoize the deletion key code:

```javascript
const deleteKeyCode = useMemo(() => ["Delete"], []);
```

which is also written in the example above, you can uncomment it to see the difference.

Hopefully this can be fixed, thank you!

OrhanTozan:
zoomOnPinch={false} and zoomOnDoubleClick={false} doesn't seem to work on mobile in v10. Not sure if it is the same in v9.

moklick:
@Sec-ant thanks for the hint! I updated the example above. When you pass arrays or object to RF, you always need to memoize it or define it outside of the component in order to prevent unnecessary re-renderings.

bimal72:
Hi,
I want to implement React Flow in one of my projects. Can I start with v10? or start with v9 and then migrate to v10?
When v10 release is expected?

OrhanTozan:
> `zoomOnPinch={false}` and `zoomOnDoubleClick={false}` doesn't seem to work on mobile in v10. (Didn't test it with v9).

Made a ticket for this #1914

Himself65:
> @alessandropisu thanks for the feedback. I thought about this too.. we will add these handlers!
> 
> UPDATE: I added `onNodesDelete` and `onEdgesDelete` handlers.

I have a question, why don't use `store.onNodesChange([{ id: node.id, type: 'remove' }])` to remove the nodes/edges

moklick:
@Himself65 it's just more convenient like the `onNodeDrag` or `onSelectionChange` handlers. You can use these handlers without a deeper knowledge of the `onNodesChange` handler.

src/types/general.ts
```diff
@@ -20,6 +20,10 @@ export type OnNodesChange = (nodes: NodeChange[]) => void;
 
 export type OnEdgesChange = (nodes: EdgeChange[]) => void;
 
+export type OnNodesDelete = (nodes: Node[]) => void;
+
+export type OnEdgesDelete = (nodes: Node[]) => void;
```
Sec-ant, line 25: A small typo, `nodes: Node[]` -> `edges: Edge[]`

moklick: thanks!

Himself65:
I find that there's a breaking change. There will be no more re-render for each node when you connect to nodes. How could I force update these two nodes?

kafkahw:
@moklick thanks for your great work on this library! I'm exploring graph libraries for one of our internal tools and find `react-flow` very impressive. 
I created an example with v9 that's integrated with `dagre` for custom layout. It works pretty well (see the example [here](https://stackblitz.com/edit/react-whyxhu?file=src%2FApp.js)). However, after upgrading to v10, I created a similar example [here](https://stackblitz.com/edit/react-ts-qfwtky?file=CustomNode.js) but found that the dragging feature is not working. The only difference btw these two examples is that I split `elements` into `nodes` and `edges` due to the new API changes. The rest code should be pretty much the same.  It might be something related to the new API `onNodesChange` and `onEdgesChange`, but I can't figure out the exact reason. Do you have ideas on this?

moklick:
Hey @kafkahw! The issue is that you are overwriting the nodes props on every render. You only should do the layouting once:  

Instead of:

```js
const [nodes, setNodes, onNodesChange] = useNodesState(initialItems.nodes);
...
<ReactFlow nodes={getLayoutedNodes(nodes)} />
```
 
you need to do:

```js
const [nodes, setNodes, onNodesChange] = useNodesState(getLayoutedNodes(initialItems.nodes, initialItems.edges));
...
<ReactFlow nodes={nodes} />
```

moklick:
@Himself65 Good point. In my view it was a buggy behaviour in v9 but it's a breaking change.. I would create a hook for the custom node and check if the node has a connection. Something like (not tested):

```js
import { useCallback } from 'react';
import { useStore, ReactFlowState } from 'react-flow-renderer';

function useIsConnected(nodeId: string): boolean {
  const isConnected = useStore(
    useCallback((state: ReactFlowState) => state.edges.some(edge => edge.target === nodeID || edge.source === nodeId), [nodeId])
  );

  return isConnected;
}

**Comment by kafkahw** [comment]
> Hey @kafkahw! The issue is that you are overwriting the nodes props on every render. You only should do the layouting only once:
> 
> Instead of:
> 
> ```js
> const [nodes, setNodes, onNodesChange] = useNodesState(initialItems.nodes);
> ...
> <ReactFlow nodes={getLayoutedNodes(nodes)} />
> ```
> 
> you need to do:
> 
> ```js
> const [nodes, setNodes, onNodesChange] = useNodesState(getLayoutedNodes(initialItems.nodes, initialItems.edges));
> ...
> <ReactFlow nodes={nodes} />
> ```

@moklick thanks for your quick reply! Yeah, I understand that if we call the layout calculation once, it will work in v10. However, in our app we're actually trying to add nodes dynamically. Based on your suggestion [here](https://github.com/wbkd/react-flow/issues/1113#issuecomment-829461998), it seems that we should re-calculate the layout whenever a new node is added. That's why I implemented this way in v9 and it works like a charm. I'm wondering if you can shed a light on how to make it work in v10 as well.

**Comment by Himself65** [comment]
```bash
react_devtools_backend.js:4061 Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of DraggableCore which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://reactjs.org/link/strict-mode-find-node
    at div
    at DraggableCore (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:7025:5)
    at NodeWrapper (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:10409:19)
    at div
    at NodeRenderer (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:8170:19)
    at div
    at Viewport (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:9690:23)
    at div
    at ZoomPane (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:4628:21)
    at FlowRenderer (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:8048:23)
    at GraphView (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:9962:24)
    at Wrapper (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:10746:23)
    at div
    at eval (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:10838:20)
    at div
 ...
```

Himself65:
> ```shell
> react_devtools_backend.js:4061 Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of DraggableCore which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://reactjs.org/link/strict-mode-find-node
>     at div
>     at DraggableCore (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:7025:5)
>     at NodeWrapper (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:10409:19)
>     at div
>     at NodeRenderer (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:8170:19)
>     at div
>     at Viewport (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:9690:23)
>     at div
>     at ZoomPane (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:4628:21)
>     at FlowRenderer (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:8048:23)
>     at GraphView (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:9962:24)
>     at Wrapper (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:10746:23)
>     at div
>     at eval (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:10838:20)
>     at div
>  ...
> ```

solution: https://stackoverflow.com/questions/63603902/finddomnode-is-deprecated-in-strictmode-finddomnode-was-passed-an-instance-of-d

Himself65:
> ```shell
> react_devtools_backend.js:4061 Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of DraggableCore which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://reactjs.org/link/strict-mode-find-node
>     at div
>     at DraggableCore (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:7025:5)
>     at NodeWrapper (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:10409:19)
>     at div
>     at NodeRenderer (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:8170:19)
>     at div
>     at Viewport (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:9690:23)
>     at div
>     at ZoomPane (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:4628:21)
>     at FlowRenderer (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:8048:23)
>     at GraphView (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:9962:24)
>     at Wrapper (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:10746:23)
>     at div
>     at eval (webpack://my-package/../../node_modules/react-flow-renderer/dist/ReactFlow.esm.js?:10838:20)
>     at div
>  ...
> ```

weird, when I change the routes, it occurs.

Himself65:
I saw there's a watermark in the right bottom

https://github.com/wbkd/react-flow/blob/af52bab2a5e185ad00b5bd54b8e05f35181b6b51/src/components/Attribution/index.tsx#L19-L22

How can I hide this "legally"?

Sec-ant:
Hey, I wonder if copy-paste will be added in v10? If so, how are `data` prop gonna be handled? shallow or deep copy? or if there will be an API for users to define their own copy paste bahaviour of `data` in a custom node or edge?

raarts:
Just trying out v10. I tried the new flow with:

```
    <ReactFlow
      nodes={nodes} 
      edges={edges}
      onNodesChange={onNodesChange}
      onEdgesChange={onEdgesChange}
    />
```

I cannot add Edges. The onEdgeChange does not fire.

OrhanTozan:
@raarts I think you should swap nodes={} and edges={} to defaultNodes={} and defaultEdges={}, then it will work.

Check the migration changes documented in the pull request.

raarts:
@OrhanTozan yes, that works, but I was trying to use the controlled API (search for 'Controlled nodes and edges' on this page).

moklick:
@Himself65 we'll launch React Flow Pro by beginning of March. It'll be possible to remove the attribution with a subscription plan or Github sponsoring. 

@raarts you still need the `onConnect` handler. We think that it's common, to pass custom styles to newly created edges. This makes it easier instead of doing it inside the `onEdgesChange` handler.

dih7:
Hello, I'm getting an error on Next JS (v 12.1.0) with the latest version (10.0.0-next.45) of the package. However, the previous version (10.0.0-next.43) works correctly.

Here is the error:

```
Error: require() of ES Module [...]/node_modules/d3-zoom/src/index.js from [...]/node_modules/react-flow-renderer/dist/ReactFlow.js not supported.
Instead change the require of index.js in [...]/node_modules/react-flow-renderer/dist/ReactFlow.js to a dynamic import() which is available in all CommonJS modules.
```

I've created a StackBlitz to replicate the error: [https://stackblitz.com/edit/nextjs-swambw?file=pages/index.js](https://stackblitz.com/edit/nextjs-swambw?file=pages/index.js)

Error from StackBlitz:
```
Error: Must use import to load ES Module: /home/projects/nextjs-swambw/node_modules/d3-zoom/src/index.js
require() of ES modules is not supported.
require() of /home/projects/nextjs-swambw/node_modules/d3-zoom/src/index.js from /home/projects/nextjs-swambw/node_modules/react-flow-renderer/dist/ReactFlow.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /home/projects/nextjs-swambw/node_modules/d3-zoom/package.json.
```

Thanks!

moklick:
Hey @dih7 

thanks for the report! Should be fixed now.

dih7:
Thanks @moklick !

I can confirm that 10.0.0-next.48 is working with Next JS.

dbitkowski:
How can I get the height of a node in v10? I used to do something kinda like this:

`const nodeHeight = useStoreState((state) => state.nodes).find(elm => elm.id === id)?.__rf?.height`

where id is the id of one of my nodes.  This worked well enough in v9, but wondering if there is a better way in v10.  This code seems to break even when using the new useStore.

Thanks! And looking forward to the release of v10

moklick:
@dbitkowski ah.. good point. We will add it to the breaking changes guide. There is no `__rf` anymore. In v10 you could access the height like this: 

```js
const nodeHeight = useStore(
  useCallback((state) => state.nodeInternals.get(id)?.height, [id])
);
```

geclos:
Hi 👋🏼 

First of all congrats on the good work! I'm considering using react-flow for a for-profit project. I've seen the license is MIT but in the PR description you hint at some pricing plans that would need to be paid? Is this gonna be recurrent or a one-time payment?

moklick:
Hey @geclos we are launching React Flow Pro. The MIT license will remain but we expect you to have a pro subscription if you want to remove the "React Flow" attribution. It's a monthly subscription.

OrhanTozan:
@moklick will the removal of the "React Flow" attribution be the only difference between regular vs pro?

moklick:
@OrhanTozan We will release two different plans "starter" and "enterprise". Both plans will allow you to remove the attribution + invitation to private discord channel. The starter plan is for small teams with one project, the enterprise one for larger teams with multiple projects (enterprise also comes with access to datablocks source code + more support possibilities). Don't take this for granted, it's still not 100% sure.

moklick:
@bimal72 March 14th, when everything works as expected :)

jsweetman:
When can we expect to hear about pricing? I really want the nested feature, but I won't be able to easy upgrade the lib in my project due to the new license. Can I start to make a deal now, I'd like to upgrade sooner than later.

OrhanTozan:
@jsweetman check @moklick 's above comment, only think that is exclusive for paid users is removing the attribution. The rest remains free under the MIT license, including the nested feature.

jsweetman:
Unfortunately, showing the watermark is not acceptable in my application.

moklick:
@jsweetman Got it! The cheapest option will be a 129$+ monthly Github sponsorship.

leejustin:
> @jsweetman Got it! The cheapest option will be a 129$+ monthly Github sponsorship.

This is for the starter edition? Or for the enterprise edition?

hgeldenhuys:
> Thanks @moklick !
> 
> I can confirm that 10.0.0-next.48 is working with Next JS.

I think this issue is back in version `10.0.0-next.51`. At least I'm getting it in Remix
https://stackblitz.com/edit/node-xkwl6v?file=app%2Froutes%2Findex.tsx

hgeldenhuys:
Wow that's not "cheap" although I understand you guys also need to be
compensated for all the good work you do.

Are there options for smaller operations or alternative models of
contribution or at least pay according to scale?

pgupta56:
first of all i want to say a BIG THANK YOU for all your wonderful work ! You guys are making our work easier 👍 We are also eagerly waiting for v10.. Requesting to pls share the info about what is "Removing the attribution" in paid version ? are we releasing a community edition as well .. ! Many thanks !

OrhanTozan:
Hi @pgupta56, check the above comments of @moklick, he elaborates more on the pricing.

RegiByte:
Hey guys, if the new react-attribution is on the source code of the library, how do you guys intend to stop people from simply forking the library, removing the attribution from the source, and using this custom version instead of your own? unless the library is not open source anymore...

RegiByte:
Oh I understand @chrtze that`s what I thought, thanks for answering that one, keep up the good work! I`m so excited about the new version release.

moklick:
## We finally launced v10 🎉🎉🎉 

Thanks everyone here for your help and feedback! It helped us a lot :) Check out the new [website](https://reactflow.dev) and especially the [new docs](https://reactflow.dev/docs) (now with lots of guides and search  💯). We also launched [React Flow Pro](https://pro.reactlflow.dev) - the subscription model for React Flow. Tomorrow we will publish a blog post with some more information.

jsweetman:
Did anyone find the performance of dragging nodes decreased in this release?

jsweetman:
Here's a video of v10. You can see the cursor is lagging behind the position of the node. This never happened in the last version.

https://user-images.githubusercontent.com/60225114/158392857-9550156e-0b8e-4d13-89c4-27adf7ea83a0.mov

Panning is similarly laggy

danieliser:
> companies who are powering their business with React Flow are interested in supporting the maintenance and development of the library through a monthly subscription.

"companies who are powering their business with React Flow are interested in supporting the maintenance and development of the library through contributing to the open-source library on github or through a very gratuitously and optional monthly subscription."

There fixed it for you. I mean what is the point of contributing issues, feedback, code, concepts, feature requests to an open source project and then paying to use it.

That would be the equivalent to having to pay McDonalds for a burger, then jumping behind the counter to help make it.

Obviously not for everyone, but definitely makes me NOT want to contribute to the code as now your monetizing any contrition's given under the guise of open-source for personal gain.

As an open source dev myself with hundreds of thousands of users using our free stuff, I understand the need to fund yourselves, but if your not offering anything substantial in terms of actual features for the monthly subscription, my guess is your trading in on your open source contributors for paying users who will demand you do things more quickly than you already are. Hope you can keep up, cause at $200+/mo you better be releasing new functionality every month to keep them coming back, hiding logo alone won't be worth the cost of not forking.

Further your watermark won't help the intended goal. The users who will see it are using an app, not building one (generally speaking). We have tried this as well and its junk traffic all around. Won't drive more users or sales unfortunately as they are not your target market.

Just my 2 cents. Otherwise we really liked the library.

moklick:
Hey @danieliser,

React Flow Pro is just a few days old and it’s very valuable for us to receive feedback from the community. Thanks!

We tried to monetize React Flow with GitHub sponsors for over a year now. We thought that companies are interested in supporting us when they use our library in their products. That was rarely the case. The sponsorships were mostly made by individuals who are using the library for side projects or just wanted to support our work. That’s awesome but they shouldn’t be the ones who are responsible for the financial security. Because the sponsorships didn’t work out we are now trying to put a big sign (the attribution) on the renderer that helps to communicate that we are really interested to maintain and develop the library further. After working 98% voluntarily for two years, we think it’s a reasonable next step. 

Your McDonalds equivalent doesn’t work for me. You can still get the burger for free. The library is still licensed under MIT. We are also pointing out that we don’t want to charge open-source, non-profit or hobby projects. It’s about bigger companies. When you are a startup/ small company and it’s too expensive or when you contributed a lot and it doesn’t feel fair, you can get in contact easily and I am sure we will find a solution. 

We are also great advocates of open source software. That’s why we didn’t want to change the license. It might be an unusual way but we are really interested to keep this library alive. To achieve this we are experimenting with React Flow Pro. If it won‘t work, we’ll try to find another way.

ottovw:
Hey @moklick, 

thanks for listening and addressing community feedback. I understand your - and open source in general - dilemma and believe you should be supported for your great work you put in.

That being said, the price tag is quite high. We're for example are a small company, and the 2.7k would be as high as our hosting costs. As an idea, maybe a model similar to Unity, which is dependent of the users revenue would be helpful. On the other side, we're just going to keep your attribution, its unintrusive enough.

Pro is new, let's let the dust settle and see how the community reacts and continue the discussion.

Thanks again for your great work.

whydea:
Hello.
I was reading the notes and saw that touchscreen support has been added, but the examples themselves in the documentation are not working as expected.
Can anyone review this?

moklick:
@whydea what is not working for you? Nodes should be draggable and an connection is made by tap on a handle and then on another one: https://reactflow.dev/docs/examples/interaction/touch-device/

moklick:
@whydea there was a bug with the latest version. This is fixed in v10.3.1

whydea:
> @whydea there was a bug with the latest version. This is fixed in v10.3.1

Hello
The problem happened when dragging a drag node, but I just saw that this was fixed in version 10.3.1 generated a few minutes ago. It is worth mentioning that the examples on the site are still having problems. Is there a schedule for updating the examples in the documentation?

mikkmartin:
Hey, thanks for the free software! Great library. That being said – I'm mildly annoyed at the mandatory attribution banner. I totally understand the point, but since the software license is MIT – you should be legally permitted to "change" the software and remove it? Maybe not having MIT or having a modified version of it would make more sense?

chrtze:
Hi @mikkmartin, thanks for your feedback. We are aware of this and are in the process of changing some things regarding the attribution. Please see https://github.com/wbkd/react-flow/issues/2288 which will be added to the next major release

Prasy12:
Hi @moklick ,
The useStoreApi.getState does not have a nodes: Node[]; attribute in v10? This was present in v9 useStoreState 

export declare type ReactFlowStore = {
    width: number;
    height: number;
    transform: Transform;
    nodeInternals: NodeInternals;
    edges: Edge[];
    onNodesChange: OnNodesChange | null;
    onEdgesChange: OnEdgesChange | null;
    hasDefaultNodes: boolean;
    hasDefaultEdges: boolean;
    domNode: HTMLElement | null;
    d3Zoom: ZoomBehavior<Element, unknown> | null;
    d3Selection: D3Selection<Element, unknown, null, undefined> | null;
    d3ZoomHandler: ((this: Element, event: any, d: unknown) => void) | undefined;
    minZoom: number;
    maxZoom: number;
    translateExtent: CoordinateExtent;
    nodeExtent: CoordinateExtent;
    nodesSelectionActive: boolean;
    userSelectionActive: boolean;
    connectionNodeId: string | null;
    connectionHandleId: string | null;
    connectionHandleType: HandleType | null;
    connectionPosition: XYPosition;
    connectionMode: ConnectionMode;
    snapToGrid: boolean;
    snapGrid: SnapGrid;
    nodesDraggable: boolean;
    nodesConnectable: boolean;
    elementsSelectable: boolean;
    multiSelectionActive: boolean;
    reactFlowVersion: string;
    connectionStartHandle: StartHandle | null;
    onNodeDragStart?: NodeDragHandler;
    onNodeDrag?: NodeDragHandler;
    onNodeDragStop?: NodeDragHandler;
    onSelectionDragStart?: SelectionDragHandler;
    onSelectionDrag?: SelectionDragHandler;
    onSelectionDragStop?: SelectionDragHandler;
    onConnect?: OnConnect;
    onConnectStart?: OnConnectStart;
    onConnectStop?: OnConnectStop;
    onConnectEnd?: OnConnectEnd;
    onClickConnectStart?: OnConnectStart;
    onClickConnectStop?: OnConnectStop;
    onClickConnectEnd?: OnConnectEnd;
    connectOnClick: boolean;
    defaultEdgeOptions?: DefaultEdgeOptions;
    fitViewOnInit: boolean;
    fitViewOnInitDone: boolean;
    fitViewOnInitOptions: FitViewOptions | undefined;
    onNodesDelete?: OnNodesDelete;
    onEdgesDelete?: OnEdgesDelete;
};

moklick:
It's a `Map` now (`nodeInternals `). You can get the nodes as an array via: `Array.from(nodeInternals.values())`

Prasy12:
Thanks @moklick . Also the setSelectedItems has been removed?
image
It shows undefined if I use this method to read it

moklick:
@Prasy12 nodes and edges now have a `selected` attribute that you can use.

Prasy12:
@moklick Is there anyway in the v10 version to highligh the edges that correspond to a node when a node is selected?

Prasy12:
@moklick 
Is there a way the bounding box of the subflows appear automatically without us specifying the style attributes? This would largely help if we have dynamic number of nodes inside a subflow and the bounding box changes size dynamically. 
Currently we have to manually specific the height and width
image

moklick:
@rammazzoti2000 @Prasy12 . I will lock this issue. Please open a new issue or discussion. This issue was meant for the v10 release.

PR #1574, opened 2021-10-11
fix(ControlButton): avoid submit form by click on default ControlButtons

EmilienLeroux:
Depending on the browser, if we use ReactFlow in a form, a default ControlButton can submit the form : “type” attribute is missing on buttons.


EmilienLeroux:
Hi @moklick!

Absolutely you're right it's more concise and elegant, I have commit this modification. :)

moklick:
thanks!

PR #1621, opened 2021-10-28
Fix Firefox not being transitioned

PhilGarb:
## Why?

While implementing react-flow in our application I wanted to move the selectedNode to the center of the screen. This makes sure the user can find it even on larger trees. 

To achieve that I added a transition to the two places the translation is applied:

```css
  .react-flow__nodes, .react-flow__edges *: {
      transition: transform 500ms ease-in-out,
    }
```

This works fine in Chrome,. When testing in Firefox the transition did not work however. This seemed to be related to the transfrom being applied as an attribute by react-flow. Attributes are only targeted by the [SVG2](https://www.w3.org/TR/SVG2/) spec which has already been partially implemented in Chrome, but not in Firefox. 

## Fix

To fix this I tested moving the transform into a style tag like it is done for the nodes.
https://github.com/wbkd/react-flow/blob/c5ce586dd9d6a53e600cbff4e9257695fb08e601/src/container/NodeRenderer/index.tsx#L61

With that fix it worked for me in both Chrome and Firefox.

### DIFF
--- a/src/container/EdgeRenderer/index.tsx
+++ b/src/container/EdgeRenderer/index.tsx
@@ -208,13 +208,13 @@ const EdgeRenderer = (props: EdgeRendererProps) => {
     connectionLineComponent,
     onlyRenderVisibleElements,
   } = props;
-  const transformStyle = `translate(${transform[0]},${transform[1]}) scale(${transform[2]})`;
+  const transformStyle = `translate(${transform[0]}px,${transform[1]}px) scale(${transform[2]})`;
   const renderConnectionLine = connectionNodeId && connectionHandleType;
 
   return (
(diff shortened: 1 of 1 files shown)


moklick:
Hey @PhilGarb 

thanks for the fix. We changed a lot for our next major release and there is no transform for the group anymore. We only do one transform in the parent of the edge- and nodes-renderer. I'll merge this anyway ;) I think we need one more month until the release is ready..

PhilGarb:
Hi @moklick 

thank you for taking the time to check out the PR. I followed your progress with version 10 and I'm glad this issue is going to be fixed there as well.

moklick:
released in v9.7.2

PR #1666, opened 2021-11-12
accept no key for selection

mikkmartin:


moklick:
This doesn't work. It just sets `selectionKeyPressed` always to `true`.

PR #1698, opened 2021-11-19
Add always on top or bottom option for nodes

yuniyakim:
I added two node options `isAlwaysOnTop` and `isAlwaysOnBottom` that allow it to stay always on top and bottom of others respectively. 
In case two or more nodes have always on top/bottom option enabled, they are shown in the same order as default nodes, i.e. the one that was rendered later is above.
If options are assigned to be `true` at the same time, both of them become `false`.

Additionally, I added `ReactFlow` interaction prop `nodesAlwaysOnTopOrBottom` to enable or disable this feature.

The corresponding example was included in `Update Node` example. Tests were also added.


moklick:
Hey @yuniyakim 

thanks for your PR! What is the advantage of having these new props over using `style : { zIndex: 999 }` for nodes that should stay on top?

yuniyakim:
> Hey @yuniyakim 
> 
> 
> 
> thanks for your PR! What is the advantage of having these new props over using `style : { zIndex: 999 }` for nodes that should stay on top? 

I suppose that on some point user will need to change node's class to default or always on bottom (or whatever) and it's more convenient not to change its' style every time but to change the corresponding prop.

moklick:
I think this is an abstraction that we don't need, because we can already do it with `z-index`. For this we could use the classname or the style option. There are already too many props and options in my view :D

PR #1699, opened 2021-11-21
Smooth transitions for useZoomPanHelper using built-in d3 functions

Mahdi-Esbati:
Smooth transitions were a highly requested feature for react-flow that could be implemented using external libraries like popmotion or tween.js, but this approach is a bit messy since you need to use the reactFlowRef and change its internal viewport directly.

There are some [built-in functions](https://github.com/d3/d3-zoom#zoom_transform) in D3.js that can help with smooth transitions without using any external libraries.  

I've refactored the `useZoomPanHelper` hook functions to use d3 transitions and added the corresponding examples to `/examples/UseZoomPanHelper` directory.

I've also added an optional duration argument to the `zoomIn, zoomOut, zoomTo, transform, fitView, setCenter` and `fitBounds` functions to help the user to change the duration of d3 transitions.

By default, all the functions above have a 300ms transition, but their animations can be disabled by setting their duration to 0.

Here is the UseZoomPanHelper example with `zoomIn` set to `1200ms`, `zoomOut` set to 0 and `setCenter` set to the default value of `300ms`:

https://user-images.githubusercontent.com/17250923/142758194-8c296261-bbe6-45ad-a0dd-52e35fde1e47.mp4

### DIFF
--- a/example/src/UseZoomPanHelper/index.tsx
+++ b/example/src/UseZoomPanHelper/index.tsx
@@ -29,7 +29,7 @@ const UseZoomPanHelperFlow = () => {
   const [elements, setElements] = useState<Elements>(initialElements);
   const onElementsRemove = (elementsToRemove: Elements) => setElements((els) => removeElements(elementsToRemove, els));
   const onConnect = (params: Connection | Edge) => setElements((els) => addEdge(params, els));
-  const { project } = useZoomPanHelper();
+  const { project, setCenter, zoomIn, zoomOut } = useZoomPanHelper();
 
   const onPaneClick = useCallback(
     (evt) => {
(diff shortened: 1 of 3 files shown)


Mahdi-Esbati:
In case anyone wonders why transitions are [implemented in such a way](https://github.com/wbkd/react-flow/pull/1699/commits/83362c187cffb01c7a4c04596a06ce6b2c263289):
```js
const getTransition = (selection: D3Selection<Element, unknown, null, undefined>, duration: number = 300) => {
  return selection.transition().duration(duration);
};

return {
// ...
  zoomIn: (duration?: number) => d3Zoom.scaleBy(getTransition(d3Selection, duration), 1.2),
  zoomOut: (duration?: number) => d3Zoom.scaleBy(getTransition(d3Selection, duration), 1 / 1.2),
  zoomTo: (zoomLevel: number, duration?: number) =>
          d3Zoom.scaleTo(getTransition(d3Selection, duration), zoomLevel),
// ...
}
```

and not like this:

```js
const animatedSelection = d3Selection.transition().duration(300);

return {
// ...
  zoomIn: (duration?: number) => d3Zoom.scaleBy(animatedSelection , 1.2),
  zoomOut: (duration?: number) => d3Zoom.scaleBy(animatedSelection , 1 / 1.2),
  zoomTo: (zoomLevel: number, duration?: number) =>
            d3Zoom.scaleTo(animatedSelection , zoomLevel),
// ...
}
```

The answer is: [d3 transitions are not reusable](https://stackoverflow.com/a/64596738/5693949) and have to be built using a factory function. If you try to reuse a transition, you'll get the following error: `Uncaught error: transition not found`

Mahdi-Esbati:
@moklick 
Do you have any updates/suggestions about this feature?

moklick:
Looks good to me :) Could you just set the default duration to `0` ?

Mahdi-Esbati:
Yes of course!
I'll create new commit and change it.

visualsitemaps:
nice work!

moklick:
awesome, thanks!

moklick:
released in 9.7.0. We will probably change the API a bit in v10 in order to always use an object for the options.

PR #1728, opened 2021-11-29
Fix bug in updateNodePosDiff action

xixixao:
This is the intended logic. Without it triggering this action with a node id will always move selected nodes, which is not desirable if we want to move just an unselected node.

The bug doesn't show up in current examples because at the moment either one selected node is moved or all the selected nodes are moved. TBH this should probably be changed to accept `ids` instead to handle the two cases instead of this tight coupling between the group selection code and this action.


moklick:
good catch, thanks!

moklick:
unfortunately this breaks our multi selection functionality. I had to pull it out again.

PR #1784, opened 2021-12-22
fix: Fix horizontal scroll issue on windows chrome with panOnScroll true

s-yadav:
Windows doesn't update deltaX when horizontally scrolled with shift + mouseWheel instead changes deltaY in such case as well, this PR fixes it by checking if shift is pressed and correct delta is not updated.

Codesandbox link for wheel event:  https://codesandbox.io/s/jolly-bogdan-b5e01?file=/src/index.js
Screenshot:
Screenshot 2021-12-22 090425

### DIFF
--- a/src/container/ZoomPane/index.tsx
+++ b/src/container/ZoomPane/index.tsx
@@ -127,8 +127,19 @@ const ZoomPane = ({
             // increase scroll speed in firefox
             // firefox: deltaMode === 1; chrome: deltaMode === 0
             const deltaNormalize = event.deltaMode === 1 ? 20 : 1;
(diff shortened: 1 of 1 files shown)


moklick:
Thanks for your help @s-yadav but this change breaks the behaviour for macos. I am wondering if there is way to fix this without checking the os.

moklick:
I would like to fix this but the issue is that there are no wheel events as soon as I press "Shift". Can anyone confirm this?

moklick:
It seems it's an issue with Browserstack ..

moklick:
Thanks for your help here. Since a lot has changed, I will take your changes and apply them to the current version and also add an OS check.

moklick:
fixed in v11.8.0 - thanks @s-yadav

s-yadav:
Thanks @moklick 🙏

PR #1862, opened 2022-01-30
update deps

andreyworkspace:
update deps
now examples work on nodejs 17
edits for latest version react-router-dom


moklick:
thanks!

PR #1926, opened 2022-02-23
fix: type forceUpdate on NodeDiffUpdate

Himself65:


Himself65:
my mistake, read something wrong

PR #1947, opened 2022-03-05
fix: type error

Himself65:
Type incorrect
Refs: https://github.com/wbkd/react-flow/pull/1555/commits/414d7ec993ed66d4ceb0c521176f94128eec6eef
Upstream: #1555


moklick:
thanks! This shouldn't happen with the non-next build, because we run the test before we publish it. We should probably do this for the next build, too..

PR #1964, opened 2022-03-11
Pass isSourceHandle flag on to onEdgeUpdateStart

andreas-schultz:
- This allows to know during an edge update which side of the edge is getting dragged by the user.
- Use cases include showing custom validation hints to the user during an edge update, e.g. which handles can be connected to and which not. This would otherwise not be possible without knowing which side is updated.


seebi:
@andreas-schultz thanks for providing this - @moklick do we need to follow a contribution guide to get this merged?

moklick:
Hey @andreas-schultz,

thanks for your PR and sorry that it got lost in the shuffle.
You you just change the boolean to "source"/ "target". That makes it a bit more generic for possible changes in the future.
So instead of `isSourceHandle`, the param should be `handleType`. Thanks!

andreas-schultz:
Hi @moklick,

sure, makes sense, I've changed the code as you proposed.

andreas-schultz:
@moklick Alright,done. I've also updated the `UpdatableEdge` example.

moklick:
released in v10.1.0

PR #1967, opened 2022-03-14
Unreal Engine blueprint style edge

joeyballentine:
This PR adds an Unreal Engine blueprint style bezier edge that looks like this:

image

Apologies for the slight formatting changes


joeyballentine:
It only just now occurred to me that the scaling problem I mentioned that happens the further apart you make the nodes can probably be fixed by just adding/subtracting a constant value rather than multiplying by a curvature amount. Will attempt tomorrow.

moklick:
awesome 🙏 ! Thanks!

PR #1968, opened 2022-03-14
Bezier Edge: Add curvature

moklick:


moklick:
Thanks again @joeyballentine 💯

PR #1971, opened 2022-03-15
Fix new BezierEdge preview when connecting from target to source

### LINKED ISSUE #1970: Connection line from handles on the left or on the top have wrong curve
### Describe the Bug

When you drag from a handle on the left side on the node, the curve looks as if the handle was on the right side.

- Go to the [Horizontal demo](https://reactflow.dev/docs/examples/horizontal/).
- Start dragging from the left handle of Node 7.
- The curve looks as if the handle was on the right side.

### Expected behavior

The curve should go to the right side of the cursor.

### Screenshots or Videos

image

joeyballentine:
Basically this just swaps the target and source positions if it needs to in order to display the right preview.

image

Closes #1970

### DIFF
--- a/src/components/Edges/BezierEdge.tsx
+++ b/src/components/Edges/BezierEdge.tsx
@@ -28,38 +28,45 @@ export function getBezierPath({
   centerY,
 }: GetBezierPathParams): string {
   const leftAndRight = [Position.Left, Position.Right];
+
+  const inverted = sourcePosition === Position.Left || sourcePosition === Position.Top
+  const _targetX = inverted ? sourceX : targetX;
+  const _targetY  = inverted ? sourceY : targetY;
+  const _sourceX = inverted ? targetX : sourceX;
+  const _sourceY = inverted ? targetY : sourceY;
+
   const hasCurvature = curvature > 0;
   let cX,
     cY = 0;
 
-  const [_centerX, _centerY] = getCenter({ sourceX, sourceY, targetX, targetY });
+  const [_centerX, _centerY] = getCenter({ sourceX: _sourceX, sourceY: _sourceY, targetX: _targetX, targetY: _targetY });
 
   if (leftAndRight.includes(sourcePosition) && leftAndRight.includes(targetPosition)) {
     cX = typeof centerX !== 'undefined' ? centerX : _centerX;
-    const distanceX = targetX - sourceX;
+    const distanceX = _targetX - _sourceX;
     const absDistanceX = Math.abs(Math.min(0, distanceX));
     const amtX = (Math.sqrt(absDistanceX) / 2) * (50 * curvature);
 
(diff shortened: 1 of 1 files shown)


Holybasil:
Hi, I applied it locally, and it looks like **source** and **target** are swapped, because **markerEnd** is at **source**.

moklick:
Hey @joeyballentine 

thanks for the PR! Unfortunately we can't swap source and target. It's important that the path goes from source to target. Otherwise the animation runs into the wrong direction and markerEnd and markerStart doesn't work either as @Holybasil pointed out.

moklick:
I pushed some adjustments for the bezier edge, so that it works for undirectional/ right to left/ bottom to top flows too. For the connection line we to implement another helper function. We don't know the target position while dragging the connection line, so it needs to behave a bit different.

Sec-ant:
@moklick I'm a little confused here: https://github.com/wbkd/react-flow/blob/8ca8972ed7308ed8f6741a62f85fd0e3a544c614/src/components/ConnectionLine/index.tsx#L74-L75 From my understanding, the connection line can either be dragged from source handle or from target handle. If I drag the connection line from the target handle of one node to the source handle of another, the source position of the connection line should be the target handle's position and the target position of the connection line should be the source handle's position. So the naming here is really confusing.

And let's say here is a left / right flow, according to these two lines of code, the target position will always be `Position.Left`, no matter from which handle I drag the conneciton line, which means both the source and target positions can be `Position.Left`, and will produce a weird line shape.

Sec-ant:
> @moklick I'm a little confused here:
> 
> https://github.com/wbkd/react-flow/blob/8ca8972ed7308ed8f6741a62f85fd0e3a544c614/src/components/ConnectionLine/index.tsx#L74-L75
> 
> From my understanding, the connection line can either be dragged from source handle or from target handle. If I drag the connection line from the target handle of one node to the source handle of another, the source position of the connection line should be the target handle's position and the target position of the connection line should be the source handle's position. So the naming here is really confusing.
> And let's say here is a left / right flow, according to these two lines of code, the target position will always be `Position.Left`, no matter from which handle I drag the conneciton line, which means both the source and target positions can be `Position.Left`, and will produce a weird line shape.
> 
> In that a source handle can only be connected to a target handle, or vice versa, and their positions are usually in the opposite direcitons, I think we should change the above two lines of code into:
> 
> ```ts
> let targetPosition: Position;
> switch (sourceHandle?.position) {
>   case Position.Left:
>     targetPosition = Position.Right;
>     break;
>   case Position.Right:
>     targetPosition = Position.Left;
>     break;
>   case Position.Top:
>     targetPosition = Position.Bottom;
>     break;
>   case Position.Bottom:
>     targetPosition = Position.Top;
>     break;
>   default:
>     targetPosition = Position.Top;
>     break;
> }
> ```

I opened a new PR here concerning this problem: https://github.com/wbkd/react-flow/pull/1981

PR #1981, opened 2022-03-18
refactor connection line

Sec-ant:
In that the connection line can be dragged from either the source or target handle, the `source` and `target` naming can be confusing. I suggest using `from` and `to` in a connection line, where the `from` or `to` can either be a `source` or a `target`.

Meanwhile, the `source` and `target` passed to the `CustomConnectionLineComponent` should be consistent with the created edge when the connection is finished.

### DIFF
--- a/src/components/ConnectionLine/index.tsx
+++ b/src/components/ConnectionLine/index.tsx
@@ -4,15 +4,7 @@ import shallow from 'zustand/shallow';
 import { useStore } from '../../store';
 import { getBezierPath } from '../Edges/BezierEdge';
 import { getSmoothStepPath } from '../Edges/SmoothStepEdge';
-import {
-  HandleElement,
-  ConnectionLineType,
-  ConnectionLineComponent,
-  HandleType,
-  Node,
-  ReactFlowState,
-  Position,
-} from '../../types';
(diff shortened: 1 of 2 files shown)


Sec-ant:
btw, I think we should change the default conneciton line back to bezier but not simple bezier, because in most cases, the source handle and target handle are in opposite positions, so even if we don't know the destination of the conneciton line, we can usually assume the destination has an opposite position.

Or, if that's unacceptable, at least we know the position from where the connection is dragged, so we can always add curvature in `from` while zero the curvature in `to`. In that case, we should refactor the Bezier function to accept two curvature parameters (one for the `source` and one for the `target`).

moklick:
Thanks for your help @Sec-ant! I really appreciate it! I'll check your PRs as soon as I find some time :)

Sec-ant:
> Thanks for your help @Sec-ant! I really appreciate it! I'll check your PRs as soon as I find some time :)

No hurry, have a good weekend 😄

joeyballentine:
Does this PR still need to be open? These changes included in #1984 right?

joeyballentine:
@moklick looking at the commits in each, yeah #1984 definitely contains these commits

Sec-ant:
> @Sec-ant does #1984 already include these changes? 

yes, should I close this?

Issue #1985, opened 2022-03-18
Nodes multi selection doesn't work with selectNodesOnDrag disabled


alessandropisu:
### Describe the Bug

I notice that nodes multi selection doesn't work when selectNodesOnDrag  is false, both in Controlled and Uncontrolled mode. It doesn't seems to be the right behavior as is doing a different functionality compared to nodes selection on drag.

### Your Example Website or App

https://codesandbox.io/s/sandpack-project-forked-3l48jx?file=/App.js

### Steps to Reproduce the Bug or Issue

Select a node, then click a second node with Meta/Command pressed

### Expected behavior

All nodes should be selected and not only the last one

### Platform

- OS: macOS
- Browser: Chrome
- Version: v10

Sec-ant:
I can confirm this has something to do with this line of code: https://github.com/wbkd/react-flow/blob/3548891b8e4f6ffbb9de1fbbef582d8eb55304c7/src/components/Nodes/wrapNode.tsx#L115

If you comment it out, you'll be able to multiselect nodes. However the logic here is too complicated to be simply disable this because this function guarantees when you drag another node, the previous selected nodes will be deselected.

Sec-ant:
I made a PR to fix this issue: https://github.com/wbkd/react-flow/pull/1988

PR #1988, opened 2022-03-19
fix multiSelection when selectNodesOnDrag is false

Sec-ant:
should close https://github.com/wbkd/react-flow/issues/1985, but I'm not sure adding more logic and making the code harder to debug will be the right choice.


Kinark:
This doesn't seem to work, I modified my node_modules files and it still seems to drags both.

moklick:
thanks @Sec-ant

PR #2014, opened 2022-03-24
refactor <NodeWrapper/> mouse event handlers

Sec-ant:
- Separated concerns of multiple mouse events: `onMouseDown`(for selection), `onClick`, `onDragStart`, `onDrag`, `onDragStop` in `<NodeWrapper/>`. Remove some duplicated "if condition" and "state management" code that may cause potential bugs (like this one: https://github.com/wbkd/react-flow/issues/1985).
- Added some comments, I hope that's acceptable.
- ⚠️ Breaking change: specifying a selector to be used as the drag handle will not prevent the whole node from being selected. For compatiblity reasons, you can use `nodrag` classname to explictly set some elements in the node nonselectable. But there're two problems here:
  1. This classname api cannot handle dynamically added elements with `nodrag` classnames, I'd prefer a hook api.
  2. If we use this classname api and somehow fix the problem in 1, the classname should still be `noselect` instead of `nodrag`. I don't think it's a good idea to not distinguish these two actions.

### DIFF
--- a/src/components/Nodes/wrapNode.tsx
+++ b/src/components/Nodes/wrapNode.tsx
@@ -179,7 +179,7 @@ export default (NodeComponent: ComponentType<NodeProps>) => {
     );
 
     // on drag stop handler, will get triggered by click
-    // but click is already handled in onClickNodeHandler
+    // but click is already handled in onNodeClickHandler
     // so we don't need to handle it in this handler
     const onDragStop = useCallback(
       (event: DraggableEvent) => {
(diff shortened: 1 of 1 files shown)


Sec-ant:
Stop mousedown propagation will break many things, I'll mark this PR as draft until I fix it.

Sec-ant:
Yay! I'll check!

moklick:
Hey @Sec-ant Thanks for your work here but too many things changed in the meantime. Please feel free to open again!

Issue #2022, opened 2022-03-26
multi selection (with Meta key) is always on when a hotkey is used


RobCvdV:
### Describe the Bug

On a mac, when using the Meta key to multiselect nodes (default), after using a hotkey like cmd+c  multiselect is always on.

The behaviour is different when using other keycodes. usaing multiSelectionKeyCode="Alt" results in not being able to multi select after using a hotkey.

Note the console logging in the example for both using "Meta" or "Alt", it appears the key pressed or not pressed won't work properly after using any hotkey, no matter which hotkey shortcut it used.

### Your Example Website or App

https://codesandbox.io/s/react-flow-multiselect-bug-h28ih8

### Steps to Reproduce the Bug or Issue

1. setup/use a boilerplate react-flow app.
2. have several nodes
3. press cmd+c anywhere while the flow component is focused
4. select nodes without the cmd/meta key
5. selection is being added instead of replaced

### Expected behavior

without the Meta key selection should replace any other selection (so deselect other nodes and edges)

### Platform

- OS: MacOS (only one tested)
- Browser: Chrome, Safari, Firefox
- Version: -

RobCvdV:
possible solution in PR #2030

moklick:
Should be fixed in v10.0.7

PR #2030, opened 2022-03-29
fixed useKeyPress issues, easily interfered by (default) hotkey

RobCvdV:
useKeyPress could get in a constant pressed (true) state when a key combo using one of the keys would be used. This resulted in the Meta or Control being "stuck", whenever trying to use very common hotkeys. It also made any usefull hotkey implementation in an app impossible.
 
[https://github.com/wbkd/react-flow/issues/2022](See bug report about multiselection getting stuck.)

The new implementation won't support combo keys itself, but they didn't work properly before anyway. Using useHotkeys for combo's is much better and well proven. A very light dependency.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,7 @@
         "d3-selection": "^3.0.0",
         "d3-zoom": "^3.0.0",
         "react-draggable": "^4.4.4",
+        "react-hotkeys-hook": "^3.4.4",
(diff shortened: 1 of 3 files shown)


RobCvdV:
fixes issue #2022

moklick:
Thanks for the hint @RobCvdV ! Should be fixed in v10.0.7

PR #2041, opened 2022-04-03
Ability to change default edge zIndex

joeyballentine:
I need this feature in order to fix an issue I have where the connection line goes behind my group node when there are no edges in the edge array. This can be fixed by changing the default Z level that the connection line goes in when no nodes are present.

I have no idea if I added all the necessary things to get this to work, and to be honest, I'm completely guessing. I also can't really test this myself so I have no idea if it works. It honestly might be better to just reimplement this idea however you all would usually do it.

Refer to this discord exchange for details: https://discord.com/channels/771389069270712320/859774873500778517/959992983082663986

### DIFF
--- a/src/components/StoreUpdater/index.tsx
+++ b/src/components/StoreUpdater/index.tsx
@@ -49,6 +49,7 @@ interface StoreUpdaterProps {
   fitViewOptions?: FitViewOptions;
   onNodesDelete?: OnNodesDelete;
   onEdgesDelete?: OnEdgesDelete;
+  defaultZLevel?: number,
 }
 
 const selector = (s: ReactFlowState) => ({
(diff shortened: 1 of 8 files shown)


joeyballentine:
For an example on what this affects, see this screenshot:
image

I have no way of fixing this, since react-flow defaults the edge group to a zIndex of 0 and my nodes are at a zIndex of 50 (due to my sorting code). I guess I could make it default to 0 if there are no edges but I think that would break how my group nodes get displayed when on top of each other. It would be much easier to just be able to specify that I want the connectionline to be on zIndex level 999 or something so it's on top by default.

joeyballentine:
Actually, doing what I mentioned above but setting the zIndex of all my nodes to -1 works well enough. But I'd still like to see this fixed at some point

bcakmakoglu:
@cliffordfajardo I don't think the z-index is the actual issue here, it looks like your nodes are simply transparent and have no background color, thus you will always see what's below them.
Try adding a background color like white.

cliffordfajardo:
@bcakmakoglu - you're 100 right - that was it 💯

PR #2042, opened 2022-04-04
Fix node handleBounds not updated on updateNodeInternals

ratasorin:
When using ```updateNodeInternals```, a ```dimensions``` event is fired, that should also update the handlers on the node. However, when reacting to the ```dimensions``` event: 
```js
 if (typeof currentChange.dimensions !== 'undefined') {
            updateItem.width = currentChange.dimensions.width;
            updateItem.height = currentChange.dimensions.height;
}
``` 
we only change the width and the height on that node, leaving the new handlers out.


moklick:
Thanks for your PR but we no longer want to expose any internals to the user (see #2178). We are already [setting the handle bounds internally](https://github.com/wbkd/react-flow/blob/main/src/store/index.ts#L61-L65) as soon as we can, so there is no need to send it via a node change.

PR #2058, opened 2022-04-08
fix: type error

Himself65:
Refs: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/56210


moklick:
thanks @Himself65 !

moklick:
added a PR to track React 18 migration progress here: #2067

PR #2062, opened 2022-04-11
Feat/umdbuild

moklick:


ERuseckas:
Hi, @moklick ,

can you achieve it while still using Node v12?
I can not use Node v14, so I wonder am I stuck with react-flow until this PR got in or is there a way to keep Node v12 support.

PR #2096, opened 2022-04-23
Implemented handleSelector functionality

furkankly:
Referenced the [react-draggable](https://github.com/react-grid-layout/react-draggable/blob/31798e920647f40308a144a9f989c771755a21db/lib/utils/domFns.js#L31) implementation.
Didn't care about vendor-prefixed implementations as in `react-draggable` due to the standardization of `element.matches`.
Realized a bug for `onNodeDragStop`, added a TODO indication for it. (The bug already existed before this PR.)


furkankly:
There was a bug with `onNodeDragStop` as mentioned before.
We had to put `onDragStop` into `useDrag`'s dep. list so that `dragging` (the global `dragging` state of each node) could affect the render cycle.
But even if we do it, a `d3-drag` gesture (`drag().on(...`) could not be modified on the fly as stated in [d3-drag-docs](https://github.com/d3/d3-drag#drag_on).
So I decided to do what the docs suggested and used `event.on(...` api to actually define the `d3-drag`'s `end` event handler in the `drag` event handler. 
Since we don't have an event handler defined for the `end` event at all times in the context of dragging now, we were able to get rid of the `dragging` state which was nothing but a flag to distingush `mouseup`s fired upon normal user clicks from the drag related ones. This let us get rid of the burden for maintaining such flag and simplifies the overall implementation IMO.

We expose `dragging` in `NodeProps` to be used with custom nodes, and since I didn't want to change the public api, I created a local state in `wrapNode` for it.

I fixed the `drag` command for cypress to behave correctly with `d3` and run all tests.

This should complete [`d3-drag` refactoring](https://github.com/wbkd/react-flow/pull/2072).

bcakmakoglu:
I'm curious:
Instead of adding `if (isDragAllowedBySelector)` to each drag event
```ts
.on('start', () => {
  if (isDragAllowedBySelector) {
  // drag handler
  }
}
```

can you not just add it to the filter to prevent dragging if it's not on the `handleSelector`?
So something like this:
```ts
 .filter((event: any) => !!(handleSelector && selectorExistsTargetToNode(event.sourceEvent.target, handleSelector, nodeRef)) && !event.ctrlKey && !event.button && !event.target.className.includes(noDragClassName));
```

furkankly:
> I'm curious: Instead of adding `if (isDragAllowedBySelector)` to each drag event
> 
> ```ts
> .on('start', () => {
>   if (isDragAllowedBySelector) {
>   // drag handler
>   }
> }
> ```
> 
> can you not just add it to the filter to prevent dragging if it's not on the `handleSelector`? So something like this:
> 
> ```ts
>  .filter((event: any) => !!(handleSelector && selectorExistsTargetToNode(event.sourceEvent.target, handleSelector, nodeRef)) && !event.ctrlKey && !event.button && !event.target.className.includes(noDragClassName));
> ```

Thanks for being an eye opener! Addressed [here](https://github.com/wbkd/react-flow/pull/2096/commits/5737d133d409983578085517946ec1d51cbc841f) with a slight difference from what you suggested as parentheses wouldn't do anything whatsoever and we would return `false` in case `handleSelector` is not defined.

bcakmakoglu:
@moklick 
Wouldn't it be enough to just calculate the deltas again? 
Something like this?
```ts
if (lastPos.x && lastPos.y) {
  onDrag(event, {
    dx: pos.x - lastPos.x,
    dy: pos.y - lastPos.y,
  })
}
```

moklick:
@bcakmakoglu sure. We also get the deltas from d3-drag but it works better for single nodes when we use the x and y values. When you move the mouse quickly, sometimes the node doesn't stick to the mouse. This bug doesn't occur when we use the x and y values.

bcakmakoglu:
Yeah I noticed the weird position de-sync when you just wildly drag around using "deltas" 😅
Good to know

src/components/Nodes/wrapNode.tsx
```diff
@@ -121,15 +130,17 @@ export default (NodeComponent: ComponentType<NodeProps>) => {
...
 
+    // As one of the props passed to a custom node
+    const [dragging, setDragging] = useState(false);
+
     const onDrag = useCallback(
       (event: UseDragEvent, dragPos: UseDragData) => {
-        updateNodePosition({ id, dragging: true, diff: { x: dragPos.dx, y: dragPos.dy } });
-
+        updateNodePosition({ id, diff: { x: dragPos.dx, y: dragPos.dy } });
+        setDragging(true);
```
moklick, line 139: could you set this in the `onDragStart` handler?

src/components/Nodes/wrapNode.tsx
```diff
@@ -146,36 +157,13 @@ export default (NodeComponent: ComponentType<NodeProps>) => {
...
-        }
-
-        updateNodePosition({
-          id,
-          dragging: false,
-        });
-
-        if (onNodeDragStop && node) {
-          onNodeDragStop(event.sourceEvent as MouseEvent, { ...node, dragging: false });
+        setDragging(false);
```
moklick, line 160: by removing the code from the original onDragStop handler you broke the `onClick` handler for draggable nodes. Could you add lines 151 - 167 again? thanks!

moklick (changes requested):

moklick:
Hey @furkankly 

thanks again for your help! I fixed some minor issues with the drag handlers and merged everything into refactor/d3-drag.

furkankly:
> Hey @furkankly
> 
> thanks again for your help! I fixed some minor issues with the drag handlers and merged everything into refactor/d3-drag.

Hey, sorry for joining the discussion a little late.
I just realized, I actually forgot the redundant `isDraggable` check in the `onSelectNodeHandler` handler with the new drag implementation. (`line 90 in wrapNode.tsx`) 
This should fix the regression related to `onClick` with the draggable nodes without requiring a ref to track if the related node is currently dragging. 
This should also fix another regression that comes with [this commit](https://github.com/wbkd/react-flow/pull/2072/commits/fa582daa7a710087ed76aec53d79c9313a3896f0) which is "not being able to fire `onClick`s again but this time  for the nodes with a drag-handle". Since we will not be defining `onDragStop` outside of the drag-handle, we won't be able to detect the clicks outside of the handle. Let me know if there is another regression I can not see with the removal of `isDraggable` check I mentioned above. If there is no more regression, this will let us to have a single place for handling `onClick`s being `onSelectHandler`.

I ran all the tests again, there were a few ones failing but those were not related to dragging (`react-18` update IMO). Since they also currently fail for `refactor/d3-drag`.

src/components/Nodes/wrapNode.tsx
furkankly: I think we can set it in `onDragStart` to be consistent with the concept of "starting the drag action".
I just think doing this in `onDrag` has the benefit of letting users play with this state in a custom node implementation more easily. 

To elaborate with an example:
When a user clicks on a node with both `onDragStart` and `onNodeClick` defined, these are the possibilities with the `d3-drag` implementation :
If the user moves the mouse upon clicking:
`mousedown` fires, `mousemove` fires and then `mouseup` fires. We call `onDragStart`, `onDrag` and `onDragEnd`.
If the user doesnt move the mouse upon clicking:
`mousedown` fires, `mouseup` fires and then `click` fires. We call `onDragStart` and `onClick`. 

So there is this intersection zone caused by `onmousedown` firings for `onDragStart` and `onClick`. But its kinda expected IMO (or we can try to find a way to diff them in the future ? ). So if we want to be consistent with the events, we can set the state in `onDragStart` but I stated my concern about the usage of it above.

moklick:
Hey @furkankly 

I can't follow you. Which `isDraggable` check are you talking about? ` 

Now that I look at all the code again, I am thinking about moving all the handler logic to the `useDrag` hook itself (or to a `useNodeDrag` hook). The biggest challenge with the new implementation is dragging multiple nodes at once. This is currently not possible, because we are no longer using deltas. When the drag related code is at one place, it would simplify the implementation. My idea is to calculate all distances from the selected nodes and calculate the new positions on every drag.

furkankly:
> Hey @furkankly
> 
> I can't follow you. Which `isDraggable` check are you talking about? `
> 
> Now that I look at all the code again, I am thinking about moving all the handler logic to the `useDrag` hook itself (or to a `useNodeDrag` hook). The biggest challenge with the new implementation is dragging multiple nodes at once. This is currently not possible, because we are no longer using deltas. When the drag related code is at one place, it would simplify the implementation. My idea is to calculate all distances from the selected nodes and calculate the new positions on every drag.

I was referring to `onClick` being broken for the draggable nodes and how I fixed it [in this commit ](https://github.com/wbkd/react-flow/pull/2096/commits/0ef3ce81a848b84a4b38d92200635b31edbe11da) by removing the unnecessary `if (!isDraggable)` in the related handler. Could you re-review the latest version of this branch ? 

I was also referring to your [fix](https://github.com/wbkd/react-flow/pull/2072/commits/fa582daa7a710087ed76aec53d79c9313a3896f0) for the same problem and how it causes an another regression for us in the above comment.

All the comments I made until now were related to minor issues you mentioned while merging [furkankly:refactor/d3-drag-handle-selector](https://github.com/furkankly/react-flow/tree/refactor/d3-drag-handle-selector) to [wbkd:refactor/d3-drag](https://github.com/wbkd/react-flow/tree/refactor/d3-drag).

I still didn't have a chance to dig into the issue related to deltas - moving multiple nodes at once, I'll share my thoughts on that when I dig into it.

moklick (approved):

moklick:
Ah .. now I get it. I didn't see your latest commit. Thanks for your help! I'll push my changes regarding the multi selection in the next days. I think I have an OKish solution for this:

1. [dragstart] Project mouse position to RF coordinates
2. [dragstart] Calculate distances of selected nodes to projected mouse position

3. [drag] Project mouse position to RF coordinates
4. [drag] Calculate new positions of selected nodes with new mouse position, current node positions and distances 
5. [drag] Call action to update node positions

moklick:
The handlers in `wrapNode` look way cleaner now :) Thanks again 💯

PR #2116, opened 2022-05-01
added npm run build instruction

fawadasaurus:
I was getting errors without running starting the example but after I ran `npm run build`, I was able to start the example


moklick:
thanks @fawadasaurus

PR #2144, opened 2022-05-13
fix: fixed an issue where Handle component onMouseDown event could not be bind

hiyangguo:
This PR fixed an issue that when setting `onMuseDown` props on `<Handle/>` where in  custom a node, the all connect event is invalid.Because the out props `onMuseDown` is pass to `<div/>` by `...rest` . and It is overwrite  the inner onMouseDownHandler.


moklick:
good catch, thanks!

PR #2148, opened 2022-05-14
Fixed `setNodes` and `setEdges`

### LINKED ISSUE #2147: useReactFlow: `setNodes([])` does nothing
### Describe the Bug

Calling `setNodes([])` in an effect or event callback does nothing. It's only the `setNodes` from `useReactFlow`. The `setNodes` from `useNodesState()` works.

_No response_

As described above. This will happen in any application.

### Expected behavior

`setNodes([])` should remove all nodes.

### Screenshots or Videos

_No response_

### Platform

- OS: Win10
- Browser: na. Electron
- Version: "10.0.8

### Additional context

[This](https://github.com/wbkd/react-flow/blob/415fc77a3cfe17dc3d4a0cfa2e60766680593a65/src/hooks/useReactFlow.ts#L32) is the implementation for `setNodes`. In my debugging, `hasDefaultNodes` was `false` and `onNodesChange` was a function, so we take the `else if` branch.

```js
      const changes = nextNodes.map((node) => ({ item: node, type: 'reset' } as NodeResetChange<NodeData>));
      onNodesChange(changes);
```

Since my `nextNodes` is empty, `changes` will also be empty. Hence, [`applyChanges`](https://github.com/wbkd/react-flow/blob/415fc77a3cfe17dc3d4a0cfa2e60766680593a65/src/utils/changes.ts#L47) find no `c.type === 'reset'` change.

---

`setEdges` has the same bug.

RunDevelopment:
This fixes #2147.

I fixed by adding a check for empty change arrays. If the change array is empty, I create a new change array that removes all edges/nodes by id.

Hacky, yes, but it doesn't change any public types. My original idea was to either:
1. Change the `reset` type to `{ items: Node<NodeData>[]; type: 'reset'; }`,
1. Change the `reset` type to `{ item: Node<NodeData> | undefined; type: 'reset'; }`, or
1. Add a new change type `{ newItems: Node<NodeData>[]; type: 'reset-all'; }`

However, all of those are a breaking change because it would change the public API of `on{Nodes,Edges}Change` and `{Node,Edge}Change`.

---

Also, when running `npm test`, I just get this:

```ps
PS C:\Users\micha\Git\react-flow> npm test

> react-flow-renderer@10.2.2 test
> BROWSER=none npm run dev:wait test:chrome

Der Befehl "BROWSER" ist entweder falsch geschrieben oder
konnte nicht gefunden werden.
```

Since Microsoft decided that translated technical error messages are a good thing, I'll translate the last line: "The command "BROWSER" is either mistyped or could not be found."

Pretty sure that setting env variables like this is (ba)sh syntax, so it won't work in PowerShell. See [this](https://stackoverflow.com/a/27090755/7595472) for more info.

### DIFF
--- a/src/hooks/useReactFlow.ts
+++ b/src/hooks/useReactFlow.ts
@@ -2,7 +2,16 @@ import { useCallback } from 'react';
 
 import useViewportHelper from './useViewportHelper';
 import { useStoreApi } from '../store';
-import { ReactFlowInstance, Instance, NodeAddChange, EdgeAddChange, NodeResetChange, EdgeResetChange } from '../types';
+import {
+  ReactFlowInstance,
+  Instance,
+  NodeAddChange,
+  EdgeAddChange,
+  NodeResetChange,
+  EdgeResetChange,
+  NodeRemoveChange,
+  EdgeRemoveChange,
+} from '../types';
 
 export default function useReactFlow<NodeData = any, EdgeData = any>(): ReactFlowInstance<NodeData, EdgeData> {
   const { initialized: viewportInitialized, ...viewportHelperFunctions } = useViewportHelper();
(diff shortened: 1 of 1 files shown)


joeyballentine (approved):

joeyballentine:
For the environment variable thing, the way to fix this is using the cross-env package

moklick:
Thanks @RunDevelopment ! (we will add cross-env in the next release #2152)

PR #2149, opened 2022-05-15
Implemented allowPanOverNodes and selectionKeyCode as boolean

Kinark:
I'm not sure if this is the most performing way to do this, but I needed, so I did it.

This should address #1001 and #964, an updated version of #1156.

Basically now you can set `selectionKeyCode` as a boolean (true or false, I couldn't decide what was best) and that basically turns select mode on forever.

I also added the prop `allowPanOverNodes`, a boolean that, when turned on, sets an internal state in zustand that is used in the `<NodeRenderer />` and `<EdgeRenderer />` components, passed to the `<NodeComponent />` and `<EdgeComponent />` components and finally, in `wrapEdge.tsx` and `wrapNode.tsx` files, gets translated to a new class called `.unclickable`, which adds `pointer-events: none !important;` to the draggable elements.

Basically with a combination of the PR and three props: `panOnDrag={spacePressed}`, `allowPanOverNodes` and `selectionKeyCode={true}`, we get a Figma like experience.

It's a bit hacky and I had to change z-index values, but it works.

What do you think?

### DIFF
--- a/example/src/SaveRestore/save.css
+++ b/example/src/SaveRestore/save.css
@@ -2,7 +2,7 @@
   position: absolute;
   right: 10px;
   top: 10px;
-  z-index: 4;
+  z-index: 5;
   font-size: 12px;
 }
 
--- a/example/src/UpdateNode/updatenode.css
+++ b/example/src/UpdateNode/updatenode.css
@@ -2,7 +2,7 @@
   position: absolute;
   right: 10px;
   top: 10px;
-  z-index: 4;
+  z-index: 5;
   font-size: 12px;
 }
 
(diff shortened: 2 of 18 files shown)


moklick:
Hey @Kinark 

Thanks for your help!

Two notes:

1. I think you don't need the new `allowPanOverNodes` prop but you can use `elementsSelectable={false}` to achieve it. 
2. Why are you doing `typeof selectionKeyCode === 'boolean'` instead of `selectionKeyCode === true`?

Kinark:
1. I actually tried setting `elementsSelectable={false}` but the node kept getting in the way of the dragging and it just stopped dragging.
2. Something wrong was happening, can't remember why.

But I'll rewrite this for the v10

moklick:
I added a discussion for this topic #2167 . Let's discuss the API first before we continue here.

moklick:
Hey @Kinark 

are you still working on this? Some things changed with v11. The `allowPanOverNodes` is not necessary anymore. It's the new default behaviour when nodes are not draggable.

moklick:
@nthypes working on this here: #2662

PR #2152, opened 2022-05-16
Next release/v10.3

### LINKED ISSUE #1006: Question about userSelection
Hey!
First of, thanks for this amazing package, I really enjoy it!

Now I'm trying manually set user selection (blue rectangle) because it's easy to work with it to get new positions of dragged nodes.
I only found an option to setSelectedNodes (which is great, but for other cases)
It doesn't fit in this case because when we move selected nodes (e.g. selected by ctrl + click, without blue rectangle) only onNodeDrag fires. And It has updated position for only one node so we have to calculate others

1. How can I manually set user selection to trigger onSelectionDrag?
2. Or how can I easily get new positions of all selected moved nodes?

Thanks!

### LINKED ISSUE #1032: Unexpected behavior when dragging nodes after area selection
Observed in React Flow v9.3.2.

Expected behavior: all nodes move in sync like when individually dragged or dragged after selecting with multiple clicks.

What actually happens (the nodes become stuck to the mouse until another click):
react-flow-sticky

This can be avoided by first clicking on a node once and then dragging, but that's not intuitive at all.

### LINKED ISSUE #1314: Drag stop listeners aren't working if multiple nodes are moved

### LINKED ISSUE #1395: Dragging a selection does not adhere to `snapGrid`
Making an area selection and dragging it around doesn't cause it to snap to the specified grid size, hence on release, the selection can be dropped in misaligned positions.

### LINKED ISSUE #1462: Deselect the selected element when clicking it with the multi selection key pressed
I am using `multiSelectionKeyCode` to select multiple elements, so far so good. Now I would also expect that clicking an element that was already selected while the key is pressed would deselect it. Currently, it does nothing (the element remains selected).

I'm not sure what's the default behaviour on Mac, but this is common behaviour on Windows at least, and also on most tools that handle multiple selection, such as Google Drive.

### LINKED ISSUE #1579: MultiSelection-Drag breaks Snap-To-Grid
It seems like selecting multiple Nodes (with shift + click + drag) and then moving them across the pane breaks out of the grid thats set via snap-to-grid, especially when zoomed in.

I manually set the elements back on to the grid in the onSelectionDragStop function, which shows the missplacement the nodes would have normally. 
snap_to_grid_error

### LINKED ISSUE #1975: onDragStop always triggle onClick in custom node component.
### Describe the Bug

```js

// index.js
const nodeTyps = {
  custom: CustomNode
}
<ReactFlow nodesDraggable />


// CustomNode.js

// handleCustomClick always trigger by onDragStop
<div onClick={handleCustomClick}></div>
 
```

https://codesandbox.io/s/react-flow-drag-jled0w

1. drag the node 2
2. see "node click" in console below

### Expected behavior

onNodeDragStop not trigger onClick.

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 91.1]

### Additional context

It works but I don't know if there are side effects

```js
// src/components/Nodes/wrapNode.tsx
...
setTimeout(() => {
      updateNodePosition({
            id,
            dragging: false,
      });
}, 300);
...
```

### LINKED ISSUE #1999: Dragging in controlled mode does not work if the props are not updated synchronously
### Describe the Bug

There are two related bug:
1. I use ClojureScript for my project with reagent, a react wrapper. It does some extra optimisation and therefore do async renders (https://reagent-project.github.io/news/reagent-is-async.html). It might be the same in React 18 but I'm not sure at all (https://github.com/reactwg/react-18/discussions/21). In this context, the drag is not working. The internal nodes are not yet up to date when another drag event occurs and so the delta position is applied to the old position. This causes the previous drag event to be skipped. It may also occur in JS/React when something asynchrone is done before setting the new props (sending the drag event to a websocket server, animating the position... etc.).
2. When using the sub flow feature, dragging outside the parent create an offset between the node and the mouse. Because the drag is relative the offset remains even if the mouse enter again into the parent.

https://codesandbox.io/s/sandpack-project-forked-nxurzp?file=/App.js:649-729

For the first bug:
1. Use react-flow in a clojurescript/reagent project
2. Drag any node fast

Alternative:
1. Put the setNodes (in the onNodeChanges callback) in a setTimeout(x, 10) : `onNodesChange={(...args) => setTimeout(() => onNodesChange(...args), 10)}`
2. Drag any node fast

For the second one:
1. Open sub flows example: https://reactflow.dev/docs/examples/sub-flows/
5. Drag the node B1 in the group B outside of it
6. Drag back (without releasing) the mouse in group B
7. The node has an offset with the mouse

### Expected behavior

The node must continue to follow the mouse no matter what (but keep respecting the bounds).

### Screenshots or Videos

_No response_

### Platform

- OS: Linux
- Browser: Chrome
- Version: 97

### Additional context

One of the main solution to solve these two related bugs is to rely on absolute mouse pos instead of adding delta pos each frame. The position will not depend on the previous pos of the props and the two bugs will be fixed.

Else, it is possible to provide the delta pos and the clamp function to the onNodeChanges callback in controlled mode.

### LINKED ISSUE #2033: Fallback node render does not work correctly.
### Describe the Bug

If declaring a node `type` that does not exist, the console will report that no renderer for this node type exists and inform you that the default layout is being used. 

eg: 

```
Node type "foo" not found. Using fallback type "default".
```

However, the behaviour of such nodes does not match the actual default node behaviour. 

https://codesandbox.io/s/bold-voice-84d6jv?file=/src/Flow.js

See linked sandbox: https://codesandbox.io/s/bold-voice-84d6jv?file=/src/Flow.js

Note that lines 24 and 25 show the same node with the same position, the difference is that the line 25 node has type 'foo'. 

```js
// Standard node
 { 
    id: "2",
    data: { label: "Node 2" },
    position: { x: 250, y: 100 }
 },
 
// Node with unrecognised 'type' property 
{
    id: "2",
    type: "foo",
    data: { label: "Node 2" },
    position: { x: 250, y: 100 }
  },
```

### Expected behavior

We would expect that the nodes should display the exact same way, but they don't. The node with type 'foo' has no border, and doesn't not respect positioning the same way. 

### Screenshots or Videos

Actual default behaviour: 
Screen Shot 2022-03-31 at 11 53 16 am

Fallback default behaviour: 
Screen Shot 2022-03-31 at 11 53 00 am

### Platform

- Version: [e.g. 91.1]

### Additional context

_No response_

### LINKED ISSUE #2034: dragHandle is not working with onNodeClick handler
### Describe the Bug

Drag Handle and Node Click Handler is working separately but when used together node click handler is not working on the node but only at draggable content area.

https://codesandbox.io/s/sandpack-project-forked-3bwpd7

1. Go to Sandbox application
2. Click on Input or Output Node
3. Console log has shown 
4. Click the custom node
5. No console log is written for click handler
6. Click drag handle on custom node
7. Console log is shown

### Expected behavior

All Nodes should be logged to console when clicked
Only draggable handle allows click events

### Screenshots or Videos

https://www.loom.com/share/f7381ae5aac74cfe927bfdfacf1533ce

### Platform

- OS: macOS 12
- Browser: Chrome
- Version: 99

### Additional context

_No response_

### LINKED ISSUE #2049: `onConnectEnd` getting triggered twice
### Describe the Bug

Instead of taking an edge out of an handle if we just click on a handle then the `onConnectStart` and `onConnectEnd` callbacks get triggered more than once

https://codesandbox.io/s/sandpack-project-forked-bgb4z0?file=/App.js

1. Click on a handle of node
2. Do not take out an edge out of the handle
3. The callbacks will get triggered more than once

### Expected behavior

The callbacks should get executed only once as it was in the earlier version of React Flow.

### Screenshots or Videos

https://user-images.githubusercontent.com/30779692/161759635-9c02d111-feec-4739-8cce-3bd12b6e9164.mov

### Platform

- OS: macOS
- Browser: Chrome
- Version: 99.0

### Additional context

_No response_

### LINKED ISSUE #2119: Major lag when dragging nodes with react 18 / v10.2
### Describe the Bug

Upgrading to React 18 and v10.2 of React Flow introduced noticeable, major lag when dragging nodes, as can be seen in the videos below.

In this example, `onNodesChange` and `onEdgesChange` are implemented as the default (calling `setNodes` and `setEdges`) and `onNodesDrag` is **not** implemented, so there shouldn't be any major logic going on on drag that could introduce this lag. I'm happy to add any code or context you might want but I'm not sure what else could be helpful here.

No other visible regressions were found when doing this update.

_No response_

1. Update react and react-dom to v18 and update react-flow-renderer to v10.2

### Expected behavior

I expected dragging nodes to be as fluent as when using v17/v10.1.

### Screenshots or Videos

React 17 / v10.1:

https://user-images.githubusercontent.com/24462089/166427288-07ad67ef-abd2-4973-85db-3e9c7d46456a.mp4

React 18 / v10.2:

https://user-images.githubusercontent.com/24462089/166427308-5bf9eaa2-eec5-420f-8f1f-436f5e2f4d74.mp4

### Platform

- OS: Windows 11
- Browser: Chrome
- Version: 100.0

### Additional context

_No response_

moklick:
This release contains the following changes:

* **replace react-draggable with d3-drag** (which is already used by d3-zoom, that is used for pan and zoom)
* **rewrite drag logic** in order to work with absolute positions instead of deltas. This works way better for slower devices and also with React18.
* **pass multiple nodes to drag handlers**: When you drag multiple nodes, we now pass an array of nodes as a third parameter to `onNodeDragStart(event: MouseEvent, node: Node, nodes: Node[])`, `onNodeDrag(event: MouseEvent, node: Node, nodes: Node[])` and `onNodeDragStop(event: MouseEvent, node: Node, nodes: Node[])`
* **Selecting nodes**: simplified code, deselect selected node when clicking it with multi selection key pressed
* **Connection line**: Show connection line always on top of nodes, add `connectionLineContainerStyle` prop
* **on click connection**: separate events for connections that are created by clicking - `onClickConnectStart`, `onClickConnectStop` 
* **lots of minor refactorings** (cleanup styles, remove unnecessary useCallbacks, cleanup types)

You can try the next release via:

```npm install react-flow-renderer@next```

closes #2119, closes #2034, closes #2033 , closes #2049, closes #1579, closes #1395, closes #1314, closes #1032, closes #1006 , closes #1462, closes #1975, closes #1999


schadocalex:
Also fixed #1999

Issue #810, opened 2020-12-30
ConnectionLine zIndex Problem


andrewdkim:
Hello!
First off, thank you for making and actively updating this wonderful project.  It's absolutely one of the most amazing libraries I have ever worked with. 

I'm currently running into a problem where I want to render a custom made Connection Line on top of the nodes. 

Here's the current situation:
![pic1](https://user-images.githubusercontent.com/46148447/103334426-38b96000-4a3f-11eb-8592-05e4c7d40ea8.png)

What I want to achieve: 
![pic1 2](https://user-images.githubusercontent.com/46148447/103334443-4c64c680-4a3f-11eb-870b-9787a8f4794b.png)

I'm aware that z index is not an option for SVG, but is there a way to render the SVG on top of other nodes in the graph? Thanks!

RickeyWard:
Edges are rendered as on SVG on a "layer" behind the nodes. If you wanted the edges to all be in front of the nodes, override the z-index of the whole svg.

```css
.react-flow__edges {
  z-index: 4 !important; /* normally 2, as nodes are 3 */
}
```

This could cause issues, but in a quick example it did work. From looking at the source code there's no library interface that allows you to put edges in front of nodes.

moklick:
Thanks @RickeyWard

PR #2159, opened 2022-05-18
Added `connectionLineContainerStyle` property

RunDevelopment:
This implements solution 3 from #2158.

The connection line will now always have an exclusive `<svg>` element. This makes the implementation easier. I kept the old class name to stay backward compatible. The fix suggested in #810 will continue to work.

### DIFF
--- a/src/container/EdgeRenderer/index.tsx
+++ b/src/container/EdgeRenderer/index.tsx
@@ -93,7 +93,7 @@ const EdgeRenderer = (props: EdgeRendererProps) => {
           style={{ zIndex: level }}
           width={width}
           height={height}
-          className="react-flow__edges react-flow__connectionline react-flow__container"
+          className="react-flow__edges react-flow__container"
         >
           {isMaxLevel && <MarkerDefinitions defaultColor={defaultMarkerColor} />}
           <g>
@@ -196,7 +196,7 @@ const EdgeRenderer = (props: EdgeRendererProps) => {
         style={connectionLineContainerStyle}
         width={width}
         height={height}
-        className="react-flow__edges react-flow__container"
+        className="react-flow__edges react-flow__connectionline react-flow__container"
       >
         <g>
           <ConnectionLine
--- a/src/container/EdgeRenderer/index.tsx
+++ b/src/container/EdgeRenderer/index.tsx
@@ -73,7 +73,13 @@ const EdgeRenderer = (props: EdgeRendererProps) => {
     return null;
   }
 
-  const { connectionLineType, defaultMarkerColor, connectionLineStyle, connectionLineComponent, connectionLineContainerStyle } = props;
+  const {
+    connectionLineType,
+    defaultMarkerColor,
+    connectionLineStyle,
+    connectionLineComponent,
+    connectionLineContainerStyle,
+  } = props;
   const renderConnectionLine = connectionNodeId && connectionHandleType;
 
   return (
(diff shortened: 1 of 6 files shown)


joeyballentine (approved):

moklick:
Thanks a lot. This looks good! Can you change the class name of the container from `react-flow__edges` to `react-flow__connectionline` and do your PR against `next-release/v10.3`. Then we have all the updates bundled.

RunDevelopment:
> Can you change the class name of the container from `react-flow__edges` to `react-flow__connectionline`

That would be a breaking change. All CSS that assumes that the connection line element is inside a `react-flow__edges` container will break. E.g. the fix from #810 would no longer work.

However, I could add this class alongside `react-flow__edges`.

moklick:
Good point! An additional class name is fine!

joeyballentine (approved):

src/container/EdgeRenderer/index.tsx
```diff
@@ -179,22 +189,29 @@ const EdgeRenderer = (props: EdgeRendererProps) => {
...
           </g>
         </svg>
       ))}
+      {renderConnectionLine && <svg
+        style={connectionLineContainerStyle}
+        width={width}
+        height={height}
+        className="react-flow__edges react-flow__container"
+      >
+        <g>
```
moklick, line 201: this <g> is not necessary

src/container/EdgeRenderer/index.tsx
```diff
@@ -75,6 +76,15 @@ const EdgeRenderer = (props: EdgeRendererProps) => {
   const { connectionLineType, defaultMarkerColor, connectionLineStyle, connectionLineComponent } = props;
   const renderConnectionLine = connectionNodeId && connectionHandleType;
 
+  let { connectionLineContainerStyle } = props
```
moklick, line 79: I think we can simplify this and change the default behaviour a bit. The connection line should always be above the other elements. I would set the z-index via CSS to 1001 (so that it's always above all nodes) and just pass the container style to the svg container. If you need a higher z-index, you can use the `connectionLineContainerStyle`.

moklick (changes requested):

src/container/EdgeRenderer/index.tsx
RunDevelopment: It is necessary for backwards compatibility. People might have CSS selectors that expect the `<g>` to be there.

src/container/EdgeRenderer/index.tsx
```diff
@@ -83,7 +93,7 @@ const EdgeRenderer = (props: EdgeRendererProps) => {
           style={{ zIndex: level }}
           width={width}
           height={height}
-          className="react-flow__edges react-flow__container"
+          className="react-flow__edges react-flow__connectionline react-flow__container"
```
RunDevelopment, line 96: I added the connection line class to the wrong SVG element... Let me fix that.

src/container/EdgeRenderer/index.tsx
RunDevelopment: Let's see if I understood this correctly.

You want me to remove the zIndex default logic in line 80-87 and replace with a simple CSS rule (in a CSS file) that sets the zIndex to 1001, right?

src/container/EdgeRenderer/index.tsx
moklick: Thanks for the changes! The connection line is already wrapped with a `<g>`.. I would say that we can neglect this one.

moklick:
Thanks! Testable with `npm install react-flow-renderer@next`.

Issue #2183, opened 2022-05-27
Node dragging property doesn't return to false


jech33:
### Describe the Bug

When dragging a node the dragging property is set to true which is ok, but when the user stops dragging the node, the dragging property remains in true, doesn't return to its original value (false)

### Steps to Reproduce the Bug or Issue

Open react dev tools
Select a node in dev tools
Start dragging that node and check its dragging value
Stop dragging the node and check the dragging value

### Expected behavior

As a user I expected the dragging property to be false when a node is not being dragged, but if I drag it and then stop, the dragging property remains whit a true value.

### Screenshots or Videos

![image](https://user-images.githubusercontent.com/67987594/170793127-b6c0d7c9-0df6-454f-89d8-23daa705130f.png)

### Platform

- OS: [Windows 11]
- Browser: [Vivaldi]
- Version: [5.2.2623.48 (Stable channel) (64-bit)]

moklick:
fix released in 10.3.3

PR #2184, opened 2022-05-27
fix(drag): set `dragging` to false on drag end

bcakmakoglu:
# What's changed?

* Set `dragging` to `false` on drag end event

# Fixes
* #2183

### DIFF
--- a/src/hooks/useDrag/index.ts
+++ b/src/hooks/useDrag/index.ts
@@ -118,6 +118,7 @@ function useDrag({
             }
 
             event.on('end', (event) => {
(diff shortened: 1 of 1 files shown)


bcakmakoglu:
Maybe I missed it but it seems like the `dragging` value is never set back to false, so this PR aims to correct that.

jech33:
Thanks for the fix [bcakmakoglu](https://github.com/bcakmakoglu)!

moklick:
Thanks @bcakmakoglu !

PR #2279, opened 2022-07-11
Major Release v11

### LINKED ISSUE #675: [Bug]: Scrolling with panOnScroll enabled, trigger multiple onMoveEnd event
When panOnScroll is enable, a single scroll on react-flow trigger multiple onMoveEnd events, the desired behaviour is that onMoveEnd should only be trigger when moving is finished for a single user scroll action.

### LINKED ISSUE #1033: Support for accessibility
Hi,

I really like the React-Flow library, but I wonder if it could be made more accessible by adding the appropriate [WAI-aria roles and attributes](https://developer.mozilla.org/en-US/docs/Learn/Accessibility/WAI-ARIA_basic)?

To help with that, I would also suggest the use of [React Testing Library](https://testing-library.com/), which [encourages the use of queries that are accessible to everyone](https://testing-library.com/docs/queries/about#priority) for targeting elements, thus ensuring the tested components are actually accessible.

Is that something you would be willing to consider?

### LINKED ISSUE #1424: onNodeDragStart without onNodeDragStop
Hi,

I have encountered a situation where `onNodeDragStop` is not getting called after `onNodeDragStart` is called when the user did not move the node out of position before releasing the mouse left button.  I believe it's due to the following logic:

https://github.com/wbkd/react-flow/blob/64d1913356d171fd2c4303c918c3e375f57cbee7/src/components/Nodes/wrapNode.tsx#L173-L195

Is it possible to update the logic as follows to ensure `onNodeDragStop` is not omitted:

```
 const onDragStop = useCallback( 
   (event: DraggableEvent) => { 
     // onDragStop also gets called when user just clicks on a node. 
     // Because of that we set dragging to true inside the onDrag handler and handle the click here 
     if (!isDragging) { 
       if (isSelectable && !selectNodesOnDrag && !selected) { 
         addSelectedElements(node); 
       } 
  
       onClick?.(event as MouseEvent, node); 

     } else {
       updateNodePosDiff({ 
         id: node.id, 
         isDragging: false, 
       });
    } 
     onNodeDragStop?.(event as MouseEvent, node); 
   }, 
   [node, isSelectable, selectNodesOnDrag, onClick, onNodeDragStop, isDragging, selected] 
 ); 
```

### LINKED ISSUE #1457: Drag and click nodes
Hello dear react flow team,
we are currently working on the interaction to move the diagram for our users. There are multiple ways to move the position of the diagram and we are trying to find the most native implementation possible. Therefor our users can interact with our diagram on mobile as well as on desktop devices. To move the diagram around it is possible to drag the pane and move the diagram where desired. The only thing that we cannot yet achieve, is the dragging of the diagram if you touch a node instead of the pane. For example, if a user interacts on a mobile phone, it is common that you can drag everywhere on a draggle area. By now our users need to avoid, trying moving the diagram by accidentally selecting a node and can instead only touch in between nodes. I am aware that there is an option called `elementsSelectable` that we can set to `false`. This would be the perfect behaviour if we only had representational nodes. In fact our interaction on click of the nodes is essential to our application. We show the data of this node to view and change it accordingly. Is there somehow a way that it would be possible to track down an `onClick` as well as an `onDragStart` event for nodes to get the desired effect?
Thanks in advance

### LINKED ISSUE #1679: Propose to make node data NOT optional
The current data type for Node is

image

This means even when you assign `T` such as `Node<ElementData>` and know in your program all Nodes would have `T`.
it will still show the data as possibly being undefined.

image

Because T defaults to `any` which could be undefined.

Changing `data?: T` to `data: T` would still allow setting `data: undefined` on new element objects if needed.

Side note, I think the default renderer for nodes actually requires the data to be `{ label: 'string }` so setting undefined on Nodes might crash application which means this is another reason it should not be typed as optional since it's actually required.

### LINKED ISSUE #2108: @testing-library/react Error: Uncaught [TypeError: Cannot redefine property: onresize]
### Describe the Bug

When running unit tests using jest & @testing-library/react, I get the following error when trying to render a component containing a ReactFlow instance:

Error: Uncaught [TypeError: Cannot redefine property: onresize]

_No response_

```jsx
import { render } from '@testing-library/react';

test('example', () => {
  window.ResizeObserver = jest.fn(() => ({
      observe: jest.fn(),
      unobserve: jest.fn(),
      disconnect: jest.fn(),
  }));

  render(
    <div style={{ height: 1200, width: 800 }}>
      <ComponentContainingReactFlow />
    </div>,
  );
});
```

### Expected behavior

The test should not throw an error.

### Screenshots or Videos

_No response_

### Platform

- OS: macOS
- Browser: jsdom
- Version: 10.1.2

### Additional context

_No response_

### LINKED ISSUE #2278: Create Monorepo with separate packages for the core, MiniMap, Controls and Background
we don't want the library to become incredibly large and have the ability to release individual packages.

### LINKED ISSUE #2285: onEdgeUpdate: hide edge while updating it
It looks weird that we are showing the original edge while updating it. We can set it to `visibility: hidden` while we are updating the edge.

### LINKED ISSUE #2288: Add boolean prop to remove attribution, delete proOptions?
The attribution is causing confusion (how to remove it, when to show it, etc..). We want to clear this up in the next major release.

### LINKED ISSUE #2299: Remove of `onConnectStop`
For historical reasons we have `onConnectStop` (triggered before `onConnect`) and `onConnectEnd` (triggered after `onConnect`). In v11 we should get rid of `onConnectStop`.

### LINKED ISSUE #2300: Cleanup connection line: remove duplicate `sourceNode` and `sourceHandle` props
Remove duplicate `sourceNode` and `sourceHandle` props and only keep `from...` and `to...` props.

### LINKED ISSUE #2301: Nodes jumping on start when using `fitView`
### Describe the Bug

When you load a flow with `fitView` you might see the nodes jumping. 

1. check the local examples

### Expected behavior

no jumping of nodes

### Platform

- OS: macOS, Windows, Linux
- Browser: Chrome, Safari, Firefox
- Version: 10.3.10

### Additional context

_No response_

### LINKED ISSUE #2312: Add `onPaneMouseEnter`, `onPaneMouseMove` and `onPaneMouseLeave`
It's not possible to add a `onMouseMove` event handler to the React Flow pane without using `addEventListener`. We should add `onPaneMouseEnter`, `onPaneMouseMove` and `onPaneMouseLeave` to the props.

### LINKED ISSUE #2313: merge `defaultPosition` and `defaultZoom` to `defaultViewport`
It's not clear what `defaultPosition` does. Merging `defaultPosition` and `defaultZoom` to `defaultViewport` is better understandable and a preparation for a controlled `viewport` prop that we might add later.

### LINKED ISSUE #2320: New background variant: Cross
It would be nice if we could add a new "cross" variant for the background component

### LINKED ISSUE #2370: panOnDrag on elements not working when onNodeClick is also active
### Describe the Bug

I'm currently writing a page that requires to click on elements while also allowing to drag the view even from inside those elements. This is because I'm using some pretty large elements, therefore not being able to drag the view by dragging from inside the node makes the page very difficult to navigate. 

https://codesandbox.io/s/jk3jng?file=/App.js:2810-2821&from-sandpack=true

1. Enable both panOnDrag and capture elementOnClick, disable everything else. 
2. Try to drag the view by dragging from inside one of the nodes (start the drag by pressing the left mouse key inside a node, then move the mouse without releasing the left mouse key). 
3. The view won't move, even though it's not activating the onClickFunction either (clicking the node activates the console.log onNodeClick function, but dragging from inside the node neither activates the onNodeClick function nor does it drag the view). 

### Expected behavior

The expected behavior is allowing to drag the view by dragging from inside an element. 

### Screenshots or Videos

_No response_

### Platform

- OS: Windows 11
- Browser: Chrome
- Version: 103

### Additional context

_No response_

### LINKED ISSUE #2384: `nodeExtent` doesn't work for initial positions
### Describe the Bug

```js
const nodeExtent: CoordinateExtent = [
  [0, 0],
  [50, 50],
];

const nodes = [{
  id: '1',
  position: { x: 200, y: 200 },
  data: { label: '1' },
}];

<ReactFlow
  nodes={nodes} 
  nodeExtent={nodeExtent}
/>
```

this should place the node at x=50, y=50 but it ignores the nodeExtent and places it at x=200, y=200

_No response_

1. use code above

### Expected behavior

Place nodes inside the node extent

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 91.1]

### Additional context

_No response_

### LINKED ISSUE #2438: Dragging of nodes group doesn't work if node mouse handler is set
### Describe the Bug

If child node (node that has `parentNode` field) has `draggable` property as `false` and node mouse event handler (e.g. `onNodeMouseLeave`) is set, it's impossible to  drag group of nodes if you pressed left button over child node.

https://codesandbox.io/s/sandpack-project-forked-5cgf76?file=/App.js

1.  Create group of nodes (via parentNode).
2. Make child nodes undraggable (`draggable: false`).
3. Add node mouse event handler for ReactFlow (e.g. `onNodeMouseLeave`).
4. Group of nodes can't be dragged if you pressed left button over child nodes (but works as expected if left button is pressed over parent node). 

### Expected behavior

It should be possible to drag a group of nodes with mouse event handler and undraggable child nodes.

### Screenshots or Videos

https://user-images.githubusercontent.com/64694520/191267295-7ed39dca-1fd7-4bbb-859b-f545f44729bd.mov

### Platform

- OS: [macOS]
- Browser: [Chrome]
- Version: [10.3.14]

### Additional context

_No response_

chrtze:
This is the PR for the next major release. You can find all informations about new features and breaking changes here:

https://github.com/wbkd/react-flow/discussions/2307#discussioncomment-3222751

fixes #2278, fixes #2285, fixes #2288, fixes #2299, fixes #2300, fixes #2301, fixes #1679, fixes #1424, fixes #2312, fixes #2313, fixes #2320, fixes #1033, fixes #2108, fixes #2384, fixes #675, fixes #1457, fixes #2370, fixes #2438


OrhanTozan:
Is there an ETA when this will be merged?

PR #2324, opened 2022-07-28
fix(validation): Remove css-classes if mouse moves directly to different handle.

fbieler:
If two handles are next to each other or overlap it is possible for the
mouse-cursor to move from one handle to the other without a
mouse-move-event being fired where isHoveringHandle is false.

Thus, the valid and connecting css-classes were not removed from the
handle.

This commit fixes this.

To reproduce you can rearrange the nodes in the validation example like so:
image
And then drag a connection from the left node across the three overlapping handles.

### DIFF
--- a/src/components/Handle/handler.ts
+++ b/src/components/Handle/handler.ts
@@ -153,6 +153,7 @@ export function handleMouseDown({
     }
 
     if (connection.source !== connection.target && elementBelow) {
+      resetRecentHandle(recentHoveredHandle);
       recentHoveredHandle = elementBelow;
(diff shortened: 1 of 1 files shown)


joeyballentine (approved):

joeyballentine:
I had to decrease the vertical hoverable areas of my handles because of this issue. Nice work fixing it!

moklick:
@fbieler thanks for your help!
@joeyballentine thanks for reviewing this! 

I will check it!

moklick:
Thanks again @fbieler . Released in v10.3.13

PR #2340, opened 2022-08-04
add initial state for zoomOnScroll & panOnScroll

lonexw:


moklick:
Hey @lonexw, thanks for your PR. Why do you think that this is necessary? We are setting a default in the component where it's used https://github.com/wbkd/react-flow/blob/main/src/container/ZoomPane/index.tsx#L43

PR #2410, opened 2022-09-07
Snap node to grid relative to 0,0 instead start position of node

wardoost:
I have an app where the user is able to toggle snapping and noticed some potentially undesirable behaviour. With the current implementation snapping is relative to the start position of the node. I think it makes more sense to snap relative to the centre point of the diagram so when snapping is enabled nodes can be neatly aligned relative to each other.

A quick example, let's say a diagram has a `gridSize=[10, 10]`. With snapping disabled `snapToGrid={false}` the user moves a node to `[7, 7]`. Then the user enables snapping `snapToGrid={true}`. With the current implementation when the user moves the same node it will snap to `[17, 17]`, `[27, 27]` and so on. I think it makes more sense the node would snap to `[10, 10]`, `[20, 20]` or `[30, 30]`.

My proposed code changes fix this though it's a quick solution but I'm keen to rethink how to implement this if the maintainers of this library think this is a good change 🙂

### DIFF
--- a/src/hooks/useDrag/index.ts
+++ b/src/hooks/useDrag/index.ts
@@ -96,14 +96,14 @@ function useDrag({
             }
           })
           .on('drag', (event: UseDragEvent) => {
-            const { updateNodePositions, nodeInternals, nodeExtent, onNodeDrag, onSelectionDrag } = store.getState();
+            const { updateNodePositions, snapToGrid, snapGrid, nodeInternals, nodeExtent, onNodeDrag, onSelectionDrag } = store.getState();
             const pointerPos = getPointerPosition(event);
 
             // skip events without movement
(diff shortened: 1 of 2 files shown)


bcakmakoglu:
#2415 

Didn't realize there's a PR for the issue - just linking for visibility.

moklick:
Hey @wardoost 

thanks for your PR! Everything looks good to me :)

moklick:
released in released in v10.3.17

WillBaker2017:
Hello, @wardoost @moklick @bcakmakoglu 
I just submitted a new feature request regarding this merge. #2440 
I was using the old snapGrid and would like a toggle between the two (0, 0 and reverencing current x, y position)

PR #2416, opened 2022-09-08
Use event.composedPath for getting event target (keydown shadow DOM fix)

jwm0:
This PR fixes a problem where keystrokes are not being registered when originating from shadow DOM inputs. Current react-flow implementation attaches a global event listener for `keydown` event and calls `preventDefault()` on it if the following 2 conditions are met:
1. The key name is registered (using props like `deleteKeyCode`, `multiSelectionKeyCode` etc.)
2. **`isInputDOMNode` returns true** 

[This article ](https://tinytip.co/tips/js-event-composed-path/)explains why the current implementation of `isInputDOMNode` doesn't work as well as shows a solution which is used in this PR

Reproducible demo sandbox: https://codesandbox.io/s/react-flow-with-shadow-dom-lymfwm?file=/index.js

### DIFF
--- a/src/hooks/useKeyPress.ts
+++ b/src/hooks/useKeyPress.ts
@@ -108,7 +108,7 @@ function useKeyOrCode(eventCode: string, keysToWatch: KeyCode): KeyOrCode {
 }
 
 function isInputDOMNode(event: KeyboardEvent): boolean {
-  const target = event.target as HTMLElement;
+  const target = event.composedPath()[0] as HTMLElement;
 
   return ['INPUT', 'SELECT', 'TEXTAREA'].includes(target?.nodeName) || target?.hasAttribute('contenteditable');
 }


moklick:
Thanks! I didn't know about this. I will have a look.

moklick:
released in v10.3.17

jwm0:
@moklick would you be up to release this fix in v9 as well if I prepare a ported PR for that? Our team is currently on v9 and it would be great if we didn't have to patch this ourselves locally

moklick:
We don't do fixes for v9. v10 was released six month ago and we are moving to v11 next week. Can't you upgrade to v10 or better v11 ?

jwm0:
It's in the pipeline but not sure how much longer it'll take. Anyway, we can leave without it for now and hopefully can migrate sooner than we need this functionality. Thanks

PR #2430, opened 2022-09-17
fix: (example) Make the newly added node position `more accurate` in DragNDrop example

pengfu:
In DragNDrop example as mentioned above, nodes are added to where the mouse is released, however,  we are able to drag from anywhere in the sidebar node, causing the position of the added node to be visually biased. The solution is to record the offset of the drag point from sidebar node when dragging, and apply this offset when dropped.


moklick:
Thanks for you solution, but this makes the example more complicated. These examples are just for testing and on the websites we try to keep the examples as simple as possible.

PR #2437, opened 2022-09-20
feat(store): cancelConnection method created

Alireza29675:
This is a way to **cancel the dragging connection** if needed.

For example, in our project, we need to cancel the connection when <kbd>ESC</kbd> is pressed. It can be done like this:

```ts
const { cancelConnection } = useStore();

useEscPress(() => cancelConnection(), [cancelConnection]);
```

@moklick, if this is fine in general:

1. is there any specific example in mind to add this one to, or should we make a separate example for it?
2. What is the e2e and unit testing conventions in your team?


moklick:
Hey @Alireza29675 

thanks for your PR :) I think it's a good addition, but could you do it against the v11 branch? I don't want to add any new features to v10. More over I am not sure what would be the best way to expose this to the user. Having it as a store action only, it would be a bit hidden, but also fine.. However it would be OK for me to start with this kind of implementation even if it's a bit hidden. 

1. A new cancel connection example would be best
2. At least one unit test would be great here

Alireza29675:
@moklick Hei Moritz!

Sure. We can implement it in the store and use the method for other internal implementations. I'll move it to `v11`.

moklick:
k, thanks!

PR #2432, opened 2022-09-17
feature(deleteSelectedElements): add `deleteSelectedElements` method

pengfu:
Thanks to the great job of react-flow, we can delete selected elements by pressing the `deleteKeyCode` , however, in some cases we may also want to delete them by clicking a delete button in the toolbar.  This PR export `deleteSelectedElements` method in useReactFlow hook so that all people can use if they need.

PR #2451, opened 2022-09-27
feat(deleteElements): add `deleteElements` method

pengfu:
hi @moklick ,
As mentioned in [https://github.com/wbkd/react-flow/pull/2432](https://github.com/wbkd/react-flow/pull/2432), I have just redo the PR againest v11 branch.
Reactflow is an excellent library, and I really appreciate your excellent job!!

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@
 
 A highly customizable React component for building interactive graphs and node-based editors.
 
-[🚀 Getting Started](https://reactflow.dev/docs/getting-started/installation) | [📖 Documentation](https://reactflow.dev/docs/api/react-flow-props) | [📺 Examples](https://reactflow.dev/docs/examples/overview) | [☎️ Discord](https://discord.gg/RVmnytFmGW) | [💎 React Flow Pro](https://pro.reactflow.dev/pricing)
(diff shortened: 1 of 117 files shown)


moklick:
now it's `main` (we already merged v11).. 

Could you run it locally now? I remember that you had some issues with the new repo structure.

pengfu:
Yes, I think your idea is better. 
Then what about the the method typed like `deleteElements(nodeIds, edgeIds): (string[], string[])=>void` , `nodeIds` and `edgeIds`  may represent the nodes or edges that user initially want to remove ? 
I'll try to do this change right away

moklick:
I would suggest an API like this one:

```js
type DeleteElementsOptions = {
  nodes?: (Partial<Node> & { id: Node['id'] })[],
  edges?: (Partial<Edge> & { id: Edge['id'] })[]
};

const deleteElements = ({ nodes, edges }: DeleteElementsOptions) => void
```

So instead of two separated params I would pass an object with optional node and edge arrays that could also just be id arrays.

pengfu:
hey @moklick , your suggesiton seems better ,  I have just tried to do the change and gave an example in usereactflow part.

pengfu:
Besides, I have found out a strange behavior that occurs before my commits yet I am not sure if it is an issue.
In useReactFlow example, as described on [https://www.techiediaries.com/react-18-useeffect/](https://www.techiediaries.com/react-18-useeffect/), useEffect may called twice in development strict mode, 
this lead to `addEdges` called twice
`useEffect(() => {
    addEdges({ id: 'e3-4', source: '3', target: '4' });
  }, [addEdges]);`
when we tried to delete Node 3 ,  finnaly what shows on canvas is:

image

moklick:
closed in favour of #2536

PR #2452, opened 2022-09-28
Add ability to customise the UI to be more aligned to other whiteboarding tools

jackfishwick:
This PR adds the following features:

- Selection box without an extra button press with the `selectBoxOnDrag` prop, which is overwritten by the `panOnDrag` prop so will need to be used in synchronicity
- Pan on `RightClick`
- Select box can select overlapping nodes that are not fully enclosed
- `onSelectionStart` and `onSelectionEnd` events that the user can subscribe to

It was a bit more complicated than I thought it would be, because I had to make sure the `selectionpane` was always there so that we can react to mouse events if the user wants selection to be the default action. I then realised we could merge the `pane` and the `selectionpane` together.


moklick:
Hey @jackfishwick

thanks for your PR! I like your ideas from what I get from your comment. It seems that the timing was a bit bad. We've just released a new version. Is is possible that you've done your changes against the previous `main` branch and not against `v11`.

jackfishwick:
Hey @moklick I'm glad you like the changes!

Yeah I realised a bit too late that there was a v11 branch so I did the changes against the old main branch, I've rebased now and it all seems to be working still. 

I just want to double check everything when I get time and write a cypress test for the new features 🤞

Bibiiii:
+1 to this PR - will be so helpful for our project. Thanks @jackfishwick

joeyballentine:
Wow, thanks for this. I've wanted these features for a long time now 👍

moklick:
Hey @jackfishwick ! Thanks again for your work here. I'll continue to work on it here: #2662

PR #2471, opened 2022-10-04
chore(examples): example page title corrected

Alireza29675:
- Example routes have names so they are getting shown better in dropdown

Screenshot 2022-10-04 at 16 10 20

- The title of the examples has been changed to "React Flow Examples - [Example Name]", so that if a development needs to be tested in two different examples at the same time, our tabs will be sufficiently organised.

Screenshot 2022-10-04 at 16 11 07
Screenshot 2022-10-04 at 16 11 22


moklick:
even when I am not sure if we will ever deploy it, I think this is a valuable change :) Thanks!

PR #2500, opened 2022-10-14
Put wrapper container in its own stacking context

neo:
This would put the wrapper and its content in "isolation" so their `z-index` wouldn't overflow to outside of the container.

Before (black pane is non react flow UI, which is not interactable because the cursor is also the grabbing hand moving the canvas behind):

image

After (black pane is interactable and the controls are now behind it):

image

### DIFF
--- a/packages/core/src/container/ReactFlow/index.tsx
+++ b/packages/core/src/container/ReactFlow/index.tsx
@@ -53,8 +53,9 @@ const initDefaultViewport: Viewport = { x: 0, y: 0, zoom: 1 };
 const wrapperStyle: CSSProperties = {
   width: '100%',
   height: '100%',
-  position: 'relative',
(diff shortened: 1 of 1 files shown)


moklick:
Thanks @neo

I don't understand which issue gets fixed with this PR. Why are the controls no inside the flow pane in the upper screenshot? Could you create a codesandbox or do you have a repo?

neo:
here's a simple repro: https://codesandbox.io/s/quizzical-panka-s4j76c?file=/App.js:296-344

you can read more from one of these links if you want:
- https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context
- https://web.dev/learn/css/z-index/
- https://www.joshwcomeau.com/css/stacking-contexts/

PR #2547, opened 2022-11-08
Prevent multiple edge label renderers collide with each other

liuruenshen:
The pull request addresses the issue when more than one ReactFlow exists in a browsing context, the EdgeLabelRenderer component used for placing the custom edge labels will collide with each other due to the same ID, causing all the edge labels to be inserted into one EdgeLabelRenderer Component that doesn't belong to those edge labels.

The solution here is to distinguish each EdgeLabelRenderer component by regarding the ReactFlow's ID property(same as the Background component).

Before the PR:
Screen Shot 2022-11-09 at 8 04 24 AM

After the PR:
Screen Shot 2022-11-09 at 8 03 17 AM


chrtze:
Thanks! This looks already very good. I am not sure about if we really should add the id from React Flow to the rendering portal element. An alternative way would be to look for the portal by doing `querySelector` from the root element which we have in the store. That way, we would find the portal that is inside the flow without needing the util function.

liuruenshen:
The alternative way you mentioned has the advantage that it eliminates unnecessary dependency(ReactFlow's ID). I can rewrite the code, using that way to fix the bug if you want to. 

> Thanks! This looks already very good. I am not sure about if we really should add the id from React Flow to the rendering portal element. An alternative way would be to look for the portal by doing `querySelector` from the root element which we have in the store. That way, we would find the portal that is inside the flow without needing the util function.

chrtze:
That would be great @liuruenshen! Thank you and let me know if you need any input or help.

liuruenshen:
@chrtze, I have updated my PR; review it when you are available; thanks.

chrtze (changes requested):
Thanks @liuruenshen for your time and effort spent on this, we really appreciate that.  However, I think your proposed changes add too much complexity for a fix that could be implemented easier. Let me clarify this: Adding a new property to the store is critical to performance and we should be really mindful about this. We already have a reference to the renderer element (`.react-flow__renderer`) with the store property called `d3Selection`. As this is a d3 object, you can get the DOM element by calling `d3Selection.node()` for example. I think by doing this, this would eliminate much of the code from your commit.

Would you be able to provide another update on this? Otherwise, let me know, I can take over as well.

liuruenshen:
Thanks for pointing it out; I couldn't find the entry to getting the DOM elements you said before, so I came up with the idea of defining another store field this way. Now I know how to do it, so if you are in a hurry, feel free to take the PR, or can you wait until next Monday(11/14)?

chrtze:
Sure, we can wait :) Thank you so much!

liuruenshen:
The last commit has taken the D3Selection instance in place. Let me know if it is done; thanks.

moklick:
@liuruenshen thanks for your work here :) You can make it even easier. We store a thing called `domNode` in the store which is the react flow container. I think you could do:

```js
const selector = (s: ReactFlowState) => ({
  edgeLabelRenderer: s.domNode?.querySelector('.react-flow__edgelabel-renderer')
});
```

Does that work?

liuruenshen:
Thanks! I will try it :)

liuruenshen:
Yes, it can work well and pass the test :)

chrtze (approved):

PR #2554, opened 2022-11-11
Always elevate by 1000 in `zIndex` when node is selected

neo:
Instead of only elevating when `zIndex` is unset

### DIFF
--- a/packages/core/src/store/utils.ts
+++ b/packages/core/src/store/utils.ts
@@ -39,7 +39,7 @@ export function createNodeInternals(nodes: Node[], nodeInternals: NodeInternals)
   const parentNodes: ParentNodes = {};
 
   nodes.forEach((node) => {
-    const z = isNumeric(node.zIndex) ? node.zIndex : node.selected ? 1000 : 0;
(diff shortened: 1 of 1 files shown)


moklick:
yep, I agree. Thanks @neo!

moklick:
will be released with #2559

neo:
> will be released with #2559

I'm wondering, when #2559 is merged including the same commit, and if we keep this PR open, it would automatically mark this as "merged" rather than "closed" currently? Not 100% sure, just a consideration to better differentiate which PRs are actually merged.

I also noticed there was #2560, which I think we could've changed the "base" of this PR to `next-release` and achieve the same? just suggesting what might be more clear and efficient 😆 

And thank you again for this library! 🎉

moklick:
@neo yep! The problem here was that I created a new branch based on your work (I needed to add a changeset) and then merged it before I changed the base branch of this PR. Afterwards it's not possible anymore unfortunately..

neo:
hey Moritz! sorry not being clear enough 😅 I meant next time for future considerations; I knew this was settled and couldn't be changed.

PR #2577, opened 2022-11-21
Fix fast zooming on windows

jackfishwick:
When you us `Ctrl+wheel` d3 increases the zoom delta by 10. It is more obvious on Windows machines, but you can see it on Mac as well. Using `Meta` as the zoom activation key means that on Windows users have to click `Ctrl` in order to get the zooming to work, and it jumps about a lot.

The answer is to make sure that if the `Ctrl` button is clicked, we do not apply our own delta (which is what we do when we are treating `Ctrl+wheel` as a pinch zoom. We also need to set the default delta so that it doesn't take into account the `Ctrl` key.

I have tested this on Windows machines and Mac and it seems to work fine!

### DIFF
--- a/packages/core/src/container/ZoomPane/index.tsx
+++ b/packages/core/src/container/ZoomPane/index.tsx
@@ -45,6 +45,10 @@ const selector = (s: ReactFlowState) => ({
   userSelectionActive: s.userSelectionActive,
 });
 
+const zoomDelta = (event: WheelEvent) => {
+  return -event.deltaY * (event.deltaMode === 1 ? 0.05 : event.deltaMode ? 1 : 0.002);
+};
+
 const ZoomPane = ({
   onMove,
   onMoveStart,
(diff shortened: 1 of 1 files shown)


moklick:
Hey @jackfishwick 

thanks for your help! Unfortunately this breaks the trackpad pinch zoom. For now I couldn't find a way to fix this. The fix would be to use another key for zoom activation on windows.

dominiczy:
@moklick `use another key for zoom activation on windows.` this doesn't work. Ctrl + scroll will always zoom (at least in Chrome), on both Windows and Mac. It can't be disabled by setting another zoom activation key. That makes the issue particularly annoying

bcakmakoglu:
@dominiczy Aren't those two different zooms though? One is a browser related zoom, which is out of the scope of React Flow and one (the one @moklick is talking about) is the React Flow canvas zoom.
As you said, I don't think there's a way to block the browser specific zoom and it's not really related to React Flow anyway, is it?

dominiczy:
@bcakmakoglu yes you are correct, but that doesn't matter for the bad user experience. There must be a way to fix the 'wrong' browser zoom behavior, apps like miro or whimsical don't have this problem

juspky:
Any updates here?

This is getting an issue for my app since ctrl is the most intuitive zoom key (Similar apps are using it too) so I don't want to change it.
ctrl zooming on windows is not usable at all in its current state.

Vochsel:
This is an issue for me too, and additionally, control + scroll, and pinch to zoom on mac seems remarkably fast. A multiplier to change scroll speed in general would be ideal.

juspky:
@moklick 👋

ArturMakly:
**plus +100** lots of Windows user complaining to us - thanks!

PR #2586, opened 2022-11-23
ref: stop using try..catch as control flow in Wrapper

sdegueldre:
Previously, Wrapper used try..catch around a call to useStoreApi to check whether there was an available StoreContext, instead of simply using useContext and checking the value. This leads to poor developer experience when using the devtools feature "pause on caught exceptions" as it causes the devtools to break on every render of a Wrapper that is not within a StoreContext provider. The catch is also universal, meaning that it could catch a completely unrelated error and happily keep going instead of crashing which is undesirable.

This commit changes this try..catch into a simple call to useContext to check if a StoreContext is available.

### DIFF
--- a/packages/core/src/container/ReactFlow/Wrapper.tsx
+++ b/packages/core/src/container/ReactFlow/Wrapper.tsx
@@ -1,18 +1,10 @@
-import type { FC, PropsWithChildren } from 'react';
+import { FC, PropsWithChildren, useContext } from 'react';
 
-import { useStoreApi } from '../../hooks/useStore';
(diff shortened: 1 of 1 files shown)


sdegueldre:
Note: on my machine, 1 test in `view-props.cy.tsx` is currently failing and another is flaky. This was also true before my changes and I don't expect these failures have anything to do with my changes. Either the code is actually broken causing these tests to fail, or they depend on the hardware configuration or something else causing them to fail only for me. Please advise.

moklick:
Thanks! That's a good finding.. it wasn't possible with zustand 3.x

PR #2620, opened 2022-11-29
Add generic support for the type property on the Node type

RichSchulz:
I have added custom Node Types in my codebase and it could be great if I could set the type property of the Node type to an Enum or string literal. In order to support that, I have added another Generic to the Node type and then set the type property to exactly that generic type. Note that, the default of that generic type is the same as the current type is, so this change should be fully backwards compatible.

Also when making this change, I was wondering why type can be undefined?

### DIFF
--- a/packages/core/src/types/nodes.ts
+++ b/packages/core/src/types/nodes.ts
@@ -5,11 +5,11 @@ import { internalsSymbol } from '../utils';
 import type { XYPosition, Position, CoordinateExtent, HandleElement } from '.';
 
 // interface for the user node items
-export type Node<T = any> = {
+export type Node<T = any, U extends string | undefined = string | undefined> = {
   id: string;
   position: XYPosition;
   data: T;
-  type?: string;
+  type: U;
   style?: CSSProperties;
   className?: string;
   sourcePosition?: Position;


moklick:
Hey! Thanks for the suggestion. Could you explain why you would like to pass the type as a generic here?

RichSchulz:
> Hey! Thanks for the suggestion. Could you explain why you would like to pass the type as a generic here?

Sure! So in my codebase I use a controlled graph with my own Nodetypes and NodeDataTypes:

```
const NodeTypes = "nodeA" | "nodeB";
const NodeDataType = {"some_value": string};
```

Then I define my own type for Node using the Node from react-flow, which I can then use in all my graph modification functions:

```
const MyNode = Node<NodeDataType>;
const someNode: MyNode = getNode(...);
```

If I want to check the type of someNode against lets say "nodeA" I get an TS error because someNode.type can be undefined. Also if I did the following I wouldnt get a TS error:

```
if (someNode.type !== undefined && someNode.type === "non-existing-type"){
...
} 
```

To solve these issues I would like to define MyNode in the following way:

`const MyNode = Node<NodeDataType, NodeTypes>;`

RichSchulz:
Hey, could someone give this PR another look @moklick ?

moklick:
Hey! We will check it again before the next release!

RichSchulz:
> Hey! We will check it again before the next release!
> 
> It was a bad idea to make the idea to make the type optional, but we won't change that (at least not in the near future) since it would be a breaking change.

Great, thanks a lot. Looking forward to that release and feel free to keep me in the loop :)

Issue #2546, opened 2022-11-07
Background state issues with multiple flows


rChaoz:
### Describe the Bug

When creating multiple flows, it seems like the background state is shared across the flows.
Duplicate of #1037

### Your Example Website or App

https://cxh0oz.csb.app/

### Steps to Reproduce the Bug or Issue

Check the website, and try to pan the 2 flows.

### Expected behavior

-

### Screenshots or Videos

Check video of #1037

### Platform

Browser: Firefox

liuruenshen:
According to the codes, you need to specify a unique ReactFlow.id value for each ReactFlow; then, the issue can be solved.

```
<ReactFlow id="a"/>
<ReactFlow id="b"/>
```

I suggest the approach can be written into the document when multiple ReactFlow components are present.

chrtze:
Thanks @liuruenshen, I have added this to our docs: https://github.com/wbkd/react-flow-docs/commit/beabb9ee161e098801a36368890be12caeefade9

PR #2638, opened 2022-12-06
Assign the random root id by default

liuruenshen:
When there are multiple flows on a page, the developer needs to specify a unique ID for each flow; otherwise, an issue will appear like [#2546](https://github.com/wbkd/react-flow/issues/2546).

The commit assigns a unique random ID by default so that the developer can ignore the ID property.

The [commit](https://github.com/wbkd/react-flow-docs/commit/beabb9ee161e098801a36368890be12caeefade9) can be reversed too.


moklick:
Thanks @liuruenshen but this doesn't work for SSR. We will change this with the next major release where we will only support React18+ so we can use [`useId`](https://beta.reactjs.org/apis/react/useId).

Issue #2622, opened 2022-11-30
`onNodeDragStop` triggered on node click


danielgek:
### Describe the Bug

Hey, I'm having some problem wrapping my head around the `onNodeDragStop` callback
Our current setup uses redux to store the nodes and edges on our side(controlled flow), and we have a redux action that updates the nodes position when the `onNodeChanges` callback fires changes of type `position` and a thunk action to actually save the position on the server once the drag ends, for this i'm using `onNodeDragStop`

The problem with this is that `onNodeDragStop` is triggered on node click when for example selecting a node

### Your Example Website or App

https://codesandbox.io/s/restless-dust-f07gxl?file=/store/boardSlice.js

### Steps to Reproduce the Bug or Issue

1. select a node
2.  `onNodeDragStop` is triggered

### Expected behavior

`onNodeDragStop` is only triggered on a drag end

### Platform

- OS:  macOS
- Browser: Chrome
- Version: [11.3.2]

danielgek:
I made a workaround this: https://codesandbox.io/s/frosty-meadow-6j8m65

but it still feels like a hack imho

danielgek:
yes a threshold for firing `onNodeDragStop` would be enough i think

moklick:
Since 11.9.0 it's possible to use the [`nodeDragThreshold` prop](https://reactflow.dev/docs/api/react-flow-props/#nodedragthreshold) to delay / prevent the drag events on click.

noah-jun-521:
As of 2024 Jan 31, it is still causing both onNodeClick and onNodeDragStop with nodeDragThreshold being 5.

kevjin:
> As of 2024 Jan 31, it is still causing both onNodeClick and onNodeDragStop with nodeDragThreshold being 5.

works for me, maybe your screen resolution?

PR #2722, opened 2023-01-02
fix(usedrag): block onNodeDragStop when user just click node

ryomahan:
this PR is useful for #2622

When I want to create a solution to autosave flow's change for my project, I find that onNodeDragStop will be call when I just click a node, that's not I want to see. So I search in issues, and just find #2622(still open), there is no simple solution to this problem without change source. After I looked at the source about onNodeDragStop I found that the solution to this problem maybe actually very simple, just add a variable to determine whether to drag or not. So I fix it and push this PR.

I hope it is useful.

### DIFF
--- a/packages/core/src/hooks/useDrag/index.ts
+++ b/packages/core/src/hooks/useDrag/index.ts
@@ -48,6 +48,7 @@ function useDrag({
       if (disabled) {
         selection.on('.drag', null);
       } else {
+        let isMoved = false
         const dragHandler = drag()
           .on('start', (event: UseDragEvent) => {
             const {
@@ -88,6 +89,7 @@ function useDrag({
             }
           })
           .on('drag', (event: UseDragEvent) => {
+            isMoved = true
             const {
               updateNodePositions,
               nodeInternals,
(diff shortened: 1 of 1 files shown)


ryomahan:
@moklick

moklick:
This was the previous behaviour and we changed it intentionally. There will always be a drag start (also when you just click) and often times you do some kind of cleanup work on drag stop and that's only possible when we trigger both events. It's weird to only trigger drag start and no drag stop.

ryomahan:
@moklick Oh, I think I understand the situation you're talking about. So could we add an new event(like onNodeDragStopWithDrag), or make onNodeDragStop provide a parameter that would allow user to determine if a dragging action was performed. 

I'm happy to code it as long as which idea you want to use is clear, because I desperately need it.

moklick:
As a workaround you could add a hasMoved flag in your code base and only trigger the onNodeDragStop function if that's true.

ryomahan:
> As a workaround you could add a hasMoved flag in your code base and only trigger the onNodeDragStop function if that's true.

Do you mean in my own project code or in my local reactflow?

ryomahan:
Yes, that's exactly what I'm doing now.

moklick:
We leave the current behaviour as it is.

PR #2726, opened 2023-01-03
Backported EdgeLabelRenderer to v10 branch.

joshwilsonvu:
As mentioned in https://github.com/wbkd/react-flow/discussions/2725, I'm interested in being able to use the `EdgeLabelRenderer` component in React Flow v10. I've done what I believe is necessary to make it available based on previous PRs. Please let me know if I should make any changes!

### DIFF
--- a/src/components/EdgeLabelRenderer/index.tsx
+++ b/src/components/EdgeLabelRenderer/index.tsx
@@ -1,0 +1,18 @@
+import type { ReactNode } from 'react';
+import { createPortal } from 'react-dom';
+import { useStore } from '../../store';
+import { ReactFlowState } from '../../types';
+
+const selector = (s: ReactFlowState) => s.domNode?.querySelector('.react-flow__edgelabel-renderer');
+
+function EdgeLabelRenderer({ children }: { children: ReactNode }) {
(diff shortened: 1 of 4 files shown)


moklick:
Thanks for your work, but this is not necessary. You can use the EdgeLabelRenderer on top of React Flow without integrating it into the library. We will not do another v10 release.

joshwilsonvu:
Thank you for your clarity; I understand that a new v10 release was not to be taken for granted.

As a user of the library, I can reimplement `EdgeLabelRenderer`, but I don't believe I can add the `<div className="react-flow__edgelabel-renderer" />` node, which it relies on, to the `GraphView` component. That's the only part of this MR that needs to be integrated, from what I can tell. I suppose I can patch the package or continue to use SVG `foreignObject`.

moklick:
Does this work for you: https://codesandbox.io/s/portaledgelabelrenderer-ikxqkb?file=/EdgeLabelRenderer.jsx

joshwilsonvu:
Very interesting—something like that could work, if I can make sure it doesn't conflict with React's management of that DOM node. Regardless, I appreciate your help!

PR #2749, opened 2023-01-12
Node-resizer: add shouldResize handler

moklick:
Two improvements are planned for the node resizer:

1. Add an ~`onBeforeResize`~  `shouldResize` handler that returns a boolean. This can be helpful to implement dimensions constraints that are based on other nodes for example.

```js
const shouldResize = (event, params) => {
 return params.width < 100;
}
```

2. Add a `direction` attribute to the passed params in the event handlers. This is helpful if your calculations are different based on the direction where a resize movement comes from. Not sure how this could look like. Maybe [1, 0] for a movement to the right, [0, -1] to the top, [1, 1] to the bottom right and so forth?

### DIFF
--- a/.changeset/pretty-spoons-punch.md
+++ b/.changeset/pretty-spoons-punch.md
@@ -1,0 +1,5 @@
+---
+'@reactflow/node-resizer': major
+---
+
+Add shouldresize, cleanup types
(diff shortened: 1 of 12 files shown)


moklick:
Should we pass the current values (dimension and position) to the onBeforeResize handler or the next values? If we would pass the next values we could also remove the onBeforeResize handler and give the user the possibility to cancel the event in the onResize handler?! Not sure about this.

PR #2760, opened 2023-01-18
Connection Radius

moklick:
Adds a new prop `connectionRadius` - this makes it possible to adjust the radius around a handle where you can connect an edge. This PR makes it easier to connect nodes because you don't need to drop the connection line above a handle but only close to it.

https://user-images.githubusercontent.com/2857535/213244354-4f19de23-21c9-4013-9bf9-718d40ddaf32.mov


moklick:
This is now part of #2754

PR #2763, opened 2023-01-23
Feat: touch connect

moklick:
This enables the user to connect two nodes on a touch screen without using click connect.

### DIFF
--- a/packages/core/src/components/Handle/index.tsx
+++ b/packages/core/src/components/Handle/index.tsx
@@ -68,9 +68,9 @@ const Handle = forwardRef<HTMLDivElement, HandleComponentProps>(
     };
 
     const onPointerDown = (event: ReactMouseEvent<HTMLDivElement> | ReactTouchEvent<HTMLDivElement>) => {
-      const isMouse = isMouseEvent(event);
+      const isMouseTriggered = isMouseEvent(event);
 
-      if ((isMouse && event.button === 0) || 'touches' in event) {
+      if ((isMouseTriggered && event.button === 0) || !isMouseTriggered) {
         handlePointerDown({
           event,
           handleId,
@@ -83,7 +83,7 @@ const Handle = forwardRef<HTMLDivElement, HandleComponentProps>(
         });
       }
 
-      if (isMouse) {
+      if (isMouseTriggered) {
         onMouseDown?.(event);
       } else {
         onTouchStart?.(event);
--- a/packages/core/src/components/Handle/index.tsx
+++ b/packages/core/src/components/Handle/index.tsx
@@ -1,11 +1,11 @@
-import { memo, HTMLAttributes, forwardRef, MouseEvent as ReactMouseEvent } from 'react';
+import { memo, HTMLAttributes, forwardRef, MouseEvent as ReactMouseEvent, TouchEvent as ReactTouchEvent } from 'react';
 import cc from 'classcat';
 import { shallow } from 'zustand/shallow';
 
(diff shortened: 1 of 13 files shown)


packages/core/src/utils/index.ts
```diff
@@ -111,3 +115,21 @@ export function isInputDOMNode(event: KeyboardEvent | ReactKeyboardEvent): boole
...
+export const getEventPosition = (
+  event: MouseEvent | ReactMouseEvent | TouchEvent | ReactTouchEvent,
+  bounds: DOMRect
+) => {
+  const isMouse = isMouseEvent(event);
+  const evtX = isMouse ? event.clientX : event.touches?.[0].clientX;
+  const evtY = isMouse ? event.clientY : event.touches?.[0].clientY;
+
+  return {
+    x: evtX - bounds.left,
```
sohamsshah, line 132: if `bounds` is undefined, won't it throw an error?

packages/core/src/components/Handle/index.tsx
```diff
@@ -66,9 +67,11 @@ const Handle = forwardRef<HTMLDivElement, HandleComponentProps>(
       onConnect?.(edgeParams);
     };
 
-    const onMouseDownHandler = (event: ReactMouseEvent<HTMLDivElement>) => {
-      if (event.button === 0) {
-        handleMouseDown({
+    const onPointerDown = (event: ReactMouseEvent<HTMLDivElement> | ReactTouchEvent<HTMLDivElement>) => {
+      const isMouse = isMouseEvent(event);
+
+      if ((isMouse && event.button === 0) || 'touches' in event) {
```
sohamsshah, line 73: Here, constant zero might be considered as a `value`; while it is used to denote a `MouseEvent` button value (0 -> no button, 1 -> primary and so on) (https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons)

So, in case if we are also having `event.button === 1` and so on; we should replace with meaningful constant vars here.

packages/core/src/components/Handle/index.tsx
```diff
@@ -66,9 +67,11 @@ const Handle = forwardRef<HTMLDivElement, HandleComponentProps>(
       onConnect?.(edgeParams);
     };
 
-    const onMouseDownHandler = (event: ReactMouseEvent<HTMLDivElement>) => {
-      if (event.button === 0) {
-        handleMouseDown({
+    const onPointerDown = (event: ReactMouseEvent<HTMLDivElement> | ReactTouchEvent<HTMLDivElement>) => {
+      const isMouse = isMouseEvent(event);
```
sohamsshah, line 71: `isMouse` variable name can be renamed to something like: `isMousePointerDown` or `isMouseEventTriggered`

packages/core/src/utils/index.ts
moklick: it's only used internally and we don't allow to pass undefined bounds (via types)

packages/core/src/components/Handle/index.tsx
moklick: thanks for your input here. button === 0 means left click here

packages/core/src/components/Handle/index.tsx
moklick: good idea

packages/core/src/components/Handle/index.tsx
sohamsshah: Ohhh, interesting. Saw the changes made; thanks!

PR #2764, opened 2023-01-23
Next release: 11.5.0

moklick:
## reactflow package

This new release comes with lots of improvements for connecting nodes and moving on the pane:

- auto pan: pane moves when you drag a node or connection line to the border of the pane
- connection radius: you can drop a connection line close to a handle to connect it (no need to drop it exactly on top of it)
- touch connect: start a connection line on a touch device
- onError: add prop to get notified when an error occurs
- types: improve specifying node type

## @reactflow/node-resizer package

After this update it should be easier to update the node resizer (no need to update the reactflow package anymore).

New props:

- `shouldResize`: user can pass a function that determines if resize should be executed
- `direction`: gets passed as an attribute on resize


packages/core/src/types/nodes.ts
```diff
@@ -5,11 +5,11 @@ import { internalsSymbol } from '../utils';
 import type { XYPosition, Position, CoordinateExtent, HandleElement } from '.';
 
 // interface for the user node items
-export type Node<T = any> = {
+export type Node<T = any, U extends string | undefined = string | undefined> = {
```
evgenius1424, line 8: Cool! Thank you.

PR #2847, opened 2023-02-16
feat(core): add option to enable/disable replacing id when updating an edge

bcakmakoglu:


moklick:
thanks man! Can you put shouldReplaceId in an object. That way we are more flexible with adding even more options later on :)

bcakmakoglu:
Sure 👍

bcakmakoglu:
@moklick PR is updated btw =)

moklick:
Thanks @bcakmakoglu 🙌

PR #2860, opened 2023-02-24
Pass actual node type

klpx:
In my application I have have node of different types. Not all of them have their specific `ComponentType`.
So I specify "default" node type. But in this case I dont have access to node type because it is being overwritten with `"default"` string.


moklick:
we are passing the type that is getting used for the node. When you are using an invalid/ non-existing type we are replacing it with "default". Only in that case `nodeType` differs from `node.type`.

PR #2872, opened 2023-02-28
fix(core): track modifier keys on useKey press

### LINKED ISSUE #2785: Zoom activationkey remains pressed
### Describe the Bug

While using figma like controls, if i press cmd (meta key on mac) inside of an input,  board seams to keep a state of meta key pressed and allways does zoom on scroll wheel instead of panning, 

https://codesandbox.io/s/busy-benji-fyenzx?file=/Board.js

1. Use scroll wheel on the board to pan
2. Use scroll wheel on the board with cmd pressed to zoom
3. Focus input
4. Click on cmd
5.  Use scroll wheel, it starts to zoom instead of panning 

### Expected behavior

Clicking cmd on inputs does not affect board cmd for zoom behaviour 

### Screenshots or Videos

_No response_

### Platform

- OS: [macOS, Windows, Linux]
- Browser: [Chrome]
- Version: [e.g. 109.0.5414.87]

### Additional context

_No response_

danielgek:
@moklick i tested this with all the examples and it seams to be working fine

Thanks @bcakmakoglu for pointing me the light 

This closes #2785


ERuseckas:
when should I expect this merged and released?

jsweetman:
Also having a problem with this code, looking forward to it being merged. It my case sometimes users cannot pan the graph and instead can only perform multi-select as if they're holding shift.

moklick:
Thanks for your PR @danielgek ! When we do this, can't we get rid of the modifier check in `inInputDOMNode` https://github.com/wbkd/react-flow/blob/main/packages/core/src/utils/index.ts#L108 ?

danielgek:
@moklick i think so, i removed it and tested it and it all seams fine ;) (force pushed the change)

PR #2874, opened 2023-03-01
feat(nodes): make it possible to set origin per node

ekiwi111:
Discussion - https://github.com/wbkd/react-flow/discussions/2871

### DIFF
--- a/packages/core/src/container/NodeRenderer/index.tsx
+++ b/packages/core/src/container/NodeRenderer/index.tsx
@@ -100,7 +100,7 @@ const NodeRenderer = (props: NodeRendererProps) => {
           y: posY,
           width: node.width ?? 0,
           height: node.height ?? 0,
-          origin: props.nodeOrigin,
+          origin: node.nodeOrigin || props.nodeOrigin,
         });
 
         return (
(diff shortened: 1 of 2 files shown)


bcakmakoglu:
This PR doesn't cover store actions like 
1. `setNodes`
2. `setDefaultNodesAndEdges`
3. `updateNodeDimensions`
4. `triggerNodeChanges`

There's probably a couple more places where `nodeOrigin` would have to be updated to use either the per node origin or the global nodeOrigin.

e-oldman:
> This PR doesn't cover store actions like
> 
> 1. `setNodes`
> 2. `setDefaultNodesAndEdges`
> 3. `updateNodeDimensions`
> 4. `triggerNodeChanges`
> 
> There's probably a couple more places where `nodeOrigin` would have to be updated to use either the per node origin or the global nodeOrigin.

Thanks for the directions, will look into that

moklick:
This will be possible with the next major or eventually minor release. It's already implemented in the "feat/packages" branch.

moklick:
This is part of the v12 release. It's still in beta, but you can already try it out. All details: https://github.com/xyflow/xyflow/discussions/3764

PR #2877, opened 2023-03-02
feat(reactflow): add isValidConnection prop closes #2876

### LINKED ISSUE #2876: isValidConnection prop for <ReactFlow />

Right now we have an `isValidConnection` prop for the Handle component. It's hard to configure, because you need to pass it to every handle and it's hard to access for the library. There are also some bugs like that you can always update an edge even if the connection is not valid. With a general `isValidConnection` we could also filter out all invalid connections that are being passed with wrongly configured edges.

moklick:


packages/core/src/store/index.ts
```diff
@@ -33,8 +33,8 @@ const createRFStore = () =>
       return Array.from(get().nodeInternals.values());
     },
     setEdges: (edges: Edge[]) => {
-      const { defaultEdgeOptions = {} } = get();
-      set({ edges: edges.map((e) => ({ ...defaultEdgeOptions, ...e })) });
```
danielgek, line 37: @moklick this change causes a bug for folks using a redux store:
```
index.js:2820 Uncaught TypeError: Cannot add property selected, object is not extensible
    at index.js:2820:18
    at Array.reduce (<anonymous>)
    at getSelectionChanges (index.js:2817:18)
    at addSelectedEdges (index.js:3544:28)
    at onEdgeClick (index.js:1210:17)
    at HTMLUnknownElement.callCallback2 (react-dom.development.js:4164:14)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:4213:16)
    at invokeGuardedCallback (react-dom.development.js:4277:31)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:4291:25)
    at executeDispatch (react-dom.development.js:9041:3)
```

Reverting this set state fixes the problem but i'm not sure if impacts other features implemented on this pr, i can summit a pr reverting to the old way

moklick: this wasn't on purpose.. but it's weird that you get that error. We are not adding a property here.

danielgek: Yep i agree it's weird, this somehow triggers  [getSelectionChanges](https://github.com/wbkd/react-flow/blob/d709c7860406a691a480bcf7a81533c80287a0e5/packages/core/src/utils/changes.ts#LL132C17-L132C37) which triggers the error, just to add more context

moklick: and you are sure that just doing:

```js
set({ edges: defaultEdgeOptions ? edges.map((e) => ({ ...defaultEdgeOptions, ...e })) : edges });
```

fixes it?

danielgek: Reverting it to `set({ edges: edges.map((e) => ({ ...defaultEdgeOptions, ...e })) });` fixes it, i can make a pr with. redux example in the vite-app tomorrow if you want to demonstrate this behaviour

moklick: ah :/ no I get it. Of course this doesn't work. Will fix and release it today.

danielgek: nice, but i~m curious, why would this change cause such problem ?

moklick: fixed in 11.6.1 -  redux toolkit uses immer under the hood. immer creates immutable objects and doesn't like that we do edge.selected = true / false in our code.

danielgek: Yes i got that i was just wondering why this would end up calling edge.selected = true, but no worries it's fixed, thanks for the fast response

PR #2887, opened 2023-03-05
Feat/toggle selection

Max-ChenFei:
Hi, I can update the prop and its class changed, but it did not render. The same question I meet in (#2885).
Esc23

### DIFF
--- a/packages/core/src/components/Edges/wrapEdge.tsx
+++ b/packages/core/src/components/Edges/wrapEdge.tsx
@@ -138,15 +138,8 @@ export default (EdgeComponent: ComponentType<EdgeProps>) => {
 
     const onKeyDown = (event: KeyboardEvent) => {
       if (elementSelectionKeys.includes(event.key) && elementsSelectable) {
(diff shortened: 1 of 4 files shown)


moklick:
The issue here is that the element is still focused even if it's not selected anymore. I pushed a change so we call `blur` when a node gets deselected: https://github.com/wbkd/react-flow/commit/738320038839bf557f1e9a729294b66917748f0c#diff-d24cb9d9a63ea3c67705c3ff409953f1a0935ee2f7801984f7103f7714312112R81

Max-ChenFei:
Hi @moklick, should I change the destination to `next-release` branch and will you merge it?

moklick:
It's will be fixed when I merge #2891

Max-ChenFei:
Thanks. I test again, and it works. Will you merge this pull request to allow user to select or deselect when pressing the Control key?

Max-ChenFei:
This "Feat/toggle selection 2887" PR allows the user to deselect the node when pressing the control key if this node is selected. And vice versa.

moklick:
I will not merge #2887 but #2891 which solves the issue described in this PR

PR #2923, opened 2023-03-17
Update types for utils/changes

Slowyn:
Little type improvements, so change functions could respect Node<T, U> U parameter

### DIFF
--- a/packages/core/src/utils/changes.ts
+++ b/packages/core/src/utils/changes.ts
@@ -115,12 +115,12 @@ function applyChanges(changes: any[], elements: any[]): any[] {
   }, initElements);
 }
 
-export function applyNodeChanges<NodeData = any>(changes: NodeChange[], nodes: Node<NodeData>[]): Node<NodeData>[] {
-  return applyChanges(changes, nodes) as Node<NodeData>[];
+export function applyNodeChanges<NodeType extends Node>(changes: NodeChange[], nodes: NodeType[]): NodeType[] {
+  return applyChanges(changes, nodes) as NodeType[];
 }
 
-export function applyEdgeChanges<EdgeData = any>(changes: EdgeChange[], edges: Edge<EdgeData>[]): Edge<EdgeData>[] {
-  return applyChanges(changes, edges) as Edge<EdgeData>[];
+export function applyEdgeChanges<EdgeType extends Edge>(changes: EdgeChange[], edges: EdgeType[]): EdgeType[] {
+  return applyChanges(changes, edges) as EdgeType[];
 }
 
 export const createSelectionChange = (id: string, selected: boolean) => ({


bcakmakoglu:
Wouldn't this be a breaking change, although a minor one, for anyone already using the existing generic to pass the `NodeData` or `EdgeData` type?
A non-breaking change for this would be to just add a second generic to `applyNodeChanges` in order to be able to pass the `U` param to the `Node` type.

```ts
function applyNodeChanges<NodeData = any, NodeType extends string | undefined = string | undefined>(changes: NodeChange[], nodes: Node<NodeData, NodeType>[]): Node<NodeData, NodeType>[] {
  return applyChanges(changes, nodes) as Node<NodeData, NodeType>[];
}
```

Slowyn:
You are actually correct — this will be a breaking change. In cases when type isn't derived, for instance
```ts
applyNodeChanges<MyNodeData>(changes, nodes)
```


I came to the decision I made in PR because I have faced this problem:

```ts
// Let's declare a few Node types
type Node1 = Node<{}, 'node-type1'>;
type Node2 = Node<{}, 'node-type2'>;
type Node3 = Node<{}, 'node-type3'>;

// Common type for these nodes
type Node = Node1 | Node2 | Nod3;

// Imagine we have a state
type State = {
  nodes: Node[];
}

// Let's try to apply changes
// Keep in mind we have this implementation
/*
function applyNodeChanges<NodeData = any, NodeType extends string | undefined = string | undefined>(changes: NodeChange[], nodes: Node<NodeData, NodeType>[]): Node<NodeData, NodeType>[] {
  return applyChanges(changes, nodes) as Node<NodeData, NodeType>[];
}
*/

function onNodesChange(changes: NodeChange[]) {
    // Here we would have a typescript error. It wouldn't be able to derive the correct types
    set({
        nodes: applyNodeChanges(changes, get().nodes),
    })
}
```

The current situation is Node generic has two arguments but change utils don't know anything about the second parameter.
Let's try to figure out how to fix it.
A possible decision is to declare another function with given signatures.

moklick:
we won't to a major release soon. The suggestion from @bcakmakoglu is something we could do in the next release.

FelipeEmos:
Actually there's a very simple way of achieving a much better typesafety without a breaking change

The problem lies here... for some reason the Type Safety for the `NodeChange` type is lost...
https://github.com/xyflow/xyflow/blob/14c4f6512465425ad3c977cc88b45705631fa7d3/packages/react/src/types/changes.ts#L43-L53

It could just be a generic
```tsx
export type NodeChange<NodeType extends Node = Node> =
  | NodeDimensionChange
  | NodePositionChange
  | NodeSelectionChange
  | NodeRemoveChange
  | NodeAddChange<NodeType>
  | NodeReplaceChange<NodeType>;
```

Same for EdgeChange

```tsx
export type EdgeChange<EdgeType extends Edge = Edge> =
  | EdgeSelectionChange
  | EdgeRemoveChange
  | EdgeAddChange<EdgeType>
  | EdgeReplaceChange<EdgeType>;
```

Then in our commit you would pass down the type of the generic into the change
```tsx
export function applyNodeChanges<NodeType extends Node = Node>(changes: NodeChange<NodeType>[], nodes: NodeType[]): NodeType[] {
  return applyChanges(changes, nodes) as NodeType[];
}

export function applyEdgeChanges<EdgeType extends Edge = Edge>(changes: EdgeChange<EdgeType>[], edges: EdgeType[]): EdgeType[] {
  return applyChanges(changes, edges) as EdgeType[];
}
```

### TL;DR
We only gotta change the passing of the already existing generic into the `changes` type.
That won't break anything cus all the previous behavior is contained in this simpler approach.

FelipeEmos:
Should I open another PR with what I proposed?

Let me know what you think...
I believe such change doesn't need to wait the next major release ( for the reasons I explained above 🙂 )

Cheers! Thanks for the amazing lib 🙏🏻

moklick:
We changed this for v12. Thanks for your input here @Slowyn, @bcakmakoglu and @FelipeEmos !

React Flow 12 is still in beta, but you can already try it out :) Details: https://github.com/xyflow/xyflow/discussions/3764

PR #2935, opened 2023-03-20
Next release

### LINKED ISSUE #2916: add current connection result in the store
storing the connection result would make it possible to style the connection line depending on the current target for example. Currently it's not possible to access the target node while a connection is ongoing.

### LINKED ISSUE #2919: react-flow 11.5.0+ breaks for nextjs 12
### Describe the Bug

react-flow 11.5.0+ breaks for nextjs 12 when doing next build

https://github.com/wbkd/react-flow-example-apps

change version to nextjs ^12 and react-flow ^11.5.0

### Expected behavior

expect no errors

got an error:

```
index.mjs:466 TypeError: A.001 is not a function
    at 10729 (index.js:29:1)
```

```
const errorMessages = {
    '001': () => '[React Flow]: Seems like you have not used zustand provider as an ancestor. Help: https://reactflow.dev/error#001',
    '002': () => "It looks like you've created a new nodeTypes or edgeTypes object. If this wasn't on purpose please define the nodeTypes/edgeTypes outside of the component or memoize them.",
    '003': (nodeType) => `Node type "${nodeType}" not found. Using fallback type "default".`,
    '004': () => 'The React Flow parent container needs a width and a height to render the graph.',
    '005': () => 'Only child nodes can use a parent extent.',
    '006': () => "Can't create edge. An edge needs a source and a target.",
    '007': (id) => `The old edge with id=${id} does not exist.`,
    '009': (type) => `Marker type "${type}" doesn't exist.`,
    '008': (sourceHandle, edge) => `Couldn't create edge for ${!sourceHandle ? 'source' : 'target'} handle id: "${!sourceHandle ? edge.sourceHandle : edge.targetHandle}", edge id: ${edge.id}.`,
    '010': () => 'Handle: No node id found. Make sure to only use a Handle inside a custom Node.',
    '011': (edgeType) => `Edge type "${edgeType}" not found. Using fallback type "default".`,
};

const zustandErrorMessage = errorMessages['001']();
function useStore(selector, equalityFn) {
    const store = useContext(StoreContext);
    if (store === null) {
        throw new Error(zustandErrorMessage);
    }
    return useStore$1(store, selector, equalityFn);
}
```

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 91.1]

### Additional context

_No response_

### LINKED ISSUE #2948: Handle: connection start from target handle

Currently it's hard to prevent starting a new connection from a specific handle. It would be nice to have an option like "connectionStartHandle" true/false to make it easier to configure.

### LINKED ISSUE #2953: If at least 1 node is set to hidden, useNodesInitialized will never return true
### Describe the Bug

`useNodesInitialized` will never return true if at least one node is hidden. Technically this behaviour is what the documentation describes, as no size will be calculated for the hidden node, but I think it would be more useful if `useNodesInitialized` only concidered visible nodes.

My use case is that I want to do automatic layout calculations, but simply skip the nodes that are not visible.

https://codesandbox.io/s/romantic-jones-wfwu0x?file=/App.js

1. Open the Codesandbox
2. Check the console, it will log `{nodesInitialized: false}`. 
3. Remove the hidden attribute from the first node
4. Observe that `{nodesInitialized: true}` is logged as expected

### Expected behavior

`useNodesInitialized` should only concider visible nodes

### Screenshots or Videos

_No response_

### Platform

Version: 11.6.1

### Additional context

_No response_

moklick:
- prefix error keys with error closes #2919
- make wheel listener non-passive explicitly thanks @Elringus
- add current connection result to the store closes #2916
- background: add id and offset to background props thanks @Elringus
- minimap: add inversePan and stepZoom prop thanks @Elringus
- use-nodes-initialized: add options, ignore hidden nodes closes #2953
- handle component: add isConnectableStart and isConnectableEnd closes #2948
- Add updatable prop to Edge type  thanks @bcakmakoglu 
    - Allows enabling/disabling edge update for specific edges
    - Allows setting updatable to either boolean, or string (either source or target)
- Add @reactflow/node-resizer package to reactflow package

### DIFF
--- a/.changeset/many-parents-applaud.md
+++ b/.changeset/many-parents-applaud.md
@@ -1,5 +1,0 @@
----
-'@reactflow/core': patch
----
-
-fix non-passive wheel event listener violation
--- a/.changeset/many-parents-applaud.md
+++ b/.changeset/many-parents-applaud.md
@@ -1,0 +1,5 @@
+---
+'@reactflow/core': patch
+---
+
+fix non-passive wheel event listener violation
--- a/.changeset/nice-impalas-doubt.md
+++ b/.changeset/nice-impalas-doubt.md
@@ -1,0 +1,5 @@
+---
+'@reactflow/core': patch
+---
+
+prefix error keys with "error"
--- a/.changeset/nice-impalas-doubt.md
+++ b/.changeset/nice-impalas-doubt.md
@@ -1,5 +1,0 @@
----
-'@reactflow/core': patch
----
-
-prefix error keys with "error"
--- a/.changeset/plenty-students-wash.md
+++ b/.changeset/plenty-students-wash.md
@@ -1,0 +1,5 @@
+---
+'@reactflow/core': patch
+---
+
+add connection result to store
--- a/.changeset/plenty-students-wash.md
+++ b/.changeset/plenty-students-wash.md
@@ -1,5 +1,0 @@
----
-'@reactflow/core': patch
----
-
-add connection result to store
--- a/.changeset/rich-papayas-hope.md
+++ b/.changeset/rich-papayas-hope.md
@@ -1,0 +1,5 @@
+---
+'@reactflow/minimap': minor
+---
+
+add inversePan and zoomStep props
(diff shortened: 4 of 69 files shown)


Slowyn:
Hi @moklick! Can you consider including this [PR](https://github.com/wbkd/react-flow/pull/2923) in this release?

PR #2940, opened 2023-03-20
add twin lines background variant

Elringus:
This adds new background variant type `TwinLines` (4th on the screenshot below).

scrn

First line is controlled by the same properties as used in `Lines` variant; second line width, gap and color are controlled with `twinLineWidth`, `twinGap` and `twinColor` properties.

Related discussion: https://github.com/wbkd/react-flow/discussions/2937

### DIFF
--- a/examples/vite-app/src/examples/Backgrounds/index.tsx
+++ b/examples/vite-app/src/examples/Backgrounds/index.tsx
@@ -36,6 +36,7 @@ const Backgrounds: FC = () => (
     <Flow id="flow-a" bgProps={{ variant: BackgroundVariant.Dots }} />
     <Flow id="flow-b" bgProps={{ variant: BackgroundVariant.Lines, gap: [50, 50] }} />
     <Flow id="flow-c" bgProps={{ variant: BackgroundVariant.Cross, gap: [100, 50] }} />
+    <Flow id="flow-d" bgProps={{ variant: BackgroundVariant.TwinLines }} />
   </div>
 );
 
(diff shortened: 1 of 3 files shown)


moklick:
Hey @Elringus 

thanks for your help here! It's a good idea to reuse the lines background variant 👍 

I would like to suggest some changes:

- rename variant to "grid" or "gridLines"
- rename `twinGap` to `gridFactor` (5 would be gap * 5, and so on)
- remove `twinColor` and just use `color`
- remove `twinLineWidth` and always use 2*lineWidth or whatever looks reasonable

what do you think about this?

Elringus:
- I thought about "grid", but "lines" are also basically a grid in the end, which could be confusing. The key here is adding one more grid pattern with a different color, width and gap. Imho, optimal solution here would be replacing pattern enum with standalone background components for each pattern type, so that they could have their own props, but that's a breaking change.
 - Fine be me, I've also thought about it, but figured someone may want to use different x,y gaps for the second pattern, given the option is already available for the main pattern.
 - An option to set different color for the second pattern is critical for my use case; changing width won't look as clean.
 - Again, fine be me, as I don't change the width for the second pattern, only color; but may be desired in some other cases.

Elringus:
I've made an alternative solution via #2941, which is probably better that this one.

moklick:
good points. I would also prefer an API like this:

```jsx
<Background>
  <BackgroundPattern />
</Background>
```

maybe that's something for v12. 

closed in favour of #2941

PR #2941, opened 2023-03-20
allow mutliple backgrounds, add offset prop

Elringus:
This allows multiple background components by introducing `id` property for the `<Background/>` component, which is used to distinguish the svg patterns. Also adds `offset` property to allow alternating patterns (4th example on the screenshot). This is (hopefully) a cleaner alternative solution for #2940.

img


moklick:
This looks good, thanks!

PR #2944, opened 2023-03-21
add minimap inverse pan and zoom step

Elringus:
This adds `inversePan` and `zoomStep` optional properties to minimap allowing to inverse pan control (making it similar with how panning work in graph flow) and to control zoom speed (again, to make it more similar with flow zooming).


moklick:
Thanks @Elringus

at the beginning of next week we will do a release!

moklick:
Thanks for your contribution 🙏 it's added in v11.7.0

PR #2955, opened 2023-03-24
Feat/minimap inverse pan and zoom step

moklick:


moklick:
Thanks @Elringus 

at the beginning of next week we will do a release!

PR #2985, opened 2023-04-05
fix(key-press): mac issue keyup not triggered

VictorWinberg:
### Issue
When pressing incorrect keys (according to the given keys to watch), when letting go of all keys some keys will stay as they are still pressed, hence making the `isMatchingKey` faulty and not trigger.

### Reason
See https://stackoverflow.com/questions/27380018/when-cmd-key-is-kept-pressed-keyup-is-not-triggered-for-any-other-key

### Solution
When retrieving an up-event for Meta key, remove all pressed keys.

### DIFF
--- a/packages/core/src/hooks/useKeyPress.ts
+++ b/packages/core/src/hooks/useKeyPress.ts
@@ -75,8 +75,8 @@ export default (keyCode: KeyCode | null = null, options: UseKeyPressOptions = { 
         }
 
         // fix for Mac: when cmd key is pressed, keyup is not triggered for any other key, see: https://stackoverflow.com/questions/27380018/when-cmd-key-is-kept-pressed-keyup-is-not-triggered-for-any-other-key
-        if (event.key === "Meta") {
-          pressedKeys.current.clear()
+        if (event.key === 'Meta') {
+          pressedKeys.current.clear();
         }
 
         modifierPressed.current = false;
--- a/packages/core/src/hooks/useKeyPress.ts
+++ b/packages/core/src/hooks/useKeyPress.ts
@@ -14,7 +14,7 @@ const doc = typeof document !== 'undefined' ? document : null;
 
 // the keycode can be a string 'a' or an array of strings ['a', 'a+d']
 // a string means a single key 'a' or a combination when '+' is used 'a+d'
-// an array means different possibilites. Explainer: ['a', 'd+s'] here the
+// an array means different possibilities. Explainer: ['a', 'd+s'] here the
 // user can use the single key 'a' or the combination 'd' + 's'
 export default (keyCode: KeyCode | null = null, options: UseKeyPressOptions = { target: doc }): boolean => {
   const [keyPressed, setKeyPressed] = useState(false);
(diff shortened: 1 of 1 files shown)


packages/core/src/hooks/useKeyPress.ts
```diff
@@ -73,6 +73,12 @@ export default (keyCode: KeyCode | null = null, options: UseKeyPressOptions = {
         } else {
           pressedKeys.current.delete(event[keyOrCode]);
         }
+
+        // fix for Mac: when cmd key is pressed, keyup is not triggered for any other key, see: https://stackoverflow.com/questions/27380018/when-cmd-key-is-kept-pressed-keyup-is-not-triggered-for-any-other-key
+        if (event.key === "Meta") {
+          pressedKeys.current.clear()
+        }
```
bcakmakoglu, line 80: If this condition should be checked (I suggested this previously as well but didn't receive any feedback by @moklick or @chrtze yet, it might be that this is *not* the desired solution as a user could keyup the meta key while still holding another one, which in turn would cause other bugs again), why isn't it just added to the existing one?

```ts
if (isMatchingKey(keyCodes, pressedKeys.current, true) || event.metaKey) { ... }
```

Aside from that your code is not linted properly, the project uses single quotes and semi-colons.

VictorWinberg: > it might be that this is not the desired solution as a user could keyup the meta key while still holding another one

I thought of that myself, but couldn't find a reason you would not want to let go of the meta key, but still hold other buttons. I rather see that you can track "Meta+c" and "Meta+v", which is not possible at the moment, since it will keep the non-meta key in state.

E.g.:
```
keydown Meta (pressed keys: ["Meta"])
keydown x (pressed keys: ["Meta", "x"])
keyup x it NOT triggered
keyup Meta (pressed keys: ["x"])

keydown Meta (pressed keys: ["Meta", "x"])
keydown c (pressed keys: ["Meta", "x", "c"])
```

Hence the `["Meta+c"]` will not trigger, since it will keep "x" in the state until you manually press it or reload the page.

> why isn't it just added to the existing one

That is very true, however you can not use `event.metaKey` on up-events (why, idk 😓) but `event.key === "Meta"` could be moved to the above statement.

bcakmakoglu: Ah, true `metaKey` comes up as undefined in keyup (that's why the `modifierPressed` ref exists in this context) 😄 

I agree, in theory this would solve the issue but like I said, it might not be what the devs actually want to do.
Let's wait what their opinion on this issue is =)

VictorWinberg: @moklick Yes what I am testing is `const copyPressed = useKeyPress(['Meta+c']);` and `const pastePressed = useKeyPress(['Meta+v']);`. 

The issue is that for `Meta`-key on Mac it suppresses all other `keyup`-events, so any key you press down while holdning down the `Meta`-key, would be ignored. You still get the `keyup`-event from the `Meta`-key. So because I don't know what keys have been let-go during the `Meta`-press, I would like to remove all `pressedKeys`.

So for the example above, if you would press copy paste and let go of the `c` or `v` before the `Meta`-key, the `c` or `v` would stay as pressed keys.

E.g.
```
-- successful, since Meta had keyup first
keydown Meta (pressed keys: ["Meta"])
keydown c (pressed keys: ["Meta", "c"])
keyup Meta (pressed keys: ["c"])
keyup c (pressed keys: [""])

-- unsuccessful, since Meta had keyup last
keydown Meta (pressed keys: ["Meta"])
keydown c (pressed keys: ["Meta", "c"])
keyup c (pressed keys: ["Meta", "c"])
keyup Meta (pressed keys: ["c"])
```

So after I have pressed the copy, it would not allow me to paste, since it would not detect the paste `["Meta", "v"]` because it thinks that I am pressing `c` still.

VictorWinberg: Ping @moklick

moklick: thanks @VictorWinberg - this will be released with v11.8.0

PR #2993, opened 2023-04-08
fix(core): nodes getting dragged infinitely on auto-pan with translateExtent

bcakmakoglu:
# What's changed?

- Add `onBeforeTransform` arg to `panBy` action
   -  Maybe there's a more elegant solution for this?
- Avoid updating node pos for auto-pan triggers which do not transform the viewport, i.e. no pan actually happened

# Fixes

- nodes either stuttering when reaching translate extent or getting dragged infinitely into the background

### DIFF
--- a/.changeset/wise-pumas-juggle.md
+++ b/.changeset/wise-pumas-juggle.md
@@ -1,0 +1,6 @@
+---
+'@reactflow/core': patch
+---
+
+Fix auto-pan moving nodes infinitely into the background when translateExtent is used. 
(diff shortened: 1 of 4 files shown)


moklick:
What do you think about this solution #3029. `panBy` returns true if a transform change happened and we only update the nodes if it did.

bcakmakoglu:
That looks cleaner than what I proposed so I say yes! 😄 
Thanks :)

I'll close this PR then

PR #3007, opened 2023-04-20
feat(core): allow array of ids as updateNodeInternals arg

bcakmakoglu:
# What's changed?

- In addition to a single id, also allow for an array of ids to be passed to `updateNodeInternals`

### DIFF
--- a/packages/core/src/hooks/useUpdateNodeInternals.ts
+++ b/packages/core/src/hooks/useUpdateNodeInternals.ts
@@ -6,12 +6,16 @@ import type { UpdateNodeInternals } from '../types';
 function useUpdateNodeInternals(): UpdateNodeInternals {
   const store = useStoreApi();
 
-  return useCallback<UpdateNodeInternals>((id: string) => {
+  return useCallback<UpdateNodeInternals>((id: string | string[]) => {
     const { domNode, updateNodeDimensions } = store.getState();
     const nodeElement = domNode?.querySelector(`.react-flow__node[data-id="${id}"]`) as HTMLDivElement;
 
     if (nodeElement) {
-      requestAnimationFrame(() => updateNodeDimensions([{ id, nodeElement, forceUpdate: true }]));
+      const updateIds = Array.isArray(id) ? id : [id];
+
+      updateIds.forEach((id) => {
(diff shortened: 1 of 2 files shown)


packages/core/src/hooks/useUpdateNodeInternals.ts
```diff
@@ -6,12 +6,18 @@ import type { UpdateNodeInternals } from '../types';
...
     const { domNode, updateNodeDimensions } = store.getState();
     const nodeElement = domNode?.querySelector(`.react-flow__node[data-id="${id}"]`) as HTMLDivElement;
 
     if (nodeElement) {
-      requestAnimationFrame(() => updateNodeDimensions([{ id, nodeElement, forceUpdate: true }]));
+      const updateIds = Array.isArray(id) ? id : [id];
+
+      requestAnimationFrame(() => {
+        updateIds.forEach((id) => {
+          updateNodeDimensions([{ id, nodeElement, forceUpdate: true }]);
```
moklick, line 18: the nodeElement would always be the same node. You also need to call `domNode?.querySelector(`.react-flow__node[data-id="${id}"]`)` multiple times. It's also possible to pass an array with ids and nodeElements to `updateNodeDimensions`.

moklick (changes requested):

packages/core/src/hooks/useUpdateNodeInternals.ts
bcakmakoglu: Ah yeah, that makes sense - I'll change that
Thanks for keeping an eye out, I didn't think too much doing this PR and it shows 🤣

moklick: querySelector should go in the `forEach` so that `nodeElement` matches the actual node.

moklick (changes requested):

packages/core/src/hooks/useUpdateNodeInternals.ts
bcakmakoglu: Done 👍

moklick:
Thanks @bcakmakoglu 🙌

PR #3034, opened 2023-05-05
feat(node-toolbar): add align prop

Noam3kCH:
adds the ability to align the toolbar to the start or end of the node's dimensions


moklick:
thanks for your PR. We will check it :)

PR #3049, opened 2023-05-11
[fix] Update link to Cypress examples in README.md

conniepaula:
Link to cypress examples was taking users to a 404 page.
This PR updates the link so users are taken to the correct path, `examples/vite-app/cypress/`, instead of the incorrect one, `examples/cypress`.


moklick:
Thanks @conniepaula ✌️

Issue #3119, opened 2023-06-05
NodeResizer doesn't use latest onResize/onResizeStart/onResizeEnd callbacks


goliney:
### Describe the Bug

NodeResizer accepts onResize/onResizeStart/onResizeEnd props. If these callbacks change during the life of the component, only the first version of the callback is used.

### Your Example Website or App

https://codesandbox.io/s/distracted-hellman-4t7b0z?file=/ResizeRotateNode.js:817-828

### Steps to Reproduce the Bug or Issue

1. Initialize node resizer with `onResizeEnd` prop:  
```js
...
const [count, setCount] = useState(0);

const handleResizeEnd = useCallback(() => {
  console.log("Resize end count:", count);
  setCount((prev) => prev + 1);
}, [count]);

return (
  <div>
    <NodeResizer onResizeEnd={handleResizeEnd} />
  </div>
 );
```

2. Resize the node a couple of times
3. `handleResizeEnd` will print the same message every time: `Resize end count: 0`

### Expected behavior

The counter should be updated:
```
Resize end count: 0
Resize end count: 1
Resize end count: 2
...
```

### Screenshots or Videos

https://github.com/wbkd/react-flow/assets/2017344/98bf9284-d3e9-4265-bd08-e9327822fdb4

### Platform

- OS: macOS
- Browser: Chrome
- Version: 113.0.5672.63

### Additional context

This is happening, because `onResize/onResizeStart/onResizeEnd` aren't listed as the dependencies of this `useEffect`:
https://github.com/wbkd/react-flow/blob/2e1f360817915f84743b6e0ca390f57f1bb2a57a/packages/node-resizer/src/ResizeControl.tsx#L55-L224

goliney:
The issue is fixed in `11.7.3` (https://github.com/wbkd/react-flow/pull/3151). Thank you!

moklick:
Thanks @goliney :)

PR #3120, opened 2023-06-05
[Bugfix] Include onResize/onResizeStart/onResizeEnd in the list of useEffect dependencies

goliney:
Fixes https://github.com/wbkd/react-flow/issues/3119


moklick:
good catch! Thanks for the fix.

PR #3145, opened 2023-06-16
Draft: add a way to add intial state for SSR

brandonFaf:
Trying to add initial state to maybe help with SSR


moklick:
This will be possible in v12.

moklick:
V12 is still beta, but it's already published on npm. All information can be found here: https://github.com/xyflow/xyflow/discussions/3764

PR #3173, opened 2023-06-25
Allow controlling offset scale of minimap

Elringus:
Sorry for not creating a discussion first, but this one is pretty trivial.

Due to the `5` hard-coded value here:

https://github.com/wbkd/react-flow/blob/993a778b80cc1e80a47983ed75407b579313a73c/packages/minimap/src/MiniMap.tsx#L70

— minimap mask has a ~5px border even when the viewport is perfectly aligned with the map and all the nodes are inside: 

img

This doesn't work for me, as I need the mask to show only when nodes are outside the view or the view aspect ratio differs from map's, eg:

https://github.com/wbkd/react-flow/assets/2056864/bd93c88d-98aa-4b1c-aa16-6ccb88cd7379

Setting the value to 1 resolved the issue (and doesn't seem to break anything), but I guess it was added for a reason, hence the added prop with the old default value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,11 @@ A highly customizable React component for building interactive graphs and node-b
 
 ---
 
+## 🚨 Upcoming Changes 
+
+The main branch (v11) is now in a feature freeze. The next version is being developed in the [xyflow branch](https://github.com/wbkd/react-flow/tree/xyflow).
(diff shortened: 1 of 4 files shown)


moklick:
Good idea! Thanks for your PR :)

PR #3210, opened 2023-07-10
Fix nodeTypes warning shown on shallow equals

avin-kavish:
Currently a warning about memoizing the the nodeTypes object  is shown when the types are shallowly equal. Instead it should be shown when the consequent nodeTypes are **not** shallowly equal.

### DIFF
--- a/packages/core/src/container/ReactFlow/utils.ts
+++ b/packages/core/src/container/ReactFlow/utils.ts
@@ -16,7 +16,7 @@ export function useNodeOrEdgeTypes(nodeOrEdgeTypes: any, createTypes: any): any 
   const typesParsed = useMemo(() => {
     if (process.env.NODE_ENV === 'development') {
       const typeKeys = Object.keys(nodeOrEdgeTypes);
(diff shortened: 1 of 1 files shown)


avin-kavish:
Not it doesn't. It gives a warning when you send a static object declared at the module level.

avin-kavish:
Think about it. Even the static objects declared at the module level are also shallow equals. 

```t
const nodeTypes =  { custom: NodeType }

shallow({ custom: NodeType }, { custom: NodeType }) // true
shallow(nodeTypes, nodeTypes) // true
```

avin-kavish:
I think you are not taking hot reloading into account or something. I can guarantee you the second case gets invoked during hot reloads. Maybe the useMemo always runs on a hot reload, even if the values are referentially equal.

moklick:
With `shallow` we are not checking the nodeTypes, but just the keys. When `useMemo` detects a change but the keys stay the same, we are showing the error message, because we assume that it was done by mistake.

Issue #3248, opened 2023-07-22
panOnScroll breaks onMoveEnd


frankandrobot:
### Describe the Bug

* When `panOnScroll` is set to false, `onMoveEnd` correctly fires whenever the user pans or zooms the graph. 
* however, once `panOnScroll is set to true, `onMoveEnd` no longer fires when the user pans or zooms the graph. It only fires when the user explictly clicks on the graph. 
*

### Steps to Reproduce the Bug or Issue

1.  Given this simple usage, enable `panOnScroll`

```ts
   <TopologyZoomProvider value={zoom}>
        <ReactFlow
          nodes={nodes}
          edges={edges}
          nodeTypes={reactFlowNodeTypes}
          edgeTypes={reactFlowEdgeTypes}
          fitView
          attributionPosition="top-right"
          onlyRenderVisibleElements
          onMoveEnd={(_, viewport) => {
            console.log("hello")
            setZoom(viewport.zoom);
          }}
          /** panOnScroll + selectionOnDrag makes it work like figma */
          panOnScroll
          selectionOnDrag
          minZoom={0.1}
        >
          <Controls />
        </ReactFlow>
```

### Expected behavior

* onMoveEnd should fire for any panning or zooming

### Platform

- OS: [macos]
- Browser: [firefox]
- Version: latest reactflow (as of today)

moklick:
thanks I will check it 🧐

moklick:
fixed in v11.8.0

PR #3265, opened 2023-07-26
fix: #3248 panOnScroll breaks onMoveEnd

checkomkar:
#3248


moklick:
Thanks but this doesn't work. It fires a move end on every scroll and doesn't handle `onMoveStart` and `onMove`.

moklick:
I found a solution for this #3290

Issue #3256, opened 2023-07-25
onMoveEnd triggered on right click even that mouse is not moved at all


lancemao:
### Describe the Bug

onMoveEnd triggered on right click even that mouse is not moved at all. Occurs only for the first time

### Steps to Reproduce the Bug or Issue

1. add a onMoveEnd callback
2. add contextmenu event by using `addEventListener("contextmenu", onContextMenu)` on document
3. start ReactFlow app and right click anywhere with no mouse moving
4. onMoveEnd is called

### Expected behavior

onMoveEnd not triggered

### Platform

- OS: [Windows]
- Browser: [Chrome]
- Version: [114.0.5735.199]

checkomkar:
Can you provide more information about the issue, I was not able to reproduce this.

lancemao:
Hi, sorry for not providing enough info. Here is a complete


```javascript
import './App.css';
import ReactFlow, { Background, Controls, SelectionMode } from 'reactflow';
import 'reactflow/dist/style.css';

import React, {
  useEffect,
  useRef,
} from 'react';

const nodes = [
  {
    id: '1',
    data: { label: 'Hello' },
    position: { x: 0, y: 0 },
    type: 'input',
  },
  {
    id: '2',
    data: { label: 'World' },
    position: { x: 100, y: 100 },
  },
];

export default function App() {

  const reactFlowWrapper = useRef(null);

  useEffect(() => {
    reactFlowWrapper.current.addEventListener("contextmenu", onContextMenu);

    return () => {
      reactFlowWrapper.current.removeEventListener("contextmenu", onContextMenu);
    };
  });

  const onContextMenu = (e) => {
    e.preventDefault();
    console.log('onContextMenu')
  }

  const onMouseClick = (e) => {
    console.log('onMouseClick')
  }

  const onMoveEnd = (e) => {
    console.log('onMoveEnd')
  }

  return (
    <div className='app-root' ref={reactFlowWrapper}>
      <div style={{ height: '100%' }}>
        <ReactFlow nodes={nodes} onMoveEnd={onMoveEnd} panOnDrag={[2]}
          selectionOnDrag={true}
          onMove={onMouseClick}
          onSelectionStart={onMouseClick}
          selectionMode={SelectionMode.Partial}
        >
          <Background />
          <Controls />
        </ReactFlow>
      </div>
    </div>
  );
}
```

checkomkar:
[...]

PR #3268, opened 2023-07-27
Fix: #3256 for onMoveEnd triggered on right click even that mouse is not mov…

checkomkar:
#3256


lancemao:
when will this PR get merged?

moklick:
Thanks @checkomkar ! 

@lancemao fix gets published with the next release.

Issue #3205, opened 2023-07-07
[Bug]: pathOptions borderRadius does not work correctly when connecting two vertically aligned nodes


binhtran04:
### Describe the Bug

I use pathOptions borderRadius: 20 to smoothstep edge. However, when two nodes are vertically aligned, the edges are not rounded. I happened with my own project and I was able to replicate in a code sandbox https://codesandbox.io/s/fast-fast-nhhj4q?file=/App.js:2525-2581

### Your Example Website or App

https://codesandbox.io/s/fast-fast-nhhj4q?file=/App.js:2525-2581

### Steps to Reproduce the Bug or Issue

1. Define edgeDefaultOptions as follow 
```
{
  type: "smoothstep",
  pathOptions: {
    borderRadius: 20,
    offset: 40
  }, 
}
```
2. Define two node vertically allined 
3. Define an edge from right handle of top node to right handle of bottom node and vice versa from left handle bottom node to left handle top node

### Expected behavior

The edges should be rounded according to borderRadius pathConfigs option

### Screenshots or Videos

![image](https://github.com/wbkd/react-flow/assets/15939436/1817dbb0-98ad-4fd0-b377-cba1ae1b286a)

### Platform

- OS: [Linux]
- Browser: [Chrome]
- Version: [reactflow 11.7.4]

moklick:
That's a bug. Thanks for the report.

GenaRazmakhnin:
hey! @moklick @binhtran04
I faced the same issue, it's affected not only vertically aligned nodes but horizontal as well, 
looks like I fixed the issue, but I do not have a lot time to find better solution please review :)

moklick:
fixed in v11.8.0

tomtitherington:
Just updated to v11.8.2 and this is still a problem for me. The edge has no corner radius when each edge is aligned horizontally with respect to each other but _also when they're not_ - see screenshots below. 

<img width="1035" alt="Screenshot 2023-08-16 at 12 32 33" src="https://github.com/wbkd/react-flow/assets/30439911/a1406aa0-0a66-4155-b7fa-2a7382b7f682">
<img width="1040" alt="Screenshot 2023-08-16 at 12 32 46" src="https://github.com/wbkd/react-flow/assets/30439911/1d6c00f1-35e4-45fd-b1e3-3b59d75eabee">

moklick:
@tomtitherington I'll check that

moklick:
[...]

PR #3275, opened 2023-07-29
fix(smoothedge): the edge corners between two vertically or horizontally aligned nodes are not rounded

GenaRazmakhnin:
https://github.com/wbkd/react-flow/issues/3205


moklick:
Hey @GenaRazmakhnin 

thanks for your help :) I will check it asap.

Issue #3272, opened 2023-07-28
PanOnScrollMode.Horizontal seems to be broken


bcrudolp:
### Describe the Bug

I set `zoomOnScroll={false}` and then set the `panOnScroll` prop on my `ReactFlow` and set the `panOnScrollMode` using ternary logic according to a state value that is modified when I hold down `Shift` (to intuitively match behavior of a plain old HTML web page section's scrolling) but it doesn't work. The logic: `shiftHeld ? PanOnScrollMode.Horizontal : PanOnScrollMode.Vertical`

I searched on the Discussions tab and found this [discussion](https://github.com/wbkd/react-flow/discussions/1288). I accordingly updated the `panOnDrag` prop (newer version equivalent of `paneMoveable` mentioned in that thread) to be false when `Shift` is held down. It still doesn't work. Am I missing something? Can anyone point me to a place that demonstrates a working example that causes horizontal panning when `Shift` is held down?

Note: I can verify my state is being conveyed correctly because it no longer pans in the vertical direction when I hold down `Shift`, but neither is it panning horizontally like it's supposed to.

### Steps to Reproduce the Bug or Issue

Attempt to configure a `ReactFlow` to panOnScroll horizontally with react-flow@11.7.0

### Expected behavior

I wanted to have the diagram pan horizontally like simple web pages scroll horizontally when you perform a `Shift`+`<mouse-wheel>` action.

### Screenshots or Videos

![ReactFlowPanOnScrollHorizontalProps](https://github.com/wbkd/react-flow/assets/97194517/494cd81f-5ef9-4a5a-b3f0-4f0f3af1a38c)

### Platform

- OS: Windows 10 Enterprise
- Browser: Chrome
- Version: Version 115.0.5790.110 (Official Build) (64-bit)

### Additional context

I have no idea what version of react-flow is used for this [demo app](https://react-flow-dev.netlify.app/interaction) and I doubt they have the specific key handling , but it at least demonstrates that PanOnScrollMode.Horizontal does not achieve out-of-the-box what I intuitively thought it should (that is, `Shift`+`mouse-wheel` panning side-to-side).

checkomkar:
Hey @bcrudolp I have raised a PR for this issue you can test it by checking out in [this](https://github.com/wbkd/react-flow/pull/3276) branch. Thanks.

bcrudolp:
[...]

PR #3276, opened 2023-07-29
Fix: issue #3272, horizontal scroll not working

checkomkar:
#3272


moklick:
this is fixed in v11.8.1 - you can now scroll on windows machines by pressing Shift (when `panOnScroll` is true and `panOnScrollMode` is "Free" or "Horizontal").

PR #3293, opened 2023-08-04
fix(zoom): handle windows zoom with ctrl

moklick:


moklick:
Hi @ArturMakly @juspky @Vochsel @jackfishwick 

could you check if this fixes the fast zooming on windows?

PR #3294, opened 2023-08-04
🐛  Fixes a bug where 'handleNodeClick' could runtime error if the node had already been removed.

hayleigh-dot-dev:
Two changes:

- `handleNodeClick` no-ops if the target node doesn't exist.
- A new error is reported:

```js
`Node with id "${id}" does not exist, it may have been removed. This can happen when a node is deleted before the "onNodeClick" handler is called. If this wasn't on purpose, please defer deleting this node until the "onNodeClick" handler is called by using "requestAnimationFrame" or "setTimeout".`
```

This can come up if a custom node has a delete button, for example, and the user _also_ has a `onNodeClick` handler set up on the `<ReactFlow>` component.

### DIFF
--- a/packages/core/src/contants.ts
+++ b/packages/core/src/contants.ts
@@ -18,5 +18,5 @@ export const errorMessages = {
   error010: () => 'Handle: No node id found. Make sure to only use a Handle inside a custom Node.',
   error011: (edgeType: string) => `Edge type "${edgeType}" not found. Using fallback type "default".`,
   error012: (id: string) =>
-    `Node with id "${id}" does not exist, it may have been removed. This can happen when a node is deleted before the "onNodeClick" handler is called. If this wasn't on purpose, please defer deleting this node until the "onNodeClick" handler is called by using "requestAnimationFrame" or "setTimeout".`,
+    `Node with id "${id}" does not exist, it may have been removed. This can happen when a node is deleted before the "onNodeClick" handler is called.`,
 };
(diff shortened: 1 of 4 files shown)


hayleigh-dot-dev:
See https://github.com/wbkd/react-flow-docs/pull/131 for the companion docs update.

moklick:
I think it's fine when this part ("This can happen when a node is deleted before the "onNodeClick" handler is called. If this wasn't on purpose, please defer deleting this node until the "onNodeClick" handler is called by using "requestAnimationFrame" or "setTimeout".") of the error message can only be found on the troubleshooting page.

packages/core/src/contants.ts
```diff
@@ -17,4 +17,6 @@ export const errorMessages = {
     }", edge id: ${edge.id}.`,
   error010: () => 'Handle: No node id found. Make sure to only use a Handle inside a custom Node.',
   error011: (edgeType: string) => `Edge type "${edgeType}" not found. Using fallback type "default".`,
+  error012: (id: string) =>
+    `Node with id "${id}" does not exist, it may have been removed. This can happen when a node is deleted before the "onNodeClick" handler is called. If this wasn't on purpose, please defer deleting this node until the "onNodeClick" handler is called by using "requestAnimationFrame" or "setTimeout".`,
```
moklick, line 21: "Node with id "${id}" does not exist, it may have been removed." - this is enough for the error message

moklick (changes requested):

Issue #3432, opened 2023-09-22
Error occurs when attempting to delete a node that has child nodes under certain conditions


asfman:
### Describe the Bug

When node A in reactflow has the `parentNode` property, if the node B pointed to by `parentNode` has an index position greater than node A in the `nodes` array, an error will be thrown when node B is deleted: `throw new Error(`Parent node ${node.parentNode} not found`)`.

### Example Website

https://openai-red.vercel.app/reactflow

### Steps to Reproduce the Bug or Issue

1. go to https://openai-red.vercel.app/reactflow
2. click "change nodes order".
3. delete "parent node 1"
4. reproduce the error "Error: Parent node 1 not found"

### Expected behavior

No matter whether I change the order of nodes or not, deletion should not result in an error.

### Screenshots or Videos

https://github.com/wbkd/react-flow/assets/169468/b39b098e-a18f-4431-b398-accda24c2e30

The error occurs when I change the order of nodes and place the parent node after the child node.

### Platform

- OS: macOS
- Browser: Chrome
- Version: Version 116.0.5845.179

asfman:
https://github.com/wbkd/react-flow/pull/3433

moklick:
As explained in the docs, the nodes array has to be sorted when you are working with subflows.

PR #3433, opened 2023-09-22
fix(delete-node): parentHit logic

asfman:
When node A in reactflow has the parentNode property, if the node B pointed to by parentNode has an index position greater than node A in the nodes array, an error will be thrown when node B is deleted: throw new Error(Parent node ${node.parentNode} not found).

 issue:  https://github.com/wbkd/react-flow/issues/3432

Ensure that parentHit is right regardless of the order of nodes.
```javascript
    const nodesToRemove = getNodes().reduce<Node[]>((res, node) => {
      let parentHit = false;
      if(!nodeIds.includes(node.id)) {
        let currentNode = node;
        while (currentNode && currentNode.parentNode) {
          if (nodeIds.includes(currentNode.parentNode)) {
            parentHit = true;
            break;
          }
          currentNode = getNodes().find((nd) => nd.id === currentNode.parentNode);
        }
      };
      const deletable = typeof node.deletable === 'boolean' ? node.deletable : true;
      if (deletable && (nodeIds.includes(node.id) || parentHit)) {
        res.push(node);
      }

      return res;
    }, []);

```

### DIFF
--- a/packages/core/src/hooks/useReactFlow.ts
+++ b/packages/core/src/hooks/useReactFlow.ts
@@ -130,7 +130,17 @@ export default function useReactFlow<NodeData = any, EdgeData = any>(): ReactFlo
     const nodeIds = (nodesDeleted || []).map((node) => node.id);
     const edgeIds = (edgesDeleted || []).map((edge) => edge.id);
     const nodesToRemove = getNodes().reduce<Node[]>((res, node) => {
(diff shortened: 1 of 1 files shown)


moklick:
Hey @asfman 

thanks for the fix for this case but there are other things like positioning that won't work properly if the nodes array isn't sorted. As the docs point out, the nodes array has to be sorted.

PR #3435, opened 2023-09-23
fix(core): export handleParentExpand

bodinsamuel:
Hey,

I tried to use this logic in a custom data store but it's not exposed. 
It would be handy instead of copy pasting the code.

lmk


moklick:
sure thing!

PR #3486, opened 2023-10-06
add `U` (`Node.type`) param in more places

ryan-williams:
see discussion on https://github.com/wbkd/react-flow/pull/2923

I believe this is backwards-compatible, per https://github.com/wbkd/react-flow/pull/2923#issuecomment-1474161604

cc @slowyn, @bcakmakoglu, @moklick


moklick:
Hey @ryan-williams,

thanks for your PR but we won't merge it, because we are working on a next version where we want to simplify typing in general. We will only fix bugs in v11. I hope that we will have a v12 by beginning of 2024 :)

PR #3524, opened 2023-10-20
fix: calculation of target node center in examples

sescer:
Hello Team!

While reviewing the changes introduced in the recent commit for the `getNodeIntersection` function, I've identified a critical concern related to how the coordinates are being calculated for `x1` and `y1`. This miscalculation could lead to inaccuracies in the resultant intersection point.

**Issue:**
In the current implementation, `x1` and `y1` are calculated using the `targetNodePosition` coordinates plus `w` and `h` variables, which represent half of the `intersectionNode`'s dimensions. This calculation assumes that the dimensions of `targetNode` and `intersectionNode` are the same, which might not always be the case. Adding `w` and `h` from the `intersectionNode` to the `targetNode`'s position offsets the target's actual center point, likely leading to an incorrect intersection calculation.

**Proposed Solution:**
To rectify this, we need to calculate `x1` and `y1` based on the actual dimensions of the `targetNode`. Extract the width and height properties from the `targetNode`. Calculate the center of the `targetNode` using its dimensions.

Thank you!

### DIFF
--- a/examples/vite-app/src/examples/EasyConnect/utils.tsx
+++ b/examples/vite-app/src/examples/EasyConnect/utils.tsx
@@ -9,15 +9,22 @@ function getNodeIntersection(intersectionNode: Node, targetNode: Node) {
     height: intersectionNodeHeight,
     positionAbsolute: intersectionNodePosition,
   } = intersectionNode;
-  const targetPosition = targetNode.positionAbsolute!;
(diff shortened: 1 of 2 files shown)


moklick:
that makes sense, thanks! I updated the docs accordingly.

Issue #2980, opened 2023-04-04
Support Mutiple useOnSelectionChange Hook


zta7:
### Please describe the feature that you want to propose

I can't use `useOnSelectionChange `in different components, it will be merged into one in the end. I think this hook should allow me to monitor selection changes in different components.

moklick:
Good catch. We will check how/ if we can support this.

jrdn91:
Any update on this? I'm running into this exact issue currently

juspky:
I'm running into this issue too.

moklick:
Thanks for your help @juspky . This is released in 11.10.1

PR #3535, opened 2023-10-24
feat(react): Support mutiple useOnSelectionChange hook subscribers

juspky:
Solves: [ Support Mutiple useOnSelectionChange Hook #2980 ](https://github.com/wbkd/react-flow/issues/2980)

### DIFF
--- a/examples/react/src/examples/UseOnSelectionChange/index.tsx
+++ b/examples/react/src/examples/UseOnSelectionChange/index.tsx
@@ -1,5 +1,5 @@
-import { useCallback } from 'react';
-import {
+import { useCallback, useState } from 'react';
+import  {
   ReactFlow,
   addEdge,
   ReactFlowProvider,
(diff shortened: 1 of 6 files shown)


packages/core/src/hooks/useOnSelectionChange.ts
```diff
@@ -4,15 +4,20 @@ import { useStoreApi } from './useStore';
 import type { OnSelectionChangeFunc } from '../types';
 
 export type UseOnSelectionChangeOptions = {
-  onChange?: OnSelectionChangeFunc;
+  onChange: OnSelectionChangeFunc;
```
juspky, line 7: I made this not optional because it makes no sense to pass undefined here.

moklick:
Thanks @juspky! Because it will take some time until we publish v12, I will also add your changes to the next v11 release.

juspky:
Amazing!

PR #3544, opened 2023-10-25
Refactor testing

peterkogo:


moklick:
This is so good! We can actually use the same tests for Svelte Flow and React Flow 🤯

PR #3555, opened 2023-10-28
getIntersectingNodes and isNodeIntersecting: Change nodeOrRect.width calls to nodeRect.width calls since the format may be undefined

### LINKED ISSUE #3554: getIntersectingNodes never detects an intersection when partially=false
### Describe the Bug

When calculating intersections, getIntersectingNodes tries to calculate the area of the input node with `nodeOrRect.width! * nodeOrRect.height!` but `nodeOrRect` may not have the width and height on it and so may return NaN->false. This may also affect isNodeIntersecting since it uses the same calculation. 

```ts
  const getIntersectingNodes = useCallback<Instance.GetIntersectingNodes<NodeData>>(
    (nodeOrRect, partially = true, nodes) => {
      const [nodeRect, node, isRect] = getNodeRect(nodeOrRect);

      if (!nodeRect) {
        return [];
      }

      return (nodes || store.getState().getNodes()).filter((n) => {
        if (!isRect && (n.id === node!.id || !n.positionAbsolute)) {
          return false;
        }

        const currNodeRect = nodeToRect(n);
        const overlappingArea = getOverlappingArea(currNodeRect, nodeRect);
        const partiallyVisible = partially && overlappingArea > 0;

        return partiallyVisible || overlappingArea >= nodeOrRect.width! * nodeOrRect.height!;
      });
    },
    []
  );
```

The thing is getOverlappingArea successfully does its job, because it's using the result of getNodeRect which is checked and guaranteed to be a rectangle type with x,y,width,height on. So the fix is probably a 1-liner

_No response_

Pass a node retrieved from useNodesState to getIntersectingNodes with partially=false and it will never return anything because that node is not a rect and may not have width/height set in all cases.

### Expected behavior

getIntersectingNodes retrieves the rect if necessary (it already does) and uses it for all calculations (it doesn't do this)

### Screenshots or Videos

_No response_

### Platform

- OS: [macOS]
- Browser: [Chrome]
- Version: [11.9.4]

### Additional context

_No response_

Nick-Lucas:
Attempt to fix #3554

### DIFF
--- a/examples/vite-app/src/examples/EasyConnect/utils.tsx
+++ b/examples/vite-app/src/examples/EasyConnect/utils.tsx
@@ -9,15 +9,22 @@ function getNodeIntersection(intersectionNode: Node, targetNode: Node) {
     height: intersectionNodeHeight,
     positionAbsolute: intersectionNodePosition,
   } = intersectionNode;
-  const targetPosition = targetNode.positionAbsolute!;
+
+  const {
+    width: targetNodeWidth,
+    height: targetNodeHeight,
+    positionAbsolute: targetNodePosition,
+  } = targetNode;
 
   const w = intersectionNodeWidth! / 2;
   const h = intersectionNodeHeight! / 2;
+  const targetW = targetNodeWidth! / 2
+  const targetH = targetNodeHeight! / 2
 
   const x2 = intersectionNodePosition!.x + w;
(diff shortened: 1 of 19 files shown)


moklick:
Thanks @Nick-Lucas ! This will be released with #3585

PR #3583, opened 2023-11-08
fix(react,svelte): allow panning by scroll when pan activation key is pressed

bcakmakoglu:
# What's Changed? 

- Allow panning by scroll when `panActivationKeyCode` is pressed

# Notes

- Idk if this is even wanted or if the current behavior is intentional, either way here would be a proposal if a change is wanted


moklick:
This makes sense :) Thanks! I will add it to v11 too.

PR #3600, opened 2023-11-10
refactor(system,panzoom)!: return promise from transitioned panzoom ac…

bcakmakoglu:
# What's Changed?

- Return a promise from panZoom actions that allow for transitions
   - Promise is resolved when the `onEnd` callback is triggered by the transition
- Allows users to react to viewport transitions without arbitrary waiting times like `setTimeout` etc.

# Notes

- I'd think this would constitute a breaking change, wouldn't it? So this, if even desired, should probably go into v12? 😄

### DIFF
--- a/.changeset/slimy-impalas-yell.md
+++ b/.changeset/slimy-impalas-yell.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/system': major
+---
+
+Return Promise from transitionable panzoom actions if a transition is applied. Promise is resolved when transition ends.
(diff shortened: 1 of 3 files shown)


moklick:
I don't want to add any more stuff to the React Flow 12 release 😅 but I it would be nice to be able to wait for the viewport helper functions. There are some more function where we would need to add this:

- setViewport
- fitView ?
- setCenter
- fitBounds

bcakmakoglu:
I think this "feature" isn't that important so it can wait for a future release for sure.
At this point this PR is kinda old, it'd need an update anyway 😄

PR #3601, opened 2023-11-10
refactor(system,styles)!: use translate to align handles to center of its side

bcakmakoglu:
# What's Changed?

- Remove `4px` offsets for handles and replace them with `translate` and `50%`
   - Aligns all handles, regardless of size, properly to the center of the side their own

# Notes

- This would probably count as another breaking change, as existing users styles might be affected by this change


moklick:
thanks!

Issue #3625, opened 2023-11-15
Panning the minimap is very difficult if you have a long graph


mirisony:
### Describe the Bug

If you have a very long graph, panning the minimap is very difficult. Almost impossible to use.
Only if you move the mouse pixel by pixel you'll be able to move the map and follow it, but the movement is unnatural.
It's as if the white square is "running away" from the cursor instead of following it.

### Your Example Website or App

https://codesandbox.io/s/8jhksy?file=/App.js&utm_medium=sandpack

### Steps to Reproduce the Bug or Issue

Zoom in a bit and try panning the minimap to navigate the graph.

### Expected behavior

White square should follow the mouse to enable a nice panning experience and an easy way to navigate the graph.

### Platform

- OS: macOS, Windows
- Browser: Chrome
- Version: 11.10.1

### Additional context

Opened a PR with a fix: https://github.com/xyflow/xyflow/pull/3626

mirisony:
Not sure how to connect the issue with my PR, but just in case - here it is:
https://github.com/xyflow/xyflow/pull/3626

Thanks! :)

moklick:
fixed in React Flow 12!

PR #3626, opened 2023-11-15
fix: Minimap panning fix. issue #3625

mirisony:
Fix to issue https://github.com/xyflow/xyflow/issues/3625


moklick:
Thanks @mirisony we will check it!

moklick:
Thanks! Works better for larger flows now :)

PR #3632, opened 2023-11-16
refactor(edges): simplify edge rendering

moklick:
This PR changes the rendering of edges. We are not grouping edges by zIndex anymore but render each edge in its own SVG. This removes the edgeTree creation code and simplifies the rendering. A nice side effect for React Flow is, that edges won't loose their local component state when zIndex changes, because the DOM tree stays the same.


bcakmakoglu:
Nice!

PR #3636, opened 2023-11-17
Allow on:connect to be triggered on source and target handles

PapaTemporal:
In the current implementation of SvelteFlow, only the drag start handle's on:connect is called when connecting two nodes. This results in an issue as seen in the gif below where HTTP REQUEST has an on:connect that is only called when the connection is started from HTTP REQUEST.

connectissue

This implementation allows for both, the source and destination, on:connects to be called. Furthermore, it includes the ability to call on:disconnect on any handle affected by an edge that is deleted.

workingconn

### DIFF
--- a/packages/svelte/src/lib/components/Handle/Handle.svelte
+++ b/packages/svelte/src/lib/components/Handle/Handle.svelte
@@ -106,14 +106,7 @@
   }
 
   function onTargetConnect(event: CustomEvent) {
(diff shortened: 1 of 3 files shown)


moklick:
With the latest released, this issue is fixed 0.0.29 - with a bunch of other things :) https://svelteflow.dev/whats-new/2023-12-13

PR #3668, opened 2023-11-25
[Perf: Selection, Dragging] Don’t rerender NodeRenderer and MiniMapNodes unless necessary

iamakulov:
This PR ensures that when you’re dragging a node around the screen (or selecting it, or updating it in any other way), `NodeRenderer` (and `MiniMapNodes`) don’t rerender.

This prevents us from re-running [this loop](https://github.com/xyflow/xyflow/blob/8fe8f73ed3d7fc2a14b2ecbefffb1b988b2edaf8/packages/react/src/container/NodeRenderer/index.tsx#L78-L149) N times (N being the number of nodes) on every store update. With 500 nodes, it gets surprisingly expensive (20-40 ms with CPU throttling)!

(You can review this PR commit-by-commit.)

#### Store-level changes

To make this possible, I had to make two significant design changes (both in 24390162e08867a7c68c309a0ebcd52dafd6d425).

**First,** I had to split the `Node` term into two: `InternalNode` and `UserProvidedNode`.
- `UserProvidedNode` means a node object that the user stores in their state and provides via `nodes`, `defaultNodes`, or `api.setNodes()`.
- `InternalNode` means a node extended from the user-provided one with additional properties (`computed` and `[internalSymbol]`). 

Note that this change is only 70% ready: this PR introduces these terms only in the React store and system store. If I was in charge of shipping this, I’d also update types; make `api.getNodes()` convert back to a user-provided node (by stripping internal properties); etc.

**Second,** I had to introduce a check that avoid recalculating the internal node if the user-provided node hasn’t changed:

https://github.com/iamakulov/xyflow/blob/d4484be0c39abd396c506de1be69c5d460279873/packages/system/src/utils/store.ts#L79

This was critical for memoization. Previously, any time the `nodes` prop would update, we’d regenerate every node object. Now, we only regenerate the changed ones.

However, this is a potential breaking change as it makes React Flow ignore the following kind of updates (previously, they were respected, even if not officially supported):

```js
const [nodes, setNodes] = useState([...]);

useEffect(() => {
  setNodes((nodes) => {
    nodes[5].draggable = false;
    return nodes
  });
}, []);

return <ReactFlow nodes={nodes} />
```

#### Component-level changes

With store changes implemented, `Node`s won’t regenerate anymore unless they actually change. This means we can now write memoization around node objects. We do this in 0186b661611347869291f2f941e8f387ce55a76c and 4fd321a8ff074a5837020ec2d4c83aaf4a678e2a for `NodeRenderer` (+ 220380685392d92d09a62cd7c4655b9ac45f6e40 for `MiniMapNodes`).

The above commits:

- split `NodeRenderer` (and `MiniMapNodes`) into two components instead of one
- make sure `NodeRenderer` (and `MiniMapNodes`) only rerender on _node ID_ changes
- make the child component rerender when its (and only its) node updates

As a result, if, before, we were continuously rerendering the whole `NodeRenderer` every time you’d drag a single node:

CleanShot 2023-11-25 at 02 40 17@2x

now we only rerender a concrete `NodeComponent`:

CleanShot 2023-11-25 at 02 42 07@2x

### Performance wins

This PR makes dragging 2-3× cheaper, from 60-75 ms per drag event to 25 ms. Compare: [trace before](https://trace.cafe/t/Zf5vGawLLA) · [trace after](https://trace.cafe/t/5gaCjjCaCY).

Selection also gets ~1.5× cheaper, _on top_ of https://github.com/xyflow/xyflow/pull/3667 (from 75-80 ms to 45-50 ms). Compare: [trace at PR 3667](https://trace.cafe/t/ViWgw3mOwC) · [trace with both PRs applied](https://trace.cafe/t/LSZkMzr9Du)

### DIFF
--- a/packages/react/src/hooks/useVisibleNodes.ts
+++ b/packages/react/src/hooks/useVisibleNodes.ts
@@ -1,21 +1,33 @@
-import { useCallback } from 'react';
 import { getNodesInside } from '@xyflow/system';
+import { createSelector } from 'reselect';
+import { shallow } from 'zustand/shallow';
 
 import { useStore } from '../hooks/useStore';
 import type { Node, ReactFlowState } from '../types';
 
-function useVisibleNodes(onlyRenderVisible: boolean) {
-  const nodes = useStore(
-    useCallback(
-      (s: ReactFlowState) =>
-        onlyRenderVisible
-          ? getNodesInside<Node>(s.nodes, { x: 0, y: 0, width: s.width, height: s.height }, s.transform, true)
-          : s.nodes,
-      [onlyRenderVisible]
-    )
-  );
+const visibleNodeIdsSelector = createSelector(
+  [
+    (s: ReactFlowState) => s.nodes,
+    (s: ReactFlowState) => s.width,
+    (s: ReactFlowState) => s.height,
+    (s: ReactFlowState) => s.transform,
+    (_s: ReactFlowState, onlyRenderVisibleNodes: boolean) => onlyRenderVisibleNodes,
+  ],
+  (nodes, width, height, transform, onlyRenderVisible) =>
+    onlyRenderVisible
+      ? getNodesInside<Node>(nodes, { x: 0, y: 0, width, height }, transform, true).map((node) => node.id)
+      : nodes.map((node) => node.id),
+  {
+    memoizeOptions: {
+      resultEqualityCheck: shallow,
+    },
+  }
+);
 
-  return nodes;
+function useVisibleNodeIds(onlyRenderVisible: boolean) {
+  const nodeIds = useStore((s: ReactFlowState) => visibleNodeIdsSelector(s, onlyRenderVisible));
+
+  return nodeIds;
 }
 
-export default useVisibleNodes;
+export default useVisibleNodeIds;
--- a/packages/system/src/types/nodes.ts
+++ b/packages/system/src/types/nodes.ts
@@ -40,6 +40,10 @@ export type NodeBase<T = any, U extends string | undefined = string | undefined>
     z?: number;
     handleBounds?: NodeHandleBounds;
     isParent?: boolean;
+    /** Holds a reference to the original node object provided by the user
+     * (which may lack some fields, like `computed` or `[internalSymbol]`. Used
+     * as an optimization to avoid certain operations. */
+    userProvidedNode: WeakRef<NodeBase>;
   };
 };
 
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -57,6 +57,7 @@
     "d3-drag": "^3.0.0",
     "d3-selection": "^3.0.0",
     "d3-zoom": "^3.0.0",
+    "reselect": "^4.1.8",
     "zustand": "^4.4.0"
   },
   "peerDependencies": {
--- a/packages/react/src/additional-components/MiniMap/MiniMapNodes.tsx
+++ b/packages/react/src/additional-components/MiniMap/MiniMapNodes.tsx
@@ -1,21 +1,30 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { memo } from 'react';
+import { ComponentType, memo, MouseEvent } from 'react';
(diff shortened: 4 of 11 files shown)


packages/react/src/hooks/useVisibleNodes.ts
```diff
@@ -1,21 +1,33 @@
...
+  (nodes, width, height, transform, onlyRenderVisible) =>
+    onlyRenderVisible
+      ? getNodesInside<Node>(nodes, { x: 0, y: 0, width, height }, transform, true).map((node) => node.id)
+      : nodes.map((node) => node.id),
+  {
+    memoizeOptions: {
+      resultEqualityCheck: shallow,
+    },
+  }
+);
```
iamakulov, line 25: We can probably optimize this a little more like this:

```suggestion
const visibleNodeIdsSelector = createSelector(
  [
    (s: ReactFlowState) => s.nodes,
    (s: ReactFlowState, onlyRenderVisibleNodes: boolean) => onlyRenderVisibleNodes ? s.width : null,
    (s: ReactFlowState, onlyRenderVisibleNodes: boolean) => onlyRenderVisibleNodes ? s.height : null,
    (s: ReactFlowState, onlyRenderVisibleNodes: boolean) => onlyRenderVisibleNodes ? s.transform : null,
    (_s: ReactFlowState, onlyRenderVisibleNodes: boolean) => onlyRenderVisibleNodes,
  ],
  (nodes, width, height, transform, onlyRenderVisible) =>
    onlyRenderVisible
      ? getNodesInside<Node>(nodes, { x: 0, y: 0, width, height }, transform, true).map((node) => node.id)
      : nodes.map((node) => node.id),
  {
    memoizeOptions: {
      resultEqualityCheck: shallow,
    },
  }
);
```

Because we don’t depend on `width`/`height`/`transform` when `onlyRenderVisibleNodes` is false, we also don’t need to recalculate the selector when they change. This change will ensure we actually don’t.

packages/system/src/types/nodes.ts
```diff
@@ -40,6 +40,10 @@ export type NodeBase<T = any, U extends string | undefined = string | undefined>
     z?: number;
     handleBounds?: NodeHandleBounds;
     isParent?: boolean;
+    /** Holds a reference to the original node object provided by the user
+     * (which may lack some fields, like `computed` or `[internalSymbol]`. Used
+     * as an optimization to avoid certain operations. */
+    userProvidedNode: WeakRef<NodeBase>;
```
iamakulov, line 46: Using `WeakRef` to avoid permanently holding user-provided nodes (so if the user removes some nodes from their list, they can be garbage-collected). `WeakRef` is available [from Safari 14.1](https://caniuse.com/mdn-javascript_builtins_weakref).

iamakulov:
**Note:** upon further testing, this seems to increase the initial rendering costs ~1.5 times (in development, in the stress test) – because now, for every node, we also have to render the `NodeComponentWrapper` (= more components to render).
- The increase might be smaller in the prod build
- We can maybe reduce this increase by doing some custom memoization instead of relying on `memo(NodeComponentWrapper)`. But that will make it harder to manage memory (we’ll need to learn when a node is removed so we can remove the cached entry for that node)

iamakulov:
> The increase might be smaller in the prod build

Oh yeah – in a production build, the difference is <10% (1004 ms of [scripting time](https://github.com/xyflow/xyflow/assets/2953267/9f8c6752-08be-4dc3-b45a-6ff1c4d2d439) [at `next`](https://trace.cafe/t/5rNL74cTKX) → 1093 ms [with this PR and #3667 applied](https://trace.cafe/t/nOa80sxHgA)).

My next PR should somewhat reduce the init time and bring that value back.

moklick:
Hey @iamakulov 

thanks again for your great work and the super detailed PR descriptions! Mapping just over the node ids was a great idea and really pushed the performance of dragging nodes. We took a lot of parts from this PR and it helped us to create some new refactoring ideas. After working on your changes we got rid of the `NodeComponentWrapper` and the `wrapNode` HOC and only use a simple `NodeWrapper` component. In addition we could remove lot of util code about `nodeTypes` and `edgeTypes`. 

This PR was continued in #3727

PR #3677, opened 2023-11-28
[Perf: Init Costs] Batch initial store updates together + more

iamakulov:
This PR does two things:

**1\) Batching.** The PR introduces a new Zustand middleware and three new store API:

- `storeApi.batchUpdates()`
- `storeApi.unsafe_startBatching()`
- `storeApi.unsafe_stopBatching()`

It then applies these APIs to batch initial store updates (1e31fbdebb7c12f345d0f9f83b1cac4aa54ba2e7 and 3066b646b711b537da000db689e8993845d4d5d1), plus prevents a few additional store updates (4deee6ffe36e8f3ed7876ec468d223874cccfa7f). This improves initial rendering costs by ~10%, compared to `next`.

10% ain’t much, but:

**2\) CSS-based updates (POC).** This PR also gets rid of an extra `NodeComponent` update, replacing it with a single update of styles (ee555b9ed0064a7a9597ec172c8d26042338895d). This saves another ~10% of rendering costs.

This change is a Proof of Concept since I had time to prototype it but not to finalize it.

### Background

If you record a Chrome DevTools trace of how the Stress example loads, you’ll see this:

CleanShot 2023-11-28 at 01 00 36@2x

There’s a lot of stuff happening here, so let’s highlight the most important pieces:

CleanShot 2023-11-28 at 01 00 50@2x

> _How do I know how it’s exactly these things happening there?_ I went over the recording and looked for some key names (“Evaluate Script” is bundle initialization, `renderRootSync` is a React render, `flushPassiveEffects` is `useEffect`s running). Then, I recorded a trace in React Profiler and matched the rough durations of each render with what I’m seeing in DevTools

Let’s walk over each of these one by one:

1. _Bundle init._ This is when your bundle `<script>` loads – and starts executing all modules defined in it, in the order of imports. This is fairly cheap, plus it’s app-dependent, so it’s not worth optimizing it. 
2. _First render._ This is when React has to render all components in the tree. Since this is the first render, we can’t optimize much there – _every_ component has to render.
3. _First render’s `useEffect`s_. This is when all `useEffect`s defined in all React Flow components execute. This is fairly expensive, so this PR addresses this. We’ll talk about this below, in the “Batching” section.
4. _Second render._ This render is a cheaper one, so this PR skips over it.
5. _Third render._ This render is also expensive; this PR addresses it as well. We’ll also talk about this below, in the “CSS-Based Updates” section.
6. _Subsequent renders._ These renders are fairly cheap and happen _after_ the grid is rendered, so this PR skips over them.

### Batching

If you zoom into the area that runs first render’s `useEffect`s [remember, it’s the area below React’s internal `flushPassiveEffects` method], you’ll notice an interesting detail: a lot (most?) of the time in that area is spent calling Zustand’s `setState`:

CleanShot 2023-11-28 at 01 15 44@2x

Debugging this (eg by adding [the logger middleware](https://github.com/pmndrs/zustand?tab=readme-ov-file#middleware) and trying to log where these `setState`s are coming from) will reveal a curious finding:

- React Flow has a lot of separate components with separate `useEffect`s that call `setState`.

  (Examples: [`StoreUpdater`](https://github.com/xyflow/xyflow/blob/8fe8f73ed3d7fc2a14b2ecbefffb1b988b2edaf8/packages/react/src/components/StoreUpdater/index.tsx#L165-L223) (a few dozen `useEffect`s!), [`ZoomPane`](https://github.com/xyflow/xyflow/blob/8fe8f73ed3d7fc2a14b2ecbefffb1b988b2edaf8/packages/react/src/container/ZoomPane/index.tsx#L94-L98), half a dozen of others)

- When all these components mount, their `useEffect`s run for the first time
- This triggers a lot of independent consecutive `useEffect` calls, each updating the store field by field

Normally, a lot of `useEffect`s aren’t an issue – as long as they’re cheap. However, in our case, calling `setState` isn’t cheap. Every `setState` call costs ~3 ms on my M2 Pro MacBook Pro (20% of the frame budget!) – with no CPU throttling. This is because:
- The Stress example has _several thousands_ of `useStore` selectors! Whenever the store updates, Zustand has to re-run _every_ selector, which (with several thousands of them) gets expensive
- Additionally, some of the selectors use shallow comparison, which means they have to do a loop over their return value every time. This increases the selector costs even further

**Optimizing.** Optimizing this is tricky. You can try to reduce the number of selectors. You can try to find the most expensive selectors and optimize them. Or – you can reduce the number of `setState` calls.

This PR does the latter. It implements a middleware that allows to batch multiple `set()` and `setState()` calls into one. And it enables batching for two critical parts of the app:

- `StoreUpdater`: this component, today, has a separate `useEffect` for every prop that changes. If a user changes three props, today, we’ll run three `useEffect`s and call `setState` three times. When the app initializes, `StoreUpdater` runs ~50 `useEffect`s and calls `setState` 50 times. Thus gets expensive.

   This PR replaces all these `useEffect`s with one – and then batches all updates that happen inside it. 1e31fbdebb7c12f345d0f9f83b1cac4aa54ba2e7

- App init: when the app initializes, apart from `StoreUpdater`, half-a-dozen of other components also call `setState()` inside their `useEffect()`s. To batch these updates, we also enable batching when the app renders – and disable it when the app finishes initializing. 3066b646b711b537da000db689e8993845d4d5d1

> _“Wait, what the heck is batching?”_ “Batching” means “take several independent operations and apply them at once, as a single operation”. In our case, we take several independent `setState` calls and apply them as one `setState()`.
> 
> Batching is similar to [React 18 render batching](https://github.com/reactwg/react-18/discussions/21) or SQL transactions.

#### Batching API design

Batching is implemented through a middleware that extends the store with three new methods:

```js
// METHOD 1: batchUpdates – batches all `setState()` calls that happen inside it
storeApi.batchUpdates(() => {
  store.setState({ a: 5 }); // → no store subscribers are called
  store.setState({ a: 6 }); // → still, no store subscribers are called
  store.getState() // → returns { a: 6 } as expected –
    // this state isn‘t committed to the store yet, but we keep track of it
  
  store.setNodes(...) // → this is also batched, no store subscribers are called
  someMethodThatUpdatesTheStoreInside() // → this also workds
});
// → Finally, the state is committed to the store, and all `useSelector()`s re-run

// METHODS 2 and 3: `unsafe_startBatching()` and `unsafe_stopBatching()` –
// work like `batchUpdates()` but can be called from separate parts of the code
storeApi.unsafe_startBatching();
setTimeout(() => storeApi.unsafe_stopBatching(), 10);
// → all `setState()` calls within the next 10 ms are going to be batched
```

The batching API and implementation are inspired by [React’s own `unstable_batchUpdates()` (aka Redux’s `batch()`)](https://react-redux.js.org/api/batch). 

### CSS-Based Updates

If you profile the third render using React Profiler, you‘ll notice that 1) it’s expensive; 2) a relatively big chunk of its cost is spent re-rendering all `NodeWrapper` elements, 3) all `NodeWrapper`s rerender because their `initialized` prop changes.

CleanShot 2023-11-28 at 01 59 51@2x

The `initialized` prop controls node visibility: when it flips to `true`, nodes go from `visibility: hidden` to `visibility: visible`. And it flips to `true` for all nodes at once – when the nodes get their dimensions measured for the first time.

Question: why don’t we set `visibility: visible` using CSS instead – and save the 100 ms we spend on rerendering all nodes?

This is what commit ee555b9ed0064a7a9597ec172c8d26042338895d does.

This commit is a Proof of Concept – it works, but the implementation is far from optimal, plus it doesn’t account for edge cases (can there be cases when not all nodes will become visible at once?). But it works, it‘s really simple, and it saves 10% of rendering costs.

### Performance Wins

In total, changes in this PR seem to cut around 20% of render-blocking costs (~1170 ms → 990 ms):

**Before ([trace link](https://trace.cafe/t/WITRynbK34))**

CleanShot 2023-11-28 at 00 57 09@2x

**After ([trace link](https://trace.cafe/t/YJ1T4Eoljw))**

Screenshot 2023-11-28 at 00 53 09

These savings will be offset a little by https://github.com/xyflow/xyflow/pull/3668#issuecomment-1826986591, but I still expect a net improvement.

### DIFF
--- a/packages/react/src/components/Edges/utils.ts
+++ b/packages/react/src/components/Edges/utils.ts
@@ -1,11 +1,9 @@
 import type { MouseEvent as ReactMouseEvent } from 'react';
-import type { StoreApi } from 'zustand';
-
-import type { Edge, ReactFlowState } from '../../types';
+import type { Edge, ReactFlowStoreApi } from '../../types';
 
 export function getMouseHandler(
   id: string,
-  getState: StoreApi<ReactFlowState>['getState'],
+  getState: ReactFlowStoreApi['getState'],
   handler?: (event: ReactMouseEvent<SVGGElement, MouseEvent>, edge: Edge) => void
 ) {
   return handler === undefined
--- a/packages/react/src/components/Nodes/utils.ts
+++ b/packages/react/src/components/Nodes/utils.ts
@@ -1,12 +1,10 @@
 import type { MouseEvent, RefObject } from 'react';
-import type { StoreApi } from 'zustand';
-
-import type { Node, ReactFlowState } from '../../types';
+import type { Node, ReactFlowStoreApi } from '../../types';
 import { errorMessages } from '@xyflow/system';
 
 export function getMouseHandler(
   id: string,
-  getState: StoreApi<ReactFlowState>['getState'],
+  getState: ReactFlowStoreApi['getState'],
   handler?: (event: MouseEvent, node: Node) => void
 ) {
   return handler === undefined
@@ -29,8 +27,8 @@ export function handleNodeClick({
 }: {
   id: string;
   store: {
-    getState: StoreApi<ReactFlowState>['getState'];
-    setState: StoreApi<ReactFlowState>['setState'];
+    getState: ReactFlowStoreApi['getState'];
+    setState: ReactFlowStoreApi['setState'];
   };
   unselect?: boolean;
   nodeRef?: RefObject<HTMLDivElement>;
--- a/packages/react/src/components/ReactFlowProvider/index.tsx
+++ b/packages/react/src/components/ReactFlowProvider/index.tsx
@@ -1,4 +1,4 @@
-import { useRef, type ReactNode } from 'react';
+import { useRef, type ReactNode, useEffect } from 'react';
 import { UseBoundStoreWithEqualityFn } from 'zustand/traditional';
 
 import { Provider } from '../../contexts/RFStoreContext';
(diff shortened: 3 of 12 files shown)


moklick:
As discussed, we only took a small part of this PR. The idea how we could re-write the `StoreUpdater` component was great and helped to make it easier to understand. Maintaining the `batchUpdate` middleware seems to costly for to little improvement here but it's a very good idea. Maybe we can add it in the future.  

closed by #3729

PR #3685, opened 2023-11-30
[Perf] Add performance tests for selecting, dragging, and remounting

iamakulov:
This PR adds additional performance scenarios to the stress test:

CleanShot 2023-11-30 at 03 21 41@2x

and introduces a system for recording and [visualizing](https://observablehq.com/@iamakulov/long-frame-visualizer) them:

Screenshot 2023-11-30 at 03 21 26

### More details

This PR adds four new buttons into the stress test, designed to test the most critical interactions:

- “select node”
- “drag within viewport”
- “drag outside of viewport”
- ”re-mount”

#### First three buttons

Unlike the existing buttons (“change pos” and etc.) which just change the React Flow state, “Select node”, ”Drag within the viewport” and “Drag outside of the viewport” actually emulate user interactions – by creating and dispatching fake mouse events:

```js
const node = document.querySelector(...);

node.dispatchEvent(new MouseEvent("mousedown", ...));
node.dispatchEvent(new MouseEvent("mousemove", ...));
node.dispatchEvent(new MouseEvent("mousemove", ...));
node.dispatchEvent(new MouseEvent("mousemove", ...));
node.dispatchEvent(new MouseEvent("mouseup", ...));
```

This achieves two goals:
- The tests are kept as close to real-world interactions as possible. Selecting a node with a mouse might [(and actually does)](https://github.com/xyflow/xyflow/pull/3667) use a different code path compared to just setting `selected: true`. Dispatching actual mouse events ensures we use the same code path a real user uses
- The tests are stable and reproducible. You can click “Drag within the viewport” in two different branches, and the node will be dragged absolutely identically in each of them – making the test results easier to compare

#### Fourth button

The “Re-mount” button doesn’t dispatch any user interactions – but, instead, re-mounts the `<ReactFlow>` element. This re-initializes React Flow from scratch and follows the same code path that happens when React Flow mounts for the first time.

This achieves two goals:
- _convenience:_ you don’t need to reload the page every time to measure React Flow remounting
- _noise reduction:_ because you don’t need to reload the page every time, perf measurements don’t get the noise from network events anymore

#### Output

Each of the new four tests tracks the duration of frames that happen during the test – and outputs them to the console:

Screenshot 2023-11-30 at 03 38 03

You can inspect these durations manually – or copy-paste them into https://observablehq.com/@iamakulov/long-frame-visualizer, which will do some basic visualizations for that data and allow you to compare across branches.

### DIFF
--- a/examples/react/src/examples/Stress/index.tsx
+++ b/examples/react/src/examples/Stress/index.tsx
@@ -1,4 +1,4 @@
-import { useState, useCallback } from 'react';
+import { useState, useCallback, useRef, useEffect } from 'react';
 import {
   ReactFlow,
   Edge,
@@ -16,6 +16,7 @@ import {
 } from '@xyflow/react';
 
 import { getNodesAndEdges } from './utils';
+import { FrameRecorder, generateMouseEventParamsTargetingNode, nextFrame } from './performanceUtils';
 
 const { nodes: initialNodes, edges: initialEdges } = getNodesAndEdges(25, 25);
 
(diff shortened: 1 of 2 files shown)


moklick:
This is super helpful 🙏

PR #3718, opened 2023-12-13
fix(edges): add suport for className in baseEdge and remove unecessary render

bearkfear:
This PR add support for className in baseEdge, for those who wants to use tailwindcss and other css processors. In the actual scenario, it's necessary to create a **path** and send the edge Path to them to style the edge. It's possible to style with the **style property** too. But, not CSS. 

Other changes: 

Related to the EdgeLabelText.

The actual implementation of the EdgeLabelText is using an approach that observes the label changes, it's more "tracking perfect" to get the changes based on the reference itself, removing a re-render caused  by the useEffect.


moklick:
That's a nice refactoring idea for the `EdgeText` component. Thanks!

PR #3723, opened 2023-12-14
Added seperate portal to put dom elements inside the flow

peterkogo:
Unsure about naming. 
Code of EdgeLabelRenderer is way to short to sensibly abstract so there is a little code duplication happening

### DIFF
--- a/packages/react/src/components/ViewPortal/index.tsx
+++ b/packages/react/src/components/ViewPortal/index.tsx
@@ -1,0 +1,19 @@
+import type { ReactNode } from 'react';
+import { createPortal } from 'react-dom';
+
+import { useStore } from '../../hooks/useStore';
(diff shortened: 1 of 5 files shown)


moklick:
We need similar default styles like we have for the edge label renderer. Could you add a new class and use the styles for edge label renderer and viewport portal?

peterkogo:
Forgot to add `/system` to the commit... 
I created a new class for the viewport portal, because it slightly deviates from the edgelabel renderer by:
```css
pointer-events: none;
user-select: none;
```

Plus, this way the classes also stay clean and descriptive in the in the DOM like so:

Screenshot 2023-12-18 at 12 00 20

moklick:
perfect, thanks!

PR #3776, opened 2024-01-09
Update component-props.ts

lumosmind:
onEdgeClick function has "node" parameter but type of the parameter is "Edge". It is more convenient to rename the parameter as "edge".


moklick:
good find :) thanks!

PR #3791, opened 2024-01-13
fix: typesafety of Node and Edge changes

FelipeEmos:
This PR is my take on a discussion regarding the same problem that can be found here:
- https://github.com/xyflow/xyflow/pull/2923
- https://github.com/xyflow/xyflow/issues/3030

I believe I improved the Lib's Typesafety without adding any breaking changes to it...
Let me know what you think 🙂 🚀 

Cheers! Thanks for the amazing lib 🙏🏻 

### Description
On some functions and type definitions related to Change Events,
the generics for the types Node and Edge were being lost.

This commit fixes those:
  - Functions
    - applyNodeChanges
    - applyEdgeChanges
  - Types
    - NodeChange
    - EdgeChange
    - OnNodesChange
    - OnEdgesChange

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2019-2023 webkid GmbH
+Copyright (c) 2019-2024 webkid GmbH
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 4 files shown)


moklick:
good catch, thanks @FelipeEmos

PR #3793, opened 2024-01-15
refactor(nodes): use initialWidth/initialHeight instead of width/height

moklick:
An issue with v12 is that flows behave differently if you pass in v11 nodes. In v12 `width`/ `height` are used an inline styles and you lose the auto default `width` / `height` behaviour. Once you set `width`/ `height`, you always need to update those values if your node dimensions change. The idea of this PR is to remove `width`/ `height` from the node attributes and add new attributes `initialWidth`/ `initialHeight`. Those values are only used if a node has no computed dimensions (which is the case on the server). Unfortunately this introduces a new issue: built-in nodes have a different size on the server and on the client if you only use `initialWidth`/ `initialHeight`, because the default styles do `.react-flow__node-input{ ... width: 150px}`. I am not sure how we could solve this with this new approach.

### DIFF
--- a/examples/astro-xyflow/src/components/ReactFlowExample/index.tsx
+++ b/examples/astro-xyflow/src/components/ReactFlowExample/index.tsx
@@ -17,8 +17,8 @@ import CustomNode from './CustomNode';
 import '@xyflow/react/dist/style.css';
 
 const nodeSize = {
-  width: 100,
-  height: 40,
+  initialWidth: 100,
+  initialHeight: 40,
 };
 
 const initialNodes: Node[] = [
@@ -32,8 +32,8 @@ const initialNodes: Node[] = [
       {
         type: 'source',
         position: Position.Bottom,
-        x: nodeSize.width * 0.5,
-        y: nodeSize.height,
+        x: nodeSize.initialWidth * 0.5,
+        y: nodeSize.initialHeight,
       },
     ],
   },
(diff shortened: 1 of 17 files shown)


moklick:
An issue with v12 is that flows behave differently if you pass in v11 nodes. In v12 `width`/ `height` are used an inline styles
and you lose the auto default `width` / `height` behaviour. Once you set `width`/ `height`, you always need to update those values if your node dimensions change. The idea of this PR is to remove `width`/ `height` from the node attributes and add new attributes `initialWidth`/ `initialHeight`. Those values are only used if a node has no computed dimensions (which is the case on the server). Unfortunately this introduces a new issue: default nodes have a different size on the server and on the client if you only use `initialWidth`/ `initialHeight`, because the default styles do `.react-flow__node-input{ ... width: 150px}`. I am not sure how we could solve this with this new approach.

calebpeterson:
I appreciate the clarity that would come from the `initial*` prefix on the names.

moklick:
we could also support `initialWidth` / `initialHeight` and `width`/ `height` 🤯

moklick:
we decided to provide `width`/ `height` (static dimensions for nodes on the server and on the client) and `initialWidth` / `initialHeight` (initial dimensions that can be used for ssr, dynamic dimensions on the client).

PR #3803, opened 2024-01-16
Add Potentially Missed Return in useReactFlow getNodeRect

rhys-okane-instil:
Was working through an issue I was having and found this array/tuple sitting alone within an if statement. I imagine a return was accidentally missed here, so I thought I would add it.

Please let me know if this is the wrong branch or wrong format :)


moklick:
what the 😳 thanks! I will close this when the fix is published.

moklick:
fixed in 11.10.2.

PR #3811, opened 2024-01-18
Added onInit and useInitialized hooks

peterkogo:


packages/svelte/src/lib/hooks/useInitialized.ts
moklick, line 1: wouldn't it make more sense to return readables here?

packages/svelte/src/lib/index.ts
```diff
@@ -31,6 +31,7 @@ export * from '$lib/hooks/useConnection';
 export * from '$lib/hooks/useNodesEdges';
 export * from '$lib/hooks/useHandleConnections';
 export * from '$lib/hooks/useNodesData';
+export * from '$lib/hooks/useInitialized';
```
moklick, line 34: we need to export `useNodesInitialized` / `useEdgesInitialized` (do we want to offer that?) too

moklick (changes requested):

packages/svelte/src/lib/index.ts
peterkogo: Oh wait `useInitialized` includes `useNodesInitialized` `useEdgesInitialized`. It is already exported

moklick (changes requested):

PR #3815, opened 2024-01-18
add getNode, getNodes, getEdge, getEdges

peterkogo:
I think it is better to keep everything simple: not support `getEdges(id)` just `getEdges([id])`


packages/svelte/src/lib/hooks/useSvelteFlow.ts
```diff
@@ -121,6 +127,10 @@ export function useSvelteFlow(): {
   return {
     zoomIn,
     zoomOut,
+    getNode: (id) => get(nodeLookup).get(id),
+    getNodes: (ids) => ids.map((id) => get(nodeLookup).get(id)),
```
moklick, line 131: I would expect to be able to call `getNodes` without passing any ids to get all nodes here

moklick (changes requested):

PR #3821, opened 2024-01-22
fitView improvements

peterkogo:
## Changes
- added waitForInit to FitViewProps 
- added fitViewScheduled to store
- removed fitViewOnInit, fitViewOnInitOptions, fitViewDone from store

## Questions
- [x] What implications are there for reset()?
- [ ] Is returning true if fitView has not been called yet and just scheduled okay?

### DIFF
--- a/packages/react/src/components/StoreUpdater/index.tsx
+++ b/packages/react/src/components/StoreUpdater/index.tsx
@@ -136,8 +136,8 @@ export function StoreUpdater(props: StoreUpdaterProps) {
         else if (fieldName === 'translateExtent') setTranslateExtent(fieldValue as CoordinateExtent);
         else if (fieldName === 'nodeExtent') setNodeExtent(fieldValue as CoordinateExtent);
         // Renamed fields
-        else if (fieldName === 'fitView') store.setState({ fitViewOnInit: fieldValue as boolean });
-        else if (fieldName === 'fitViewOptions') store.setState({ fitViewOnInitOptions: fieldValue as FitViewOptions });
+        else if (fieldName === 'fitView') store.setState({ fitViewScheduled: fieldValue as boolean });
+        else if (fieldName === 'fitViewOptions') store.setState({ fitViewScheduled: fieldValue as FitViewOptions });
         // General case
         else store.setState({ [fieldName]: fieldValue });
       }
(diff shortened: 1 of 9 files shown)


ogroppo:
Hi, is this PR aimed to render the nodes already fitted instead of seeing the transition on page load?

moklick:
@ogroppo  This PR is about making it possible to call `fitView` right after setting new nodes. 

If you use the `ReactFlow` `fitView` prop, it should fit before the first render.

ogroppo:
Apologies if I am polluting this conversation with this issue, so redirect me in case, I am seeing this behaviour with minimal setup using version 12.0.0-next.17

https://github.com/xyflow/xyflow/assets/4820803/87392f47-93cc-413c-9425-8b157e76d7f5

```
  const [nodes, setNodes, onNodesChange] = useNodesState(initialNodes);
  ...
  <ReactFlow
          nodes={nodes}
          edges={edges}
          onNodesChange={onNodesChange}
          onEdgesChange={onEdgesChange}
          onConnect={onConnect}
          fitView
          colorMode={colorMode}
        >
```

When removing fitView the zoom does not adjust but the edges are rendered later 

https://github.com/xyflow/xyflow/assets/4820803/b622933e-f9b2-4d0f-913b-92b27a88bbad

moklick:
How do your nodes look like?

ogroppo:
```

**Comment by moklick** [comment]
I really like the idea but it would be nice if we could find a more generic solution like:

```js
setNodes([...])

await nodeRenderTick();

fitView();
```

Let's open a new PR for a generic solution.

@ogroppo The reason for this behaviour is that you are initializing your nodes with dimensions. If you remove width and height, there won't be any flickering. It's only necessary for server side rendering.

dagadbm:
so why was this closed? the issue still  exists

moklick:
@dagadbm Which issue still exists?

dagadbm:
fitview generally being unreliable.

i had to do a set timeout to make sure it works

moklick:
Which version are you using?

dagadbm:
"@xyflow/react": "^12.4.4",

moklick:
Please update to the latest version 12.8.6.

PR #3824, opened 2024-01-23
child nodes get repositioned correctly when resizing top or left

peterkogo:


peterkogo:
- Child nodes get positioned correctly when resizing parent (top/left)
- extent="parent" gets respected during resize of children
- extent="parent" & expandParent gets respected during resize of parent
- resizing (top/left) of a child with expandParent does not cause moving away of node
- all of the above works for keepAspectRatio

moklick:
🔥🔥🔥 super nice update. Thanks @peterkogo

PR #3834, opened 2024-01-26
React Flow: TS node types

moklick:
In v12 we want to provide a clear way how to define custom node types (the TS types, not the React Flow `nodeTypes`) and how to use them. 

### Issues with v11

In v11 there are some functions that take a generic to pass the nodes data object. This isn't helpful, because you loose the `node.type` information, which is interesting especially because TS knows about the related data object. This was the implementation for `applyNodeChanges`

```js
function applyNodeChanges<NodeData = any>(
  changes: NodeChange[],
  nodes: Node<NodeData>[]
): Node<NodeData>[] { ... }
```

As you can see, we loose all the information, because `Node` is being used and only `NodeData` gets inferred or passed. 

### Possible solution for v12

In v12, users define a union of your node types and then use that type to specify the node type of imported function:

```js
type TextNode = Node<{ value: string }, 'text'>;
type NumberNode = Node<{ value: number }, 'number'>;
type ResultNode = Node<{}, 'result'>;

type MyNode = TextNode | NumberNode | ResultNode;

const nodes: MyNode[] = [{ ... }]
```

now the `MyNode` type can be used to specify the type for `useReactFlow` for example:

```js
const reactFlowInstance = useReactFlow<MyNode>();
```

Moreover we are using the passed `nodes` to infer the type for the functions that you pass to the `<ReactFlow />` component, like `onNodeDrag` or `onNodesChange`. The `applyNodesChanges` now looks like this:

```js
function applyNodeChanges<NodeType extends Node = Node>(
  changes: NodeChange<NodeType>[],
  nodes: NodeType[]
): NodeType[] { ... }
```

#### The `Node` type

Currently we can pass the data structure and the type like this:

```js
type TextNode = Node<{ value: string }, 'text-node'>;
``` 

The underlying Node type looks like this:

```js
type Node<Data = any, Type extends string | undefined = string | undefined> = {
  id: string;
  data: Data;
  type?: Type;
...
```

I am wondering if we could add some more constraints to the `NodeData` generic. Should we only allow object? Or is it fine, to do pass a primitive as data `Node<5, 'five-node'>`? 

A possible solution: 

```js
type Node<Data extends Record<string, unknown> = Record<string, unknown>, Type extends string = string> {
  id: string;
  data: Data;
  type: Type;
}
```

#### `useStore`

For now I wasn't able to type the `useStore` hook correctly and we need some ugly casting workarounds, to get the ~correct node type:

```js 
const node = useStore(s => s.nodeLookup.get(id) as NodeType).
```

Since `useStore` doesn't know about the NodeType, we need to cast the returned node, because it has the type `Node`. It wouldn't be too bad if this was only being used on library side, but we are also exporting it.

I am wondering if there could be any downsides with this new approach or if there might be better solutions for handling node types.


moklick:
@gavan1 thanks for the hint! The docs are a bit misleading here. You would need to add your custom nodes if you have some. I will add a comment.

gavan1:
@moklick I did upate my comment.
You and your team have built something truly important and unique, don't let outdated docs stiffle the DX. I'de glady help if you guys need it.

moklick:
@gavan1 I just removed that lines.. won't help in that example

gavan1:
I'de highly recommend to follow Information Architecure rules, Great products suffer from what I've coined DD., Document Diarrhea.

Following the 9 principles of IA will go a long way, espesially the Principle Of Front Doors.

PR #3842, opened 2024-01-30
type: remove repeated type

li-jia-nan:
type: remove repeated type


moklick:
good catch! thanks @li-jia-nan :)

PR #3850, opened 2024-01-31
Svelte Flow: Use provided default node before falling back to built in default

mr-josh:
Try and fallback to a provided default node from the nodeTypes store and if that returns a falsey value, fallback to the builtin DefaultNode.

### Example:

If you're unsure about the available nodeTypes available at runtime and want to provide a more clear fallback, svelteflow won't fallback to the provided default in nodeTypes and instead fallback to the svelteflow provided default.

```svelte
<script>
  // ...

  const nodeTypes = {
    default: FallbackNode,
  };
</script>

<SvelteFlow {nodeTypes}>
  <!-- ... -->
</SvelteFlow>
```


moklick:
Thanks for your help but this is not what we want to do here. `default` is just a name for a node type and has no special functionality here.

PR #3851, opened 2024-01-31
Adding id property to Handle component result

ricardosouza-logmein:


moklick:
Hey @ricardosouza-logmein ! Thanks for your help but this won't work. We are using the `id` prop as a handle id (which is necessary if you are working with multiple handles). There is no way for us to distinguish between a user who wants to pass a normal `id` attribute or an actual handle id. If you need it, you can wrap your Handle components with a div and add an `id` there:


```jsx
<div id="some-id">
  <Handle />
</div>
```

PR #3853, opened 2024-02-01
fix Couldn't create edge for ${handleType} handle id: 

stallboy:
```javascript
setNodes(newNodes);
setEdges(newEdges);
```

when handles in newNode has changed, but id not change, and the newEdge use the new handle,  the error  `Couldn't create edge for ${handleType} handle id: ` will show in console and the edge will not show.

so i change code at two places.
1. to not use the old stale handle cache . 
2. to make the resize observe the new node in order to make the updateNodeDimensions happen. then trigger the new edge  show.

### DIFF
--- a/packages/react/src/components/NodeWrapper/index.tsx
+++ b/packages/react/src/components/NodeWrapper/index.tsx
@@ -88,7 +88,7 @@ export function NodeWrapper({
 
       return () => resizeObserver?.unobserve(currNode);
     }
-  }, [node.hidden]);
+  }, [node]);
 
   useEffect(() => {
     // when the user programmatically changes the source or handle position, we re-initialize the node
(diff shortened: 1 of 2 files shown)


stallboy:
```ts
  useEffect(() => {
    if (nodeRef.current && !node.hidden) {
      const currNode = nodeRef.current;
      resizeObserver?.observe(currNode);

      return () => resizeObserver?.unobserve(currNode);
    }
  }, [node]);
```

the `useEffect` to listen on `[id, nodeType, node.sourcePosition, node.targetPosition]` code after this can be deleted imho.

moklick:
Thanks for the report. Could you open an issue with a codesandbox or repo for this? Your fix doesn't work, because it introduces other problems like calling `resizeObserver?.observe` whenever the node object changes.

stallboy:
Thanks for the response. the bug exists on 12.0.0-next.8. after a try. I can't reproduce the bug on 12.0.0-next.9. That's great. Also thanks for the great library!

PR #3897, opened 2024-02-12
fix(nodes): re-measure nodes when dimensions are missing 

### LINKED ISSUE #3886: Subflow parent nodes doesn't update size when nodes are re-initialized
If you specify a size for a parent node it should always be overwritten when one of the child nodes uses the `expandParent` option. As you can see in the example, the size gets overwritten initially but not when you reset the nodes (with the same set of nodes).

### Example 

https://codesandbox.io/p/sandbox/subflow-bug-dlqsnr?file=/src/Flow.tsx

1. open codesandbox link
2. click "reset nodes"
3. parent node has a different size

### Expected behavior

parent node should keep the adjusted size.

### Screenshots or Videos

_No response_

### Platform

- OS: macOs 
- Browser: Chrome
- Version: 11.10.3

### Additional context

Initially we measure all nodes and send the updates to the `applyNodeChanges` function. There we adjust the size of the parent node and everything looks good. If you then use the same nodes to re-initialize the flow, the child node is already there and we won't measure it again. That's why there is no change that could lead to a resizing of the parent node.

moklick:


mcupito:
Hi @moklick  - any update on this?

moklick:
Hey @mcupito 

this is still WIP.

moklick:
@abi-spyro, @mcupito is this issue solved for you?

Sam-Apostel:
Hi @moklick 

I updated my project from v11.10.3 to v11.10.4 and I got an infinite loop of `NodeDimensionChange` triggers. This causes my nodes to toggle between `visibility: visible` and `visibility: hidden`. Which causes the nodes not to be selectable/draggable.

Any ideas on where this could've gone wrong?

I am using external state management (state is stored using react-hook-form).

Sam-Apostel:
@moklick using applyNodeChanges instead of my custom resolver fixed my issues. Thanks for pointing me towards that function. 

Would it be helpful if we added a link to the applyChanges functions on https://reactflow.dev/api-reference/hooks/use-nodes-state ?

PR #3907, opened 2024-02-13
Refactor(setNodes): improve current batching solution

moklick:


moklick:
🔥🔥🔥 thanks for your help here @hayleigh-dot-dev . Looks really good!

Issue #3927, opened 2024-02-19
`useNodesData`: return `type` and `id` besides data object


moklick:
Currently we only return the actual data object. It would be helpful to return the node id and the type, too.

bogris:
I don't think it's bad, as I have this implemented in my custom hooks, 

but if we go this way, I think you overlap a lot with getNode from useReactFlow.

moklick:
This was released in Svelte Flow 0.0.37 and React Flow 12.0.0-next.10


@bogris  `useNodes` returns all nodes. This means that the component that uses it, re-renders multiple times when a node is being dragged. This isn't the case for `useNodesData` and additionally you can pass an array of ids of the nodes data you want to receive.

PR #3929, opened 2024-02-20
Extend useNodesData by {id, type, data} 

peterkogo:


peterkogo:
closes #3927

peterkogo:
We have to double check the typing in Svelte, it does not seem to work yet 100%

Issue #3836, opened 2024-01-27
cursor: grab set on pane when panOnDrag is false


jagregory:
### Describe the Bug

In the CSS for the pane the cursor is always set to `grab`, even if you've set `panOnDrag` to `false`.

See: `xy-flow__pane`/`react-flow__pane` https://github.com/xyflow/xyflow/blob/f0117939bae934447fa7f232081f937169ee23b5/packages/system/src/styles/init.css#L63-L74

Given panning is actually disabled when `panOnDrag` is `false`, it would make sense to not show the cursor as indicating dragging is possible.

### Steps to Reproduce the Bug or Issue

1. set `panOnDrag={false}`
2. mouse over the component and see the cursor is `grab`/the little hand

### Expected behavior

When `panOnDrag={false}` set the cursor to `default` OR provide a way we can pass our own styles down into the Pane component so we can do this ourself.

### Platform

All

moklick:
Hey @jagregory, I agree that this would be the better default. I put it on the list for v12.

moklick:
published in v12.0.0-next.11

c0d3ster:
related but it feels like cursor should also be pointer whenever 0 is not in the `panOnDrag` array since the cursor is usually tied to your left click operation. 
Current behavior is that the cursor is grab if any values are in the `panOnDrag` array from my understanding of it. I'd be happy to make a new issue for this if it feels independent from this one, just lmk @moklick

PR #3933, opened 2024-02-21
Draggable cursor fix

peterkogo:
We need to check if the sensible default for nodes is cursor `pointer` or `default`


peterkogo:
closes #3836

moklick:
these defaults are way better 👌

PR #3934, opened 2024-02-21
disableKeyboardA11y now correctly prevents escape key presses

### LINKED ISSUE #3899: Abillity to disable/override Escape key functionality

**Request**

I would like to share use of the 'Escape' keyboard press with other parts of my web page.

** Current Issue **

Currently the Escape key functionality in flow cannot be disabled.  If the Escape key is used in other parts of the webpage then it is difficult to know whether flow acted on an Escape key press.  We'd rather only act on key presses that did not have an effect on the flow (stacking the event).

Detecting if flow did anything with the event:
- Flow does not stop propagation of the event (even if it was acted upon)
- One way to detect if the Escape key did anything is using an useEvent and setTimeout 0, checking whether the activeElement has changed
- Snatching the event in the capture event does not fully solve the issue as we would like the retain the flow 'Escape' behaviour too

** Possible Suggestions ** 

- exposing an optional 'Escape' keyboard handler
- a flag to disable 'Escape' and maybe 'Enter' handling

peterkogo:
closes #3899

### DIFF
--- a/packages/react/src/components/EdgeWrapper/index.tsx
+++ b/packages/react/src/components/EdgeWrapper/index.tsx
@@ -34,6 +34,7 @@ export function EdgeWrapper<EdgeType extends Edge = Edge>({
   edgeTypes,
   noPanClassName,
   onError,
+  disableKeyboardA11y,
 }: EdgeWrapperProps<EdgeType>): JSX.Element | null {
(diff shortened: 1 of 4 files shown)


moklick:
thanks 👌

Issue #3975, opened 2024-03-04
useNodesData: fix equality check for return value


moklick:
Currently `useNodesData` always returns a new array when you pass an array of ids.

moklick:
fix is released in the latest version.

PR #3976, opened 2024-03-04
improved selector equality function for useNodesData

peterkogo:


packages/react/src/hooks/useNodesData.ts
```diff
@@ -4,10 +4,23 @@ import { shallow } from 'zustand/shallow';
...
+function nodeDatasEqual(a: NodeData | NodeData[], b: NodeData | NodeData[]) {
+  const _a = Array.isArray(a) ? a : [a];
+  const _b = Array.isArray(b) ? b : [b];
+
+  if (_a.length !== _b.length) {
+    return false;
+  }
+
+  for (let i = 0; i < _a.length; i++) {
+    if (!(_a[i].id === _b[i].id && _a[i].type === _b[i].type && shallow(_a[i].data, _b[i].data))) {
```
moklick, line 18: do we really want to use `shallow` here? I think `Object.is` is enough as an equality check, because I would expect that something has changed, if user created a new object for `data`

moklick (changes requested):

peterkogo:
closes #3975

PR #3988, opened 2024-03-06
Update handle lookup onPointerMove

### LINKED ISSUE #3983: useUpdateNodeInternals doesn't update handle positions during active connection
### Describe the Bug

I'm using custom nodes in my graph component and these nodes collapse the handles when user is not near them. This is done with absolute position on handles and with animations. Now, I've managed to handle animation and position updates on the source node by calling `updateNodeInternals` on every `requestAnimationFrame` until animation is done.

However, if the user has initiated a connection then target nodes doesn't update their handle positions. I think internally they are being updated but no one told the connection about it :)

https://codesandbox.io/p/devbox/react-flow-edge-issue-dvgwvh

This is a very crude example, especially `requestAnimationFrame` being called all the time but it's the same behaviour with optimized version anyway.

1. Try to connect the source to target
2. See that even though you are hovering over target the connection thinks handle position is on the left (as opposed to where you are seeing)

### Expected behavior

I expect that when `updateNodeInternals` is called the connection should also know where the target handles are.

### Screenshots or Videos

_No response_

### Platform

Browser: Chrome (122.0.6261.94)
React Flow version: 11.10.4

### Additional context

_No response_

DenizUgur:
This is an attempt to solve issues laid in #3983. It's obviously not efficient but if there is a way to get notified when node internals are updated then the result can be memoized.

closes #3983

### DIFF
--- a/packages/system/src/xyhandle/XYHandle.ts
+++ b/packages/system/src/xyhandle/XYHandle.ts
@@ -113,13 +113,6 @@ function onPointerDown(
   let isValid = false;
   let handleDomNode: Element | null = null;
 
-  const handleLookup = getHandleLookup({
-    nodes,
-    nodeId,
-    handleId,
-    handleType,
-  });
-
   // when the user is moving the mouse close to the edge of the canvas while connecting we move the canvas
   function autoPan(): void {
     if (!autoPanOnConnect || !containerBounds) {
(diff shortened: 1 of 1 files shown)


moklick:
Hey @DenizUgur 

thanks for the proposed solution. I will have a look in the next days.

DenizUgur:
Hi @moklick, did you had a chance to look at this?

Issue #4008, opened 2024-03-13
React Flow v12: Cancel Connection does not work anymore


peterkogo:

moklick:
fixed in SF 0.0.39 and RF 12-next.12

PR #4010, opened 2024-03-13
Connections can be correctly cancelled

peterkogo:
added check to xyhandle pointermove if connection is cancelled


peterkogo:
closes #4008

PR #4041, opened 2024-03-19
Remove white bar under MiniMap

OrhanTozan:
Before:
image

After:
image

By changing the display from the svg from the default "inline" to "block", the svg and parent div get the same height (which I think is the intended behaviour + less ugly because of the removal of the white bar)


moklick:
Thanks @OrhanTozan . This already part of the v12 base css.

PR #4054, opened 2024-03-21
Proposal on how to handle measured node dimensions and internals 

moklick:
V12 is a few weeks in alpha testing and it seems that a lot of people are still struggling with the measured node dimensions. This PR is an attempt to make it easier to understand how to access those values. 

## v11 

In v11 users receive the measured dimensions in a dimension change event. After applying the changes, the dimensions are stored in `node.width` and `node.height. This caused a lot of confusion, because users assume that they can use those attributes for setting the actual dimensions (which is not the case..) when it's only used to save the measured ones. We are also using those attributes as an indicator if a node is initialized.

1. user passes nodes
2. react flow measures nodes
3. rf sends dimension changes
4. user applies dimension (node.width and node.height is set now) 
5. node is initialized and visible

## v12 alpha

In v12 we wanted to make this more clear, so we decided to add another attribute called `node.computed` and store the measured dimensions there. We also needed to make this change, because wanted to be able to use `node.width` and `node.height` for specifying node dimensions on user side which is necessary for SSR. In my view this solution is better than what we have in v11, but it's still a bit confusing and you need to know that measured dimensions get added after you receive the initial dimension change events.   

1. user passes nodes
2. react flow measures nodes
3. rf sends dimension changes
4. user applies dimension (node.computed.width and node.computed.height is set now) 
5. node is initialized and visible

## a better solution for v12?

With this PR I am proposing another solution for the node dimensions dilemma. With this PR we are not adding the measured dimensions to the user nodes, but only store them internally. This means that users won't receive the initial dimension change event and need to get the values with new helper functions and hooks: 

1. user passes nodes
2. react flow measures nodes
3. rf stores dimensions internally
4. node is initialized and visible

This PR introduces a new `InternalNode` type that combines the previous `computed` and `[internalsSymbol]` values in an `internals` attribute. The existing `Node` type got a cleanup and doesn't include `computed` and `[internalsSymbol]` anymore. This separation between the node types makes it more clear what to expect on library and on user side. User can access the the internals in a custom node with the `useInternalNode` hook:

```js
const internalNode = useInternalNode(id);
const nodeWidth = internalNode.internals.width;
```

or with the `getInternalNode` helper function:

```js
const { getInternalNode } = useReactFlow();

const onLayout = (direction: string) => {
  const dagreGraph = new dagre.graphlib.Graph();
  dagreGraph.setGraph();
  
  nodes.forEach((node) => {
    const internalNode = getInternalNode(node.id);

    dagreGraph.setNode(node.id, {
      width: internalNode?.internals.width ?? 150,
      height: internalNode?.internals.height ?? 50,
    });
  });
 ...
```

## wdyt?

Any feedback is highly appreciated. We are not sure, if we really want to go this path, because it would introduce a new library behaviour that is way more different than the current v12 (and v11) implementation.

### DIFF
--- a/examples/react/src/examples/EasyConnect/FloatingEdge.tsx
+++ b/examples/react/src/examples/EasyConnect/FloatingEdge.tsx
@@ -1,11 +1,10 @@
-import { useCallback } from 'react';
-import { useStore, getStraightPath, EdgeProps } from '@xyflow/react';
+import { getStraightPath, EdgeProps, useInternalNode } from '@xyflow/react';
 
 import { getEdgeParams } from './utils.js';
 
 function FloatingEdge({ id, source, target, markerEnd, style }: EdgeProps) {
-  const sourceNode = useStore(useCallback((store) => store.nodes.find((n) => n.id === source), [source]));
-  const targetNode = useStore(useCallback((store) => store.nodes.find((n) => n.id === target), [target]));
+  const sourceNode = useInternalNode(source);
+  const targetNode = useInternalNode(target);
 
   if (!sourceNode || !targetNode) {
     return null;
--- a/examples/react/src/examples/EasyConnect/utils.tsx
+++ b/examples/react/src/examples/EasyConnect/utils.tsx
@@ -9,8 +9,8 @@ function getNodeIntersection(intersectionNode: Node, targetNode: Node) {
     width: intersectionNodeWidth,
     height: intersectionNodeHeight,
     positionAbsolute: intersectionNodePosition,
-  } = intersectionNode.computed || {};
-  const targetPosition = targetNode.computed?.positionAbsolute!;
+  } = intersectionNode.internals || {};
+  const targetPosition = targetNode.internals?.positionAbsolute!;
 
   const w = intersectionNodeWidth! / 2;
   const h = intersectionNodeHeight! / 2;
@@ -33,7 +33,7 @@ function getNodeIntersection(intersectionNode: Node, targetNode: Node) {
 
 // returns the position (top,right,bottom or right) passed node compared to the intersection point
 function getEdgePosition(node: Node, intersectionPoint: XYPosition) {
-  const n = { ...node.computed?.positionAbsolute, ...node };
+  const n = { ...node.internals?.positionAbsolute, ...node };
   const nx = Math.round(n.x!);
   const ny = Math.round(n.y!);
   const px = Math.round(intersectionPoint.x);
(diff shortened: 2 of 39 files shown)


bogris:
for my use cases, I find it usefull to have the width / height in the nodes array, because I do a lot of layout computations where parents depend on the children. (simmilar to HTML flex behaiviour). 

If the dimensions will not be available in the nodes array, things will get another layer of complexity. But it's more or less negligible. 
I don't know if other libraries will suffer the same issue. 

On the other hand,  the proposed approch seems more straight foreward and more clear. 

If I see less steps in the logic it looks like a win for me. Humans are adaptable :))

+1 for the propsed approach.

rob-gordon:
**(My thoughts - I'm not a React Flow expert!)**

Intuitively speaking, it does seem like computed sizes belong on the internal node because the react flow internals create/update them and the user isn't meant to set them directly. The helpers/hooks make sense to me as well 👍🏻 

My only concern is what if I want to fire off a layout function _as soon as_ the internal node sizes are computed. 
From what I can tell I don't really have a way to do that. I would need to bind to the regular node/edge dimension changes, which, from what I understand aren't exactly triggered to the internal computed size update. I guess that's the part that's fuzzy for me.

OrhanTozan:
It makes sense to have a seperate "InputNode" (which is currently `Node`)  and an "InternalNode".

moklick:
Thanks for the feedback so far! 

@rob-gordon Good call! I will add an `onNodesMeasured` callback or something like that.

ncthbrt:
What I'm a bit confused about is why width and height and handles and so on don't shouldn't just go into data for the purposes of SSR?

moklick:
@ncthbrt passing node.width/node.height also affects frontend only apps. As soon as you set it, we are using it as an inline style.

sroussey:
I start with all nodes having position of 0,0 (and wouldn't mind just having them blank instead), and then they go through layout. In order to do a layout, I need the height and width in order to calc positions.

Right now, I do this:
```ts
  const initialized = useNodesInitialized() && !nodes.some((n) => !n.computed);
  useEffect(() => {
    if (initialized) {
      setNodes(computeLayout(GraphPipelineLayout, nodes, edges));
      // window.requestAnimationFrame(() => fitView()); // todo: not working correctly 
    }
  }, [initialized]);
```

which seems hacky.

```ts
  const initialized = useNodesInitialized() && !nodes.some((n) => !getInternalNode(node.id)?.internals.width);
```
seems about the same. 

Anhow, does this mean that the user nodes are never mutated?

moklick:
The new approach is nice, but introduces new challenges:

a. it's very hard to implement stuff like `expandParent`, if we don't adjust dimensions and positions after the initial rendering. 
b. it's hard to tell if a node was already measured, if we don't add that information to the user node

I really like the separation of user nodes and internal nodes but not adding the dimensions doesn't work unfortunately. So the idea is to combine the current implementation with the new ideas from this proposal. For this I will try to keep the new internal node type and add the dimensions to the user nodes again.

PR #4061, opened 2024-03-23
Expose parentNode to custom nodes

RedPhoenixQ:
Exposing parentNode to custom nodes so that they can react to the parentNode changing. 

This implementation is only done for Svelte as I haven' t used and don't fully understand were and how to add it. The NodeProps type in system was changed which may affect the react types. I could try to make the changes I think would needed.


moklick:
Thanks, I think it's a good idea :) We will check it.

moklick:
Thanks for your input! We will rename `parentNode` to `parentId` and expose it.

PR #4129, opened 2024-04-10
Svelte: Refactor internals

peterkogo:


packages/svelte/src/lib/container/NodeRenderer/NodeRenderer.svelte
```diff
@@ -46,8 +41,8 @@
   {#each $visibleNodes as node (node.id)}
     {@const nodeDimesions = getNodeDimensions(node)}
     {@const posOrigin = getPositionWithOrigin({
-      x: node.computed?.positionAbsolute?.x ?? 0,
-      y: node.computed?.positionAbsolute?.y ?? 0,
+      x: node.internals.positionAbsolute.x ?? 0,
```
moklick, line 44: check if we need `??` and `positionAbsolute?.x`

packages/svelte/src/lib/store/initial-store.ts
```diff
@@ -91,20 +93,21 @@ export const getInitialStore = ({
   let viewport: Viewport = { x: 0, y: 0, zoom: 1 };
 
   if (fitView && width && height) {
-    const nodesWithDimensions = nextNodes.filter(
+    const nodesWithDimensions = Array.from(nodeLookup.values()).filter(
```
moklick, line 96: can we use `nodes` here?

moklick (changes requested):

packages/svelte/src/lib/store/index.ts
```diff
@@ -64,52 +64,71 @@ export function createStore({
...
+        }
+        case 'position':
+          node.position = nodeUpdate.position ?? node.position;
+          break;
+      }
+    }
+
+    store.nodes.update((nds) => nds);
 
     if (!get(store.nodesInitialized)) {
```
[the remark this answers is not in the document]
peterkogo: I could have just written `store.nodesInitialized.set(true);` but i don't want to trigger updates on nodeinitialized if it did not change.

PR #4146, opened 2024-04-11
Fix NodeResizer, handleParentExpand & add parentLookup

peterkogo:


packages/system/src/utils/store.ts
```diff
@@ -145,64 +157,72 @@ function calculateXYZPosition<NodeType extends NodeBase>(
   );
 }
 
-export function handleParentExpand(
-  nodes: InternalNodeBase[],
-  nodeLookup: NodeLookup
+type ExpandParentNode<NodeType> = NodeType & { parentId: string; expandParent: true };
+
+export function handleExpandParent(
+  nodes: ExpandParentNode<InternalNodeBase>[],
```
moklick, line 163: Let's change the API here and use something like `[{ parentId, rect }]` instead of nodes

packages/system/src/utils/store.ts
```diff
@@ -84,20 +87,19 @@ export function adoptUserNodes<NodeType extends NodeBase>(
 ) {
   const tmpLookup = new Map(nodeLookup);
   nodeLookup.clear();
+  parentLookup.clear();
+
   const selectedNodeZ: number = options?.elevateNodesOnSelect ? 1000 : 0;
-  const parentNodeIds = new Set<string>();
+  // const parentNodeIds = new Set<string>();
```
moklick, line 93: cleanup

packages/system/src/utils/store.ts
```diff
@@ -27,7 +28,7 @@ export function updateAbsolutePositions<NodeType extends NodeBase>(
     elevateNodesOnSelect: true,
     defaults: {},
   },
-  parentNodeIds?: Set<string>
+  parentLookup?: Map<string, InternalNodeBase<NodeType>[]>
```
moklick, line 31: when this is part of the store, it shouldn't be optional

packages/system/src/utils/store.ts
```diff
@@ -47,20 +48,21 @@ export function updateAbsolutePositions<NodeType extends NodeBase>(
           ...node.position,
           z: (isNumeric(node.zIndex) ? node.zIndex : 0) + (node.selected ? selectedNodeZ : 0),
         },
-        parentNode?.origin || options.nodeOrigin
+        parentNode?.origin ?? options.nodeOrigin
       );
 
       const currPosition = node.internals.positionAbsolute;
       const positionChanged = x !== currPosition.x || y !== currPosition.y;
```
moklick, line 55: do we still need this?

packages/system/src/utils/store.ts
```diff
@@ -47,20 +48,21 @@ export function updateAbsolutePositions<NodeType extends NodeBase>(
...
+      node.internals = {
+        ...node.internals,
+        positionAbsolute: positionChanged ? { x, y } : currPosition,
+        z,
+      };
 
-      if (parentNodeIds !== undefined) {
-        node.internals.isParent = !!parentNodeIds?.has(id);
+      if (parentLookup !== undefined) {
+        node.internals.isParent = !!parentLookup.has(id);
```
moklick, line 64: only use parentLookup for checking if a node is a parent

moklick (changes requested):

PR #4172, opened 2024-04-17
refactor(utils): only use node origin position helper

moklick:


moklick:
this is a bigger task and we should revise the whole position calculations: https://github.com/xyflow/xyflow/issues/4173

PR #4176, opened 2024-04-18
[v12] call `onEnd` in `XYResizer`'s `end` drag event handler

tonyf:
`XYResizerParams` defines an optional `onEnd` callback. 

```
type XYResizerParams = {
  domNode: HTMLDivElement;
  nodeId: string;
  getStoreItems: () => {
    nodeLookup: NodeLookup;
    transform: Transform;
    snapGrid?: [number, number];
    snapToGrid: boolean;
    nodeOrigin: NodeOrigin;
  };
  onChange: (changes: XYResizerChange, childChanges: XYResizerChildChange[]) => void;
  onEnd?: () => void;
};
```

`NodeResizeControl` sets this to send the final `dimensions` change with `resizing: false`.

```
onEnd: () => {
          const dimensionChange: NodeDimensionChange = {
            id: id,
            type: 'dimensions',
            resizing: false,
          };
          store.getState().triggerNodeChanges([dimensionChange]);
        },
```

Currently, the `onEnd` call is missing from XYResizer which means the final `dimensions` change is never fired. This fixes it


moklick:
Oh no! You were right then .. Thanks for the fix 🙏

PR #4180, opened 2024-04-18
Do not reset nodes and edges — they are controlled by the user

darabos:
I could be wrong about this. See the discussion in https://github.com/xyflow/xyflow/discussions/4179.


darabos:
I don't know if this breaks something else, but at least it fixes my issue.

moklick:
Thanks @darabos. We will check it next week. I am pretty sure that we can remove the nodes and edges reset.

PR #4195, opened 2024-04-22
Svelte: Fix Node event types

RedPhoenixQ:
Using `createEventDispatcher` from another imported function seems to break the svelte types and thus none of the node events where typed.

This PR fixes the issue by sharing only the types and not a create function. The `NodeEventMap` type was placed in the types directory to be accessible for both `NodeWrapper` and `NodeSelection`.

I noticed this when upgrading past 0.0.40 were there was a breaking change but no type errors to alert me were and what the new types were


moklick:
Thanks for your PR @RedPhoenixQ we will check it next week.

PR #4226, opened 2024-04-29
fix typing for `UpdateNodeInternals`

DenizUgur:


moklick:
good catch, @DenizUgur thanks!

Issue #4173, opened 2024-04-17
Internal nodes: calculate render position


moklick:
There are several places where we adjust a position based on the passed extent or origin. Maybe we could use those values to calculate a valid render position and then only work with those positions internally instead of `positionAbsolute`.

PR #4301, opened 2024-05-22
WIP: Include origin directly in positionAbsolute of node

peterkogo:
Resizing in subflows does not work yet 100%, my brain hurts...

Not so sure if this is a good idea. In some places node origin calculations disappear, just to reappear some other place. 
I have to double check if node origin can be removed from more places.
Also updating the `positionAbsolute` now also has to happen prematurely on measure otherwise the nodes snap into place a frame too late.


peterkogo:
related to #4173

PR #4323, opened 2024-06-03
exposed selectable, deletable, draggable to custom nodes & edges

### LINKED ISSUE #4269: Include 'deletable' property with EdgeProps and NodeProps

The `deletable` property exists on Nodes and Edges, but isn't exposed to NodeProps and EdgeProps. This means you can't use the `deletable` property for conditional rendering for Node and Edge types.

In my specific case, I want to hide the delete icon I normally show on Edges in my EdgeType if `deletable` is false. I tried to access `deletable` anyway, hoping it was just at the type level, but it's always undefined.

As a workaround, I'm adding my own `deletable` property in my Edge's `data` property.

peterkogo:
fixes #4269 

I took the liberty to add selectable & deletable to the node & edge props, as well as draggable to node props.


moklick:
Thanks!

PR #4397, opened 2024-06-26
Streamline Connections

peterkogo:


peterkogo:
If this gets merged we need to update the docs.

@moklick there are 2 places where I use Typescripts non-null assertion operator to assume that the node from the `nodeLookup` is actually present in XYHandle. I am pretty sure there is no way they would be undefined because the `handleBounds` from these nodes are used to determine them - double check me on this please

moklick:
This looks way better! Can we adjust the `useConenction` hook, so that it doesn't trigger an update when the connection is snapped to a handle?

Issue #4406, opened 2024-07-02
Is it possible to add highlighting to nodes only at a certain percentage of overlap?


gorbatkoff:
### Please describe the feature that you want to propose

Is it possible to add highlighting to nodes only if they were overlapped by more than X percent.

Example from official documentation

![image](https://github.com/xyflow/xyflow/assets/97447842/a9f014b2-1af5-447b-b995-366195f30337)

Here the nodes barely overlap each other, but the second node is highlighted.

Is it possible to implement similar functionality in future versions?

checkomkar:
Hey @gorbatkoff I have added option to add percentage value which you can give as input param to `getIntersectingNodes` function as follows
```js
getIntersectingNodes(node, true, undefined, 10)
```
Here I have added overlapping percentage as 10%, you can use value between 0-100.

The [PR](https://github.com/xyflow/xyflow/pull/4423)

Hope this helps. Thanks.

gorbatkoff:
What version will this MP go to and can it be duplicated with a hotfix for 11.11.5?

moklick:
We won't add this because it's too specific.

PR #4423, opened 2024-07-08
Option to intersection on certain input percentage. Issue #4406

checkomkar:
Issue #4406 
Added option through which user can define percentage area overlap.

```js
getIntersectingNodes: (nodeOrRect, partially = true, nodes, inputPercentOverlap)
```


moklick:
thanks but this is a bit too specific for the core library.

PR #4442, opened 2024-07-14
fix: type error

themuuln:
fixed: Type '"dots"' is not assignable to type 'BackgroundVariant | undefined'.ts(2322)

example code:         <Background variant='dots' bgColor='#0a0a0a' />


themuuln (commented):
just a minor type error fix

themuuln (commented):
just a minor change

moklick:
This is not a type error. You want to do `<Background variant={BackgroundVariant.Dots} />`, instead of `<Background variant="dots" />`

PR #4456, opened 2024-07-16
feat(types): accepts readonly arrays in props

pcorpet:
The library never modifies the arrays of nodes and edges given in properties, so we can accept readonly arrays.


peterkogo:
Thanks for the PR! This is correct. However, we will have to check what implications this has in the grand scheme of things. Unwanted mutations would be successfully prevented by this.

PR #4463, opened 2024-07-18
feat(hooks): useEdgesData

jclumingkit:
Added `useEdgesData()` hook that works the same as `useNodesData()` but for Edges.


moklick:
Thanks for your PR @jclumingkit ! We will have a look.

Issue #4448, opened 2024-07-15
getNodesBounds returns wrong values for nodes in subflow


hiepnguyen3001:
### What platform were you using when you found the bug?

- React Flow / Svelte Flow version: 12.0.1
- Browser and version: Chrome 126.0.6478.127
- OS and version: MacOS 13.3

### Live code example

https://codesandbox.io/p/sandbox/getnodesbound-test-k2f3p8?file=%2Fsrc%2FFlow.tsx

### Describe the Bug

When subflow's child nodes get passed in `getNodesBounds`, bounds are calculated based on their relative positions to the subflow, not their absolute positions to viewport.

### Steps to reproduce the bug or issue

1. Go to codesandbox link
2. Click on **Group A** node then **Group A.1** node to see their bounds calculated by `getNodesBounds`

### Expected behavior

**Group A** and **Group A.1** should share the same x,y values but instead **Group A.1** has x = 0, y = 0 which is its relative position in **Group A**

PR #4477, opened 2024-07-22
Added getNodesBounds function to useReactFlow/useSvelteFlow

peterkogo:
#4448

### DIFF
--- a/.changeset/dull-bags-know.md
+++ b/.changeset/dull-bags-know.md
@@ -1,0 +1,7 @@
+---
+'@xyflow/react': minor
+'@xyflow/svelte': minor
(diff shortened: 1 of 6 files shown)


peterkogo:
Possible solution for #4448

peterkogo:
We could kick getNodesBounds from xyflow/system, because it is inaccurate - or we restrict it to internal nodes only.

moklick:
@peterkogo what do you think about the idea to extend `getNodesBounds` with a `nodeLookup` param and then use it for `useReactFlow/getNodesBounds`. I also adjusted the deprecation message a bit (it's not really a deprecation message anymore).

Issue #4509, opened 2024-08-01
`nodesConnectable` prop doesn't seem to be reactive to value changes


mcmxcdev:
### What platform were you using when you found the bug?

- Svelte Flow version: 0.0.40
- Browser and version: Version 1.68.128 Chromium: 127.0.6533.73 (Official Build) (64-bit)
- OS and version: Ubuntu 24.04

### Live code example

https://stackblitz.com/edit/svelte-flow-ts-ppskg3?file=src%2FFlow%2Findex.svelte

### Describe the Bug

In the Stackblitz above, `elementsSelectable={!isLocked}` updates reactively, meaning that nodes are correctly not selectable if canvas is locked.

`nodesConnectable={!isLocked}` on the other hand does not seem to update reactively and nodes are still connectable even when `isLocked` is true

### Steps to reproduce the bug or issue

See Stackblitz for reproduction

### Expected behavior

I would expect not to be able to draw any edges between nodes whenever `nodesConnectable` evaluates to false.

peterkogo:
This should be fixed in the newest update

mcmxcdev:
Thanks, already using 0.1.15 which now works perfectly for our use case!

PR #4510, opened 2024-08-01
fix(svelte): rename `isConnectable` to `handleConnectable`

bcakmakoglu:
# 🚀 What's changed?

- Renamed `isConnectable` prop locally to `isConnectableProp` to avoid naming collision with derived var `isConnectable`
	- Apparently naming it the same thing as the prop just causes the cmp to prefer the prop instead of the local variable (Peter would know more about this though)

# 🐛 Fixes

- [x] [#4509](https://github.com/xyflow/xyflow/issues/4509)

### DIFF
--- a/.changeset/curly-llamas-obey.md
+++ b/.changeset/curly-llamas-obey.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/svelte': patch
+---
+
+Rename `isConnectable` prop locally to `isConnectableProp` to avoid naming collision with derived value of `isConnectable` in `<Handle>` component.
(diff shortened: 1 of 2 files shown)


peterkogo:
Even though `handleConnectable` is the naming in React as well, I would much rather rename the prop internally to `isConnectableProp` via 
```svelte
let isConnectableProp: $$Props['isConnectable'] = undefined;
export { isConnectableProp as isConnectable }
```
and leave the internal variable `isConnectable`. `handle-` is a little ambiguous and can be read as a verb aka function definition.

Otherwise, this does indeed seem to fix #4509

bcakmakoglu:
@peterkogo PR is updated :) thanks for the feedback

moklick:
Thanks @bcakmakoglu and @peterkogo for the review

PR #4517, opened 2024-08-05
fix(svelte): make svelte-preprocess a dev dep

ghostdevv:


ghostdevv:
I added a changeset, let me know if I should remove it :pray:

moklick:
Thanks @ghostdevv

ghostdevv:
@moklick It's probably not worth it since you don't have vite already setup and running something in `packages/svelte`

PR #4536, opened 2024-08-12
fix(onlyRenderVisible) render nodes with fixed width & height intially

### LINKED ISSUE #4516: onlyRenderVisibleElements with width and height set on node doesn't render edge if one of the nodes is offscreen

- React Flow / Svelte Flow version: 12.0.4 (react)
- Browser and version: Chrome 126.0.6478.183
- OS and version: MacOS 14.5

### Live code example

https://codesandbox.io/s/prod-moon-8fw277?file=/App.tsx

### Describe the Bug

If `onlyRenderVisibleElements` is set and nodes have an explicit `width` and `height` set, then edges between nodes will not render if one of the nodes is initially offscreen. If I move the `width` and `height` into the `style` object then everything works as expected.

This may be user error, but it was quite frustrating because _everything else was working fine_, the only symptom of something being wrong was the edge not rendering if one node was offscreen.

1. Open the codesandbox link
2. See that "Node 1" is visible and there's no edge visible
3. Pan the screen down until "Node 2" appears
4. See an edge pop into existence

If you change the codesandbox and set `nodes={initialNodesWithStyles}` you'll see it behave as expected.

### Expected behavior

1. Edge should be immediately visible because one of the two nodes is visible.

### Screenshots or Videos

https://github.com/user-attachments/assets/b9b96a23-cbd3-46ac-828e-b329d73d3d05

### Additional context

_No response_

peterkogo:
closes #4516 

Rearranged some code for readability but only 1 line changes from 
```js
const notInitialized = width === null || height === null;
```
to
```js
const forceInitialRender = !node.internals.handleBounds;
```

### DIFF
--- a/.changeset/tidy-swans-clap.md
+++ b/.changeset/tidy-swans-clap.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/system': patch
+---
+
+fix(onlyRenderVisible) edges to offscreen nodes with fixed width & height displayed correctly
--- a/packages/system/src/utils/graph.ts
+++ b/packages/system/src/utils/graph.ts
@@ -191,21 +191,22 @@ export const getNodesInside = <NodeType extends NodeBase = NodeBase>(
 
   const visibleNodes: InternalNodeBase<NodeType>[] = [];
 
-  for (const [, node] of nodes) {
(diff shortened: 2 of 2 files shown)


moklick:
nice!

PR #4538, opened 2024-08-12
fix(system,errors): use handle id of matching handle type for errors

bcakmakoglu:
# 🚀 What's changed?

- Use the handle id of the matching handle type when warning about edges that can't be created due to missing handle ids

# 🐛 Fixes

- [x] the warning `e008` always showing `null` as the handle id even if one was passed (although no matching handle was found)


moklick:
thanks @bcakmakoglu

Issue #4539, opened 2024-08-12
[Feature]: Add `XYFlowError` that provides error arguments and allows for easier discerning of errors


bcakmakoglu:
### Please describe the feature that you want to propose

If `<ReactFlow>` is given an `onError` callback, this callback is fed two arguments - the error *id* and the error *message*.
Most users are not familiar with what the error id corresponds to, requiring a lookup in the original source code or reading the message itself to understand what the error is about.
This makes it a bit difficult when writing a custom `onError` handler that possibly does something specific to react to such errors (besides just logging it).

For example:
```tsx
<ReactFlow 
	/* ... some props */
	onError={(id, message) => {
		// '008' seems like an obscure id to compare against since it doesn't really tell us what error this actually is unless we specifically look it up
		if (id === '008') {
			// an edge couldn't be created due to missing source/target handle - let's throw
			throw new Error('sad')
		}
	}
```

Maybe the ids could be assigned a more obvious name like instead of `008` it's id would be `err_handle_missing`.
Or an actual enum could be used here to make it easier.

It would also be nice if the errors could provide the info that's contained in the message as well, like in the error above we could be informed of what *type* of handle this failed for, what it's id was supposed to be etc.

I was thinking of possibly just adding an actual `XYFlowerror` for this purpose.

For example:
```ts
export enum ErrorCode {
  NODE_MISSING_PARENT = 'NODE_MISSING_PARENT',
  // ... more errors
}

const messages = {
  [ErrorCode.NODE_MISSING_PARENT]: (id: string, parentId: string) => `Node is missing a parent\nNode: ${id}\nParent: ${parentId}`,
  // more messages
} as const

type ErrorArgs<T extends ErrorCode> = (typeof messages)[T] extends (...args: any[]) => string
  ? Parameters<(typeof messages)[T]>
  : never

export class XYFlowError<T extends ErrorCode = ErrorCode, Args extends ErrorArgs<T> = ErrorArgs<T>> extends Error {
  name = 'XYFlowError'
  code: T
  args: Args

  constructor(code: T, ...args: Args) {
    super(messages[code]?.(...args))
    this.code = code
    this.args = args
  }
}
```

```ts
// call the `onError` callback
onError(new XYFlowError(ErrorCode.NODE_MISSING_PARENT, parentId)
```

```tsx
function isErrorOfType<T extends ErrorCode>(error: XYFlowError, code: T): error is XYFlowError<T> {
  return error.code === code
}


// receive the error in the callback
[...]

PR #4540, opened 2024-08-12
feat: add `XYError`

bcakmakoglu:
# 🚀 What's changed?

- Add `XYError` class to handle errors
- Add `XYErrorCode` enum to handle assignable error codes
- Add `isXYError` utility function to check if an error is an instance of `XYError`
- Allow an object of `XYError` as third argument for the `onError` callback
	- Pass `XYError` when handling errors and pass it to the `onError` callback

# 🌻 Resolves

- [x]  #4539


moklick:
Thanks for your help @bcakmakoglu . When I see it right this wouldn't be a breaking change anymore, since you just added another param to the onError function. Is that correct?

bcakmakoglu:
@moklick yep, that is correct :)

Seems like a conflict came up though, I'll update the PR asap

bcakmakoglu:
Could probably be improved by passing `lib` to the error as well so that the messages become something like `[ReactFlow]: <message>` / `[SvelteFlow]: <message>` if you'd like that

PR #4548, opened 2024-08-13
Improve search algorithm for close handles

### LINKED ISSUE #3983: useUpdateNodeInternals doesn't update handle positions during active connection

### LINKED ISSUE #4027: Unexpected connection behaviour when using a updateNodeInternals
### Describe the Bug

I want to change height of node when starting connection. But it becomes invisible and node is not updating. How to fix it?

https://codesandbox.io/p/sandbox/mystifying-meadow-lc6kfs?file=%2Fstyle.css&utm_medium=sandpack

Start connecting
Current connection becomes invisible

### Expected behavior

Current connection visible and handle moves to the center (and doesn't stay in old center)

### Screenshots or Videos

_No response_

### Platform

- Version: [e.g. 91.1]

### Additional context

_No response_

### LINKED ISSUE #4520: Promote handleLookup to be persistent in the store
The way we currently do connections for performance reasons is to create a handleLookup on PointerDown and use that for the remainder of the connection process.

This approach does not work if any nodes change during the connection process, and will quite frankly break all collaborative flows eventually.

peterkogo:
fixes #4027 
closes #4520 
related #3988
fixes #3983

### DIFF
--- a/packages/react/src/components/Handle/index.tsx
+++ b/packages/react/src/components/Handle/index.tsx
@@ -159,6 +159,7 @@ function HandleComponent(
       isValidConnection: isValidConnectionStore,
       lib,
       rfId: flowId,
+      nodeLookup,
     } = store.getState();
 
     if (!nodeId || (!connectionClickStartHandle && !isConnectableStart)) {
@@ -187,6 +188,7 @@ function HandleComponent(
       flowId,
       doc,
       lib,
+      nodeLookup,
     });
 
     if (isValid && connection) {
--- a/.changeset/mighty-nails-cover.md
+++ b/.changeset/mighty-nails-cover.md
@@ -1,0 +1,6 @@
+---
+'@xyflow/react': patch
+'@xyflow/system': patch
+---
+
+Replaced algorithm used for searching close handles while connecting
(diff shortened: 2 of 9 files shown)


packages/react/src/components/Handle/index.tsx
```diff
@@ -159,6 +159,7 @@ function HandleComponent(
       isValidConnection: isValidConnectionStore,
       lib,
       rfId: flowId,
+      nodeLookup,
```
moklick, line 162: I think we would need the same changes for Svelte Flow too?

moklick (changes requested):

peterkogo:
related #3988 
fixes #3983

packages/react/src/components/Handle/index.tsx
peterkogo: This is just used in context of click connect which is not present in Svelte Flow

PR #4572, opened 2024-08-19
fix node extent

### LINKED ISSUE #4529: Different behaviour between coordinate extent and parent extent

- React Flow / Svelte Flow version: 12

### Live code example

https://codesandbox.io/p/sandbox/frosty-raman-vz4qfz

### Describe the Bug

Extent "parent" will prevent the node from exiting the parent, while a coordinate extent will prevent only the node origin from exiting the extent.

Make a child node with coordinate extent

### Expected behavior

Child node is entirely limited to the extent same as when using "parent" extent

### Screenshots or Videos

_No response_

### Additional context

_No response_

peterkogo:
closes #4529

### DIFF
--- a/.changeset/rude-rockets-hope.md
+++ b/.changeset/rude-rockets-hope.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/svelte': minor
+---
+
+Added ability to set global nodeExtent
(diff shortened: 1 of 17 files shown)


peterkogo:
Decide if we want to tackle fixing the `nodeExtent` behavior once and for all. As things are streamlined now we can see if we'd like to streamline position clamping on a more fundamental level.

peterkogo:
Things work as expected now, but we should see if there is a way to simplify the code.

moklick:
This looks good already! One thing that doesn't work is an initial nodeExtent for child nodes.

PR #4578, opened 2024-08-21
Fix: Include `onViewportChange` in `useEffect` Dependency Array to Prevent Stale References

### LINKED ISSUE #4579: `onViewportChange` Function Not Updated When Dependencies Change in useEffect Hook

- React Flow / Svelte Flow version: 12.1.1
- Browser and version: n/a
- OS and version: n/a

### Live code example

https://codesandbox.io/p/sandbox/cocky-leftpad-w9z9nz

### Describe the Bug

There is a bug in the `useEffect` hook within the `zoomPane` setup code where the `onViewportChange` function is used. The `useEffect` hook does not include `onViewportChange` in its dependency array. As a result, if the `onViewportChange` function is updated, the effect continues to reference the stale closure of the function, leading to unexpected behavior where the old version of the function is called instead of the new one.

1. Open sandbox and run preview
2. Click "Use ViewportStore2" Button
3. Try to move or zoom the pane

### Expected behavior

The pane should move. But since the stale `onViewportChange` changes the ViewportStore1, all changes will be applied to that. When you switch back you will see exactly that.

### Screenshots or Videos

_No response_

### Additional context

_No response_

alexanderhorner:
This merge request fixes the issue where the `onViewportChange` function was not included in the `useEffect` dependency array (inside `ZoomPane`), causing stale references to be used. This caused the viewport to break if a new `onViewportChange` function was passed.

Fixes #4579

### DIFF
--- a/packages/react/src/container/ZoomPane/index.tsx
+++ b/packages/react/src/container/ZoomPane/index.tsx
@@ -103,7 +103,7 @@ export function ZoomPane({
         panZoom.current?.destroy();
       };
     }
-  }, []);
+  }, [onViewportChange]);
 
   useEffect(() => {
     panZoom.current?.update({


peterkogo:
Thanks for the PR! Good catch, all of the callback functions should be updateable. 
Unfortunately it's not this easy. We need to add the funcitons to the update function of XYZoom and add them to the dependecy array there.

PR #4628, opened 2024-09-05
test(adoptUserNodes): add unit test

moklick:


examples/react/package.json
```diff
@@ -16,6 +16,7 @@
   "dependencies": {
     "@reduxjs/toolkit": "^2.2.3",
     "@xyflow/react": "workspace:*",
+    "@xyflow/system": "^0.0.41",
```
peterkogo, line 19: this should also bei workspace:* right?

moklick: yep

Issue #4652, opened 2024-09-13
Calculate viewport dimensions in `fitView` action


bcakmakoglu:
### What platform were you using when you found the bug?

- React Flow / Svelte Flow version: current
- Browser and version: -
- OS and version: -

### Live code example

https://codesandbox.io/p/sandbox/distracted-sea-9vjlqr

### Describe the Bug

There is a discussion right here: https://github.com/xyflow/xyflow/discussions/4575#discussioncomment-10410026

> 
> * When you open/close the sidebar, the dimensions of the viewport are re-calculated
> * The dimensions are used when calling fitView
>      * Even though the sidebar opens, fitView is triggered at a time where it probably uses the previously recorded dimensions instead of those with/without the sidebar, making it seem like fitting the view "failed"

I tried it out locally and in `useViewportHelper` I had `fitView` calculate the current dimensions directly instead of using the `width` / `height` values from the store state, which seems to solve this issue without having to add some arbitrary waiting time to it.

```ts
fitView: (options) => {
  const { nodeLookup, minZoom, maxZoom, panZoom, domNode } = store.getState();
  const fitViewNodes = getFitViewNodes(nodeLookup, options);

  const size = getDimensions(domNode!);

  return panZoom
    ? fitView(
        {
          nodes: fitViewNodes,
          ...size,
          minZoom,
          maxZoom,
          panZoom,
        },
        options
      )
    : Promise.resolve(false);
},
```

I assume it's because `useResizeHandler` relies on `useEffect` to re-calculate the dimensions and put them into the store but by the time that is done `fitView` was already triggered from the user-side with dimensions that don't match what they expected

I don't see any particular downside to calculate the dimensions in `fitView` itself instead of reading the stored values but maybe I'm missing something.

### Steps to reproduce the bug or issue

1. Open Sandbox
2. Remove timeout
3. Toggle Sidebar
4. fitView won't work properly because of false dimensions used in the calculations

With the timeout we give the `useEffect` enough time to run and update the stored dimensions for fitView but preferably this wouldn't be necessary

### Expected behavior

`fitView` uses the current viewport dimensions without the need to wait for effects to run first.

PR #4653, opened 2024-09-13
fix(react,svelte): calculate viewport dimensions inside of fitView action

bcakmakoglu:
# 🚀 What's changed?

- Calculate viewport dimensions for fitView in the action itself instead of using the stored viewport dimensions
	- When using stored dimensions, the dimensions update happens too late because it's based on an effect that has to run first; if we just grab the dimensions directly from the `domNode` element we don't run into the issue of using stale dimensions when the size of the viewport has changed

# 🐛 Fixes

- [x] [#4652](https://github.com/xyflow/xyflow/issues/4652)

### DIFF
--- a/.changeset/smart-insects-tease.md
+++ b/.changeset/smart-insects-tease.md
@@ -1,0 +1,6 @@
+---
+'@xyflow/react': patch
+'@xyflow/svelte': patch
+---
+
+Calculate viewport dimensions in fitView instead of using stored dimensions. Fixes [#4652](https://github.com/xyflow/xyflow/issues/4652)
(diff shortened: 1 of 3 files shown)


bcakmakoglu:
Seems to work well for React but in Svelte I can't get it to use the latest dimensions even if I do `getDimensions(get(store.domNode))` @peterkogo any idea why that might be the case here? 🤔

bcakmakoglu:
PR is updated to include Svelte as well :)

peterkogo:
Awesome! Will take a last close look tomorrow, but this already looks super mergable :)

PR #4672, opened 2024-09-19
Chore: Update all dev-dependencies

peterkogo:
In the process of getting rid of the false-alarm security advisories on the repository, I took the time to simply update all the dev dependencies in the mono repo, and check if any major releases broke something - they did not.


peterkogo:
2 new typescript errors popped up.

- `nodes.values().next().value.selected`: object might be undefined. Fixed by: `nodes.values().next().value?.selected`
- For some reason `PropsWIthoutRef` got more strict when forwarding refs. To be fair, our type was indeed not 100% correct but props > propWithoutRef

Anyway, both seem pretty minor but I am not sure if this warrants a minor/patch because the type was changed slightly.

moklick:
We still support React 17. This change would drop support for React17 types. Is the `PropsWithoutRef` change related or does it with with 17 types, too?

peterkogo:
Learnings: 
- We can't update past Rollup ^4.22.4 because we will get Typescript `outDir` to Rollup `file` mismatch error. With our current setup it's probably not fixable. https://github.com/rollup/plugins/issues/1230
- Newest typescript version pose compatibility problems (at least we now have 1 typescript version across the project)
- sometimes you need to delete all node_modules and run 'pnpm prune store' because it sometimes just does not do what it should...
- Using meta or backspace keys in playwright on webkit does not work
- e2e is the equally the most reliable and most unreliable type of testing

Issue #4724, opened 2024-10-10
useHandleConnection for broader selections


peterkogo:
Based on the request #4721 it is actually doable to collect all connections for each level of the current `connectionLookup` structure: `nodeId`, `node-id`-`handleType` instead of only `nodeId-handleType-handleId`.

moklick:
What do you think about a `useNodeConnections` hook? I think @bcakmakoglu also mentioned this idea once.

nilaygit-10721:
can you assign me this issue?

PR #4725, opened 2024-10-10
add `useNodeConnections`, deprecate `useHandleConnections`

peterkogo:
PR for the idea proposed in #4724.

We track the handleConnections for `node` and `node-handleType` as well — not only for `node-handleType-handleID`.

This way `useHandleConnections()` returns all handleConnections for the node and `useHandleConnections({ type: 'source' })` returns all handleConnections for the node of type `source`.


moklick:
I think it would be more clear if we create a new hook called `useNodeConnections` or something like that. wdyt @peterkogo ?

peterkogo:
@moklick I agree that it makes sense to have a deliberate `useNodeConnections`  hook because people are actually looking for something like this. 

However, I would not deliberately take out new functionality from `useHandleConnections`.
I think it's pretty sleek to be able to just do `useHandleConnections({type: 'source'})` and in extent also just allow `useHandleConnections()`as all of this would be a non-breaking enhancement.

moklick:
Good point! Then we could offer this solution first and then add a `useNodeConnections` hook later.

peterkogo:
Fixed the naming and deprecated the old functions. Will create docs shortly. 

One last thing: the returned type is still a HandleConnection. Should we do something about it?

moklick:
I think the return type is OK. Two tiny things: 1. the url should be /use-node-connections and could you use the new hook in the internal examples instead of `useHandleConnections`?

PR #4772, opened 2024-10-30
chore(NA): adds explicit browser, node and default exports resolution

mistic:
This PR follows up on what was done at https://github.com/xyflow/xyflow/pull/4755.

Even having into account the changes in the previous PR, webpack v5 still resolves `@xyflow/react/dist/esm/index.mjs` which then tries to resolve `react/jsx-runtime` within (and in that case it loaded from a .mjs which needs to be `react/jsx-runtime.js`).

The changes I'm proposing introduces a more explicit split in the resolution without room for wrong resolves going further.

\cc @peterkogo @moklick

### DIFF
--- a/.changeset/famous-lobsters-ring.md
+++ b/.changeset/famous-lobsters-ring.md
@@ -1,0 +1,6 @@
+---
+'@xyflow/system': patch
+'@xyflow/react': patch
+---
+
+Splits exports field in package.json to support an explicit resolution for browser, node and default to resolve issues with webpack esm module resolution.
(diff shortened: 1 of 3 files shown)


peterkogo:
Thank you so much for this PR! This saved us a lot of time! 

I tested this export configuration with React 17 & 18, with `Next.js`, `CRA`, vanilla Webpack and Vite. I really hope I covered all possible configurations this time. 

What a mess this whole thing is... 😄

mistic:
@peterkogo thank you so much for taking a look into this. I'm confident this will be enough to solve the problem. Unfortunately feels like currently the ecosystem is in a transition state so we got to live with this mess for a while 😃

PR #4789, opened 2024-11-07
Enhance useKeyPress top also support '+'

### LINKED ISSUE #4777: `useKeyPress` does not allow for a combination of a key and the plus button

- React Flow / Svelte Flow version: current

### Live code example

_No response_

### Describe the Bug

Currently it's not possible to do a key combination like `ctrl++` in order to track a keypress combination of a key in addition to the plus button since it's key would just be `+` which is used to assert that you want to create a combination of keys.

The relevant code bit would be this part in `useKeyPress.ts`

```ts
const keys = keyCodeArr.filter((kc) => typeof kc === 'string').map((kc) => kc.split('+'));
```

One use-case would be to overwrite the browser zooming and allow a user to zoom in/out using something like ctrl and the plus/minus btn (minus isn't an issue here though, it's only the `+` btn that's a problem obviously 😄)

```ts
// this does *not* work
useKeyPress('ctrl++')
```

### Expected behavior

It should somehow be possible to still use the `+` btn for a key combination.

### Screenshots or Videos

_No response_

### Additional context

_No response_

peterkogo:
Closes #4777

### DIFF
--- a/.changeset/clever-keys-pay.md
+++ b/.changeset/clever-keys-pay.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': minor
+---
+
+Support key combinations which include '+' (e.g., 'Control++' resolves to the combination 'Control' and '+').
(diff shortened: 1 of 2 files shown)


bcakmakoglu:
nice 👍

Issue #4774, opened 2024-11-01
multi call updateEdge not working


zhenghan2017:
### What platform were you using when you found the bug?

- React Flow / Svelte Flow version: 12.3.4
- Browser and version: Chorme 130.0.6723.70
- OS and version: macOS Sonoma 14.0

### Describe the Bug

When I call `updateEdge` in loop，only the first update is active.

like below code
```
for (const edgeUpdate of edgeUpdates) {
  updateEdge(edgeId, edgeUpdate);
}
```

### Steps to reproduce the bug or issue

I read  `useQuene` and find the code 

```
useIsomorphicLayoutEffect(() => {
    // Because we need to flip the state back to false after flushing, this should
    // trigger the hook again (!). If the hook is being run again we know that any
    // updates should have been processed by now and we can safely clear the queue
    // and bail early.
    if (!shouldFlush) {
      queue.reset();
      return;
    }

    const queueItems = queue.get();

    if (queueItems.length) {
      runQueue(queueItems);

      queue.reset();
    }

    // Because we're using reactive state to trigger this effect, we need to flip
    // it back to false.
    setShouldFlush(false);
  }, [shouldFlush]);
```

If push `updateEdge` Faster thanReact state update，the quene will be reset。

### Expected behavior

updateEdge can active all

peterkogo:
Please create a minimal reproduction for this at [new.reactflow.dev/ts](https://new.reactflow.dev/ts).

zhenghan2017:
I see the fixed, it's also good for me. thanks

Issue #4680, opened 2024-09-23
`updateNodeData` + React 17 


ludenge:
### What platform were you using when you found the bug?

- React Flow / Svelte Flow version: 12.3.0
- React version: 17.0.2
- Browser and version: 128.0.6613.138
- OS and version: 14.1.2 (23B92)

### Describe the Bug

Based on official examples, Modify content NumberInput.js
https://reactflow.dev/learn/advanced-use/computing-flows#computing-data

```
 import React, { useCallback, useState, useEffect } from "react";
import { Handle, Position, useReactFlow } from "@xyflow/react";

function NumberInput({ id, data }) {
  const { updateNodeData } = useReactFlow();

  const onChange = useCallback((evt) => {
    const cappedNumber = Math.min(255, Math.max(0, evt.target.value));
    updateNodeData(id, { value: cappedNumber });
  }, []);

// ------- add start --------
const [refreshKey, setRefreshKey] = useState(String(Date.now()));

  useEffect(() => {
    setRefreshKey(String(Date.now()));
  }, []);

  useEffect(() => {
    updateNodeData(id, {
      text: refreshKey,
    });
  }, [refreshKey]);
// ------- add end --------

  return (
    <div className="number-input">
      <div>{data.label}</div>
      <input
        id={`number-${id}`}
        name="number"
        type="number"
        min="0"
        max="255"
        onChange={onChange}
        className="nodrag"
        value={data.value}
      />
      <Handle type="source" position={Position.Right} />
    </div>
  );
}

export default NumberInput;

```

NumberInput will become non-editable, updateNodeData will not execute

### Steps to reproduce the bug or issue

1. React version 17.0.2
2. add the following code block to node
```
const [refreshKey, setRefreshKey] = useState(String(Date.now()));

  useEffect(() => {
    setRefreshKey(String(Date.now()));
  }, []);

  useEffect(() => {
    updateNodeData(id, {
      text: refreshKey,
    });
  }, [refreshKey]);
```

### Expected behavior

Expected to be able to input content

### Screenshots or Videos

![img_v3_02f0_7cd5e0e5-c80f-48f6-ae74-891ca9193acg](https://github.com/user-attachments/assets/efd984a4-4253-4e3c-b7c1-251060132960)

ludenge:
Changing useIsomorphicLayoutEffect to useEffect will fix this

bcakmakoglu:
[...]

PR #4796, opened 2024-11-08
Fix deadlock in useQueue

### LINKED ISSUE #4779: Calling reactFlow.updateEdgeData or reactFlow.updateNodeData in multiple places at the same time freezes the graph

- React Flow / Svelte Flow version: Latest
- Browser and version: Chrome, Latest
- OS and version: Windows 11, Latest

### Describe the Bug

I have a project where I'm making multiple API calls that can arrive at different times, and I'm trying to update node & edge data as they arrive. 

Whenever I start making a lot of API calls and two happen to arrive at the same time, I update edge and node data simultaneously. I'm yet to figure out how to trigger it every time, but in my app, the more requests the user makes, the more likely it is that the bug triggers.

From the moment it happens, the graph can no longer be updated with new edges or nodes, or edge or node updates, it essentially turns into "read-only mode".

1. Simulate a lot of async calls using setTimeout
2. When those calls finish, call useReactFlow().updateNodeData or updateEdgeData
3. Nodes/Edges stop being updated, added, or removed!

### Expected behavior

I would expect the graph to be able to accept multiple modifications at the same time and properly figure out how to deal with them (merging updates), or at least be able to accept updates across different nodes and edges. Instead, the state becomes locked completely whenever the bug happens.

### Screenshots or Videos

_No response_

### Additional context

- I disabled all updateEdgeData calls, and the bug started appearing only for nodes (so nodes couldn't be added or updated, but edges were fine)
- I disabled all updateNodeData calls, and the bug disappeared from nodes completely!
- This happens for updates across different nodes and edges, not just updating the same edge/node.
- I've only recently updated my app to use updateNode/EdgeData. Previously, I would use React's `useContext` hook to pass setters/updaters around, and that worked fine with lots of requests arriving at the same time.
- All graph updates/modifications stop working, not just updateNodeData or updateEdgeData.

Aki-7:
This pull request includes changes to the `useQueue` hook to fix deadlock due to automatic batching in React.

Details:

* Replaced the `shouldFlush` boolean state with a `serial` state using `BigInt` to avoid deadlock issues with automatic batching.
* Updated the queue creation to increment the `serial` state instead of setting the `shouldFlush` flag.
* Removed unnecessary state flipping logic and simplified the `useIsomorphicLayoutEffect` dependency to only depend on `serial`.

closes #4779 
Related Issues: #4680, maybe #4774

@peterkogo Could you please review this?

### DIFF
--- a/.changeset/funny-shrimps-remain.md
+++ b/.changeset/funny-shrimps-remain.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
+Fixed number of issues connected to batching node & edge updates.
--- a/packages/react/src/components/BatchProvider/useQueue.ts
+++ b/packages/react/src/components/BatchProvider/useQueue.ts
@@ -13,28 +13,21 @@ import { Queue, QueueItem } from './types';
  */
 export function useQueue<T>(runQueue: (items: QueueItem<T>[]) => void) {
   // Because we're using a ref above, we need some way to let React know when to
-  // actually process the queue. We flip this bit of state to `true` any time we
-  // mutate the queue and then flip it back to `false` after flushing the queue.
-  const [shouldFlush, setShouldFlush] = useState(false);
+  // actually process the queue. We increment this number any time we mutate the
+  // queue, creating a new state to trigger the layout effect below.
+  // Using a boolean dirty flag here instead would lead to issues related to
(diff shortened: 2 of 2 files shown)


peterkogo:
Related Issues: #4680, maybe #4774

moklick:
Hey @Aki-7 thanks for your help here 🙏 Highly appreciated!

PR #4826, opened 2024-11-18
fix(panel): forward ref to inner panel div

chrtze:
this pull requests adds a `ref` prop to the panel component that can be used to forward a ref to the inner div


moklick:
good idea!

PR #4834, opened 2024-11-20
fix getNodeRect loses node origin

gmvrpw:
Good afternoon, I really like your library and its approach to openness and monetization. I'm using `react-flow` to demo my open source library and came across a bug, which can be demonstrated with an example from the documentation (Intersection example)

To get this bug I changed the description of the second node by adding `origin: [0.5, 0]` and got the wrong `rect` for the changed node:
issue

Digging into the code I found the following bug in the use of `nodeToRect` in the `getNodeRect` function:
`nodeToRect` expects the actual `origin` value when retrieving a node, but `getNodeRect` forgets about this field when creating `nodeWithPosition`.

I fixed this call in what I think is the most non-invasive and correct way and now everything works as expected
`origin: [0.5, 0]`
fixed

I have also investigated all other places where `nodeToRect` is used (thx to `grep`) and all of them use correct nodes (actual origin, since they are obtained from the store) so this is the only problem of this kind.

### DIFF
--- a/packages/react/src/hooks/useReactFlow.ts
+++ b/packages/react/src/hooks/useReactFlow.ts
@@ -60,6 +60,7 @@ export function useReactFlow<NodeType extends Node = Node, EdgeType extends Edge
         width: nodeToUse.measured?.width ?? nodeToUse.width,
         height: nodeToUse.measured?.height ?? nodeToUse.height,
         data: nodeToUse.data,
+        origin: nodeToUse.origin,
       };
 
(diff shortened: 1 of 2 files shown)


peterkogo:
Thank you very much for your PR. By looking at your code I realized we are making things more complicated to begin with. I created a 'proper' fix for this issue over at #4854

peterkogo:
Your solution was actually correct in the end! I make sure to mention you in the changelog.

PR #4854, opened 2024-11-28
Fix/intersection origin

peterkogo:
#4834


peterkogo:
This is a no good. I made a mistake in my reasoning here.

Issue #4821, opened 2024-11-17
Ability to provide pathLength to edge


mhuggins:
### Please describe the feature that you want to propose

Hello,

I'm trying to animate newly added edges via custom CSS rules.  I'd like to leverage the `stroke-dasharray` CSS rule to make it appear as though the edge is being drawn from the source to the target.  The problem is that without specifying a [`pathLength`](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/pathLength) property on the `path` element, there's not a definitive way to consistently perform this using CSS alone.

Would is be possible to add this as an option for edges without the need to create a custom Edge component that's nearly identical to the ones provided?  Alternatively, would it be possible for `BaseEdge` to support other `path` attributes as props such that it can be leveraged easily when implementing custom Edge components?

Thank you!

peterkogo:
We should definitely add the ability to pass any props to the path that a `BaseEdge` rendered. We should also forward the ref of the path when we are at it.

We don't want to add `pathLengh` as an option for the `BuiltInNodes` because they are just really there to simplify the first steps of working with React Flow. We want to prevent adding complexity to the whole data model, for something that can be easily solved within custom edges.

mhuggins:
Makes sense.  I'm happy to put together a PR for the props + ref aspect, if that helps!

mhuggins:
This was fixed in the linked PR! 😄

moklick:
Thanks! This is fixed in 12.3.6.

PR #4855, opened 2024-11-28
feat: accept any `path` element attribute as a prop

mhuggins:
This is an attempt to implement a solution for #4821.

Unfortunately, I could not seem to get my development environment set up following the contributing instructions.  After running `pnpm install`, I was still seeing lint/typecheck errors before even beginning to make changes.

### DIFF
--- a/.changeset/selfish-tables-tie.md
+++ b/.changeset/selfish-tables-tie.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': minor
+---
+
+Support all `path` element attributes in React edge components.
--- a/.changeset/selfish-tables-tie.md
+++ b/.changeset/selfish-tables-tie.md
@@ -2,4 +2,4 @@
 '@xyflow/react': minor
 ---
 
-Support all `path` element attributes in React edge components.
+Support passing `path` element attributes to BaseEdge component.
--- a/.changeset/selfish-tables-tie.md
+++ b/.changeset/selfish-tables-tie.md
@@ -1,5 +1,5 @@
 ---
-'@xyflow/react': minor
+'@xyflow/react': patch
 ---
 
-Support passing `path` element attributes to BaseEdge component.
+Support passing `path` element attributes to `BaseEdge` component.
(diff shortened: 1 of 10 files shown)


peterkogo:
@mhuggins Thanks for the initial PR! I saw that we can simplify a couple of things here and there.

I tried passing `pathLength` to the `<BaseEdge />` which now works but Typescript is complaining. 

I don't understand why even this is throwing a Typescript error:
```ts
const x : React.HTMLAttributes<SVGPathElement> = { pathLength: 5 };
```

Do you have any idea?

mhuggins:
@peterkogo I fixed this by changing `HTMLAttributes<...>` to `SVGAttributes<...>` for the typing.  I realized there were a couple other places that appeared to use this generic type incorrectly, so I updated those too.

I also added an example that demonstrates the `pathLength` being used in conjunction with a CSS animation that leverages the known path length.

moklick:
Thanks for your help @mhuggins! I like the idea to pass svg attrs to the `BaseEdge` component but I am not sure if I would do the same for all built in edges. As soon as you want to do something fancy, you should use the `BaseEdge` and build your own edge component I would say. This would also make this change smaller since we would only need to update the `BaseEdge` component and not care about the prop drilling of the built in edges. wdyt?

peterkogo:
I removed quite a lot of manual prop drilling from some of the components. Some, like the straight edge, were using `...props` already. 

In reverse this means we have to prop drill for all components manually. Either way is fine for me.

moklick:
The prop drilling was introduced with this PR I think. Current main: https://github.com/xyflow/xyflow/blob/main/packages/react/src/components/Edges/StraightEdge.tsx

peterkogo:
You're right! I must have mistaken the StepEdge for a regular Edge. Will revert it.

peterkogo:
@moklick done, it's better this way.

mhuggins:
Thanks for helping see this through, everyone! 😄

PR #4859, opened 2024-12-01
Add 'group' to `BuiltInNode` type definition

sjdemartini:
As is, the 'group' node type is used in internal utilities for the built-in node types
https://github.com/xyflow/xyflow/blob/616d2665235447e0280368228ac64b987afecba0/packages/react/src/components/NodeWrapper/utils.tsx#L16-L21 

and is indeed a supported built-in node per https://reactflow.dev/learn/layouting/sub-flows and https://reactflow.dev/examples/layout/sub-flows, and simply renders the node without any handles. As such, it should seemingly be part  of the type definition for `BuiltInNode`.

### DIFF
--- a/examples/svelte/src/routes/examples/overview/CustomNode.svelte
+++ b/examples/svelte/src/routes/examples/overview/CustomNode.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
-	import { Handle, Position, type BuiltInNode, type NodeProps } from '@xyflow/svelte';
+	import { Handle, Position, type LabeledBuiltInNode, type NodeProps } from '@xyflow/svelte';
 
-	type $$Props = NodeProps<BuiltInNode>;
+	type $$Props = NodeProps<LabeledBuiltInNode>;
 	$$restProps;
 
 	export let data: { label: string } = { label: 'Node' };
--- a/examples/svelte/src/routes/examples/overview/CustomNodeDragHandle.svelte
+++ b/examples/svelte/src/routes/examples/overview/CustomNodeDragHandle.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
-	import type { BuiltInNode, NodeProps } from '@xyflow/svelte';
+	import type { LabeledBuiltInNode, NodeProps } from '@xyflow/svelte';
 
-	type $$Props = NodeProps<BuiltInNode>;
+	type $$Props = NodeProps<LabeledBuiltInNode>;
 	$$restProps;
 
 	export let data: { label: string } = { label: 'Node' };
--- a/packages/react/src/components/Nodes/DefaultNode.tsx
+++ b/packages/react/src/components/Nodes/DefaultNode.tsx
@@ -1,14 +1,14 @@
 import { Position } from '@xyflow/system';
 
 import { Handle } from '../../components/Handle';
-import type { BuiltInNode, NodeProps } from '../../types/nodes';
+import type { LabeledBuiltInNode, NodeProps } from '../../types/nodes';
 
 export function DefaultNode({
   data,
   isConnectable,
   targetPosition = Position.Top,
   sourcePosition = Position.Bottom,
-}: NodeProps<BuiltInNode>) {
+}: NodeProps<LabeledBuiltInNode>) {
   return (
     <>
       <Handle type="target" position={targetPosition} isConnectable={isConnectable} />
--- a/packages/react/src/components/Nodes/InputNode.tsx
+++ b/packages/react/src/components/Nodes/InputNode.tsx
@@ -1,9 +1,9 @@
 import { Position } from '@xyflow/system';
 
 import { Handle } from '../../components/Handle';
-import type { BuiltInNode, NodeProps } from '../../types/nodes';
+import type { LabeledBuiltInNode, NodeProps } from '../../types/nodes';
 
-export function InputNode({ data, isConnectable, sourcePosition = Position.Bottom }: NodeProps<BuiltInNode>) {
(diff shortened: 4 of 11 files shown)


sjdemartini:
On second look, I guess this type definition should be something like:

```ts
export type LabeledBuiltInNode = Node<{ label: string }, 'input' | 'output' | 'default'>;
export type GroupNode = Node<Record<string, never>, 'group'>;
export type BuiltInNode = LabeledBuiltInNode | GroupNode;
```

since `label` is not used in `GroupNode`:

https://github.com/xyflow/xyflow/blob/616d2665235447e0280368228ac64b987afecba0/packages/react/src/components/Nodes/GroupNode.tsx#L1

https://github.com/xyflow/xyflow/blob/616d2665235447e0280368228ac64b987afecba0/packages/svelte/src/lib/components/nodes/GroupNode.svelte#L2-L5

---

However, I see that internally the `BuiltInNode` type is used in several places to represent the typing for either input, output, or default nodes specifically (despite the more general name)—i.e. a node that accepts a `label` prop. e.g.:

https://github.com/xyflow/xyflow/blob/4a12a9f7818b3d7dd3d3e8aba2804103e0918abb/packages/react/src/components/Nodes/InputNode.tsx#L4-L6

https://github.com/xyflow/xyflow/blob/d4bfea9b27b2e4e357ace7bbed88099a58ca4727/examples/svelte/src/routes/examples/overview/CustomNodeDragHandle.svelte#L1-L8

So I suppose those places would need to switch to referencing a type like the `LabeledBuiltInNode` in my example snippet above. Please let me know what approach is preferred. 

And thanks so much for this fantastic library!

peterkogo:
Like you pointed out the group node does not do anything with a label. In fact it does not do anything at all.

```ts
export function GroupNode() {
  return null;
}
```

Why do you need this change? Adding a node with type "group" just works and there is nothing else to type for the reason above.

sjdemartini:
@peterkogo Thank you for the quick reply. Right, the main motivation is so that consumers of xyflow can import these types and use them in their project to improve static type-checking. In particular for my project, when I define the `const nodes = [...]` array which gets passed to `<ReactFlow nodes={nodes} ... />`, I'd like to specify its stricter type, like:

```ts
import type { BuiltInNode } from "@xyflow/react";

/** The Node<..., ...> types specific to my project, which are added to `<ReactFlow nodeTypes={...} />` */
type CustomNode = MyCustomNode1 | MyCustomNode2 | MyCustomNode3 | GhostNode;
/** All allowed node types, both built-in and custom. */
type Node = BuiltInNode | CustomNode;

const nodes: Node[] = [
  {
    ...
  }
]
```

This typing is super valuable, because it enforces that whenever I define an object in the node array, whether it be a custom node I defined with the `data` I specified is used for that node/component, or a built-in node (`input`, `group`, or otherwise), like:

```ts
  {
    id: '1',
    type: 'input',
    data: { label: 'Node 0' },
    position: { x: 250, y: 5 },
  },
  {
    id: '2',
    type: 'group',
    data: {},
    position: { x: 100, y: 100 },
    style: { width: 200, height: 200 },
  },
  {
    id: '3',
    type: 'myCustomNode1',
    data: {
      fieldA: 42,
      fieldB: 'foo',
    },
    position: { x: 0, y: 0 },
  },
  {
    id: '4',
    type: 'ghost',
    data: {},
    position: { x: 500, y: 500 },
  },
```

my TypeScript type-checking ensures that each object's `data` is correct for the given `type` I specify, and all the `type`s are valid. As such, it protects against mistakenly passing in a `label` for `group` and expecting it to render (even though that prop is not used with `group`), mistakenly _not_ passing in a `label` for other built-in nodes,  making typos in a `data`'s key name or a `type` name, etc. 

Without this PR's change, if I use the code above, I will get a type error since `"group"` is not part of the exported `BuiltInType` definition, and there's no similarly-exported types from xyflow that I can reference:

```
Type '"group"' is not assignable to type '"input" | "output" | "default" | ...
```

I could do this manually with a workaround myself in my project (which is what I'm doing now):

```ts
import type { BuiltInNode } from "@xyflow/react";

type CustomNode = MyCustomNode1 | MyCustomNode2 | MyCustomNode3 | GhostNode;
type GroupNode = Node<Record<string, never>, 'group'>; // Define this in my own project
type BulitInNodeIncludingGroup = BuiltInNode | GroupNode;
type Node = BulitInNodeIncludingGroup | CustomNode;
```

but it seems like xyflow can make this accessible to consumers directly, and most easily via an improvement to the correctness of `BuiltInNode` (since "group" _is_ a built-in node, supported by xyflow, so intuitively to me should be included in that exported type). This seems preferable vs requiring consumers to manually define the `Node` types on their end to account for "group" nodes.

moklick:
Hey @sjdemartini! Thanks for your input here. I regret that we have built-in nodes at all but that's how it is now. I would not distinguish between labeled and non-labeled nodes. I think it's enough to change the `BuiltInNode` type to:

```ts
export type BuiltInNode = Node<{ label: string }, 'input' | 'output' | 'default'> | Node<Record<string, never>, 'group'>;
```

What do you think @peterkogo @sjdemartini?

peterkogo:
I agree. The most minimal change is enough here.

sjdemartini:
Thanks for making the change to include `group` in that exported type! I think it can be handy to distinguish between individual nodes' types for various purposes (e.g., if I want to enforce that a specific function returns only a certain variety of node(s), it can be handy to have a specific type for that, like `GroupNode`), but this is still a useful improvement. 👍

moklick:
Thanks again @sjdemartini! This is fixed in 12.3.6.

PR #4868, opened 2024-12-05
Fix unnecessary rerenders of edges in viewport

### LINKED ISSUE #4867: EdgeRenderer renders everytime when updating viewport dimensions

- React Flow version: 12.3.5
- Browser and version: Google Chrome 131.0.6778.87
- OS and version: Windows 11 Enterprise 22631.4460

### Live code example

_No response_

### Describe the Bug

The **EdgeRenderer** renders everytime when updating viewport height or width, for example by resizing the browser window or enlarging the dev tools. This causes major perfomance issues if the canvas contains a larger amount of custom edges with complex path calculation.

The reason for this is that the **EdgeRenderer** component contains the **width** and **height** states in the ReactFlowState **selector**, although both are not necessary in this component. This means, when one of both is updated, the **EdgeRenderer** and all its children renders...

If both states are not necessary in the **EdgeRenderer**, please remove them: 

image

1. Open any react flow project that contains (custom) edges
3. Open dev tools
4. Enable "**Highlight updates when components render.**" in the **React Developer Tools** extension to visualize rendered components.
5. Resize dev tools
6. You will now see that the edges re-render while resizing dev tools

### Expected behavior

As a user I expect stable perfomance throughout the UX, but instead I'm seeing perfomance decrease when changing the viewport dimensions.

### Screenshots or Videos

In this video you can see that the **EdgeRenderer** and all **CustomEdges** render when the width changes, visualized by an orange border. I used the Chrome extension **React Developer Tools**:

https://github.com/user-attachments/assets/3409369f-ce81-46a2-9b39-40061038fc40

### Additional context

_No response_

peterkogo:
closes #4867.

Remove unused references to viewport dimensions in state selector for EdgeRenderer.

Instead of determining visibleEdgeIds first and then looking up the edge in the edgeWrapper, we just return the visibleEdges right away. Why were we even doing this?

### DIFF
--- a/.changeset/young-insects-cry.md
+++ b/.changeset/young-insects-cry.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
+Prevent unnecessary rerenders of edges.
--- a/packages/react/src/components/EdgeWrapper/index.tsx
+++ b/packages/react/src/components/EdgeWrapper/index.tsx
@@ -17,6 +17,7 @@ import type { Edge, EdgeWrapperProps } from '../../types';
 
 export function EdgeWrapper<EdgeType extends Edge = Edge>({
   id,
+  edge,
   edgesFocusable,
   edgesReconnectable,
   elementsSelectable,
@@ -36,7 +37,6 @@ export function EdgeWrapper<EdgeType extends Edge = Edge>({
   onError,
   disableKeyboardA11y,
 }: EdgeWrapperProps<EdgeType>): JSX.Element | null {
-  let edge = useStore((s) => s.edgeLookup.get(id)!) as EdgeType;
   const defaultEdgeOptions = useStore((s) => s.defaultEdgeOptions);
   edge = defaultEdgeOptions ? { ...defaultEdgeOptions, ...edge } : edge;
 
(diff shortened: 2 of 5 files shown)


bcakmakoglu:
> Instead of determining visibleEdgeIds first and then looking up the edge in the edgeWrapper, we just return the visibleEdges right away. Why were we even doing this?

Doesn't the same concept that is described in the NodeRenderer apply here?
```
// The split of responsibilities between NodeRenderer and
// NodeComponentWrapper may appear weird. However, it’s designed to
// minimize the cost of updates when individual nodes change.
//
// For example, when you’re dragging a single node, that node gets
// updated multiple times per second. If `NodeRenderer` were to update
// every time, it would have to re-run the `nodes.map()` loop every
// time. This gets pricey with hundreds of nodes, especially if every
// loop cycle does more than just rendering a JSX element!
//
// As a result of this choice, we took the following implementation
// decisions:
// - NodeRenderer subscribes *only* to node IDs – and therefore
//   rerender *only* when visible nodes are added or removed.
// - NodeRenderer performs all operations the result of which can be
//   shared between nodes (such as creating the `ResizeObserver`
//   instance, or subscribing to `selector`). This means extra prop
//   drilling into `NodeComponentWrapper`, but it means we need to run
//   these operations only once – instead of once per node.
// - Any operations that you’d normally write inside `nodes.map` are
//   moved into `NodeComponentWrapper`. This ensures they are
//   memorized – so if `NodeRenderer` *has* to rerender, it only
//   needs to regenerate the list of nodes, nothing else.
```

PR #4874, opened 2024-12-09
fix(system): safely handles non-html-element object

crimx:
`event.composedPath` and `event.target` do not always return `HTMLElement` object(as their types suggested).

> [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element), and its children, as well as [Document](https://developer.mozilla.org/en-US/docs/Web/API/Document) and [Window](https://developer.mozilla.org/en-US/docs/Web/API/Window), are the most common event targets

This PR prevents exception from being thrown when the object is not an `HTMLElement` (e.g., a `window`).

### DIFF
--- a/.changeset/brown-pugs-pump.md
+++ b/.changeset/brown-pugs-pump.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/system': patch
+---
+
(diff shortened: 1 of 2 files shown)


moklick:
Thanks @crimx ! This will be part of the next release :)

PR #4878, opened 2024-12-09
Svelte Flow 1.0.0-next

peterkogo:
Svelte Flow 1.0 supports Svelte 5 and drops support for Svelte 4. This migration guide can be used to update from Svelte Flow 0.x to 1.0. 

> [!IMPORTANT]
>  You can already check out Svelte Flow 1.0 (alpha): `npm install @xyflow/svelte@next` - Let us know if you have feedback or find some bugs ✌️ 

### `nodes` & `edges` are now using `$state.raw` instead of `writable`

Svelte 5 introduces runes which are now getting used for nodes and edges. 

**Old API**

```js
const nodes = writable([...]);
const edges = writable([...]);
```

**New API**
```js
let nodes = $state.raw([...]);
let edges = $state.raw([...]);
```

### Updating Nodes & Edges

Previously it was possible to update single node properties. Theoretically, this would also be possible with `$state`, however the [performance implications](https://svelte.dev/playground/e6f804ba6da348bc8b6a0a13c59672cb?version=5.19.0) of this are unfortunately too great, so we opted to using `$state.raw`.

This means that `nodes` and `edges` are to be treated as immutable from now on. If you are making updates manually make sure you:

1. create a new node/edge object, when updating a property.
2. reassign the nodes/edges array (this was technically required before anyway)

### `nodes` & `edges` need to be bound from `<SvelteFlow />`

**Old API**

```js
<SvelteFlow {nodes} {edges} />
```

**New API**

```js
<SvelteFlow bind:nodes bind:edges />
```

### Custom Node & Edge Props

This is by enlarge a general change in Svelte 5 but it does have quite a big impact on typing the props of Custom Nodes & Edges.

**Old API**

```js
// CustomNode.svelte

type $$Props = NodeProps;

export let data: $$Props['data'];
export let position: $$Props['position'];
export let selected: $$Props['selected'];
```

**New API**

```js
let { data, position, selected } : NodeProps = $props();
```

### Hooks

Hooks now return reactive values instead of writables. Because `$state` values cannot be [returned by functions directly](https://svelte.dev/docs/svelte/$state#Passing-state-into-functions) we have to return an object with a `.current` property to keep reactivity. In this regard, we are [following the official trend](https://svelte.dev/docs/svelte/svelte-reactivity#MediaQuery) set by the Svelte library authors. 

**Old API**

```js
const edges = useEdges();
$: console.log(edges);
```

**New API**

```js
const edges = useEdges();
$inspect(edges.current)
```

Note that in case of `useNodes`, `useEdges` and `useViewport` reassignments to `.current` work! 

```js
const nodes = useNodes();

function updateNodes() {
   nodes.current = [...]
}
```

### useSvelteFlow

`useSvelteFlow` provides access functions that operate on the internal state. When using the hook inside descendants of the `<SvelteFlow />` component (probably custom nodes/edges), apart from changes to some of the functions it exports there is nothing you have to change.

```js
// CustomNode.svelte

const { updateNode, screenToFlowPosition } = useSvelteFlow(); //still works
```

However, if you are using the `<SvelteFlowProvider />` to extend the scope of internal store to components outside of the main `<SvelteFlow />` component, you will have to wrap the hook call in a `$derived`.

**Old API**

```js
// Sidebar.svelte

const { addNode, deleteNodes, fitView } = useSvelteFlow();
```

**New API**

```js
// Sidebar.svelte

let { addNode, deleteNodes, fitView } = $derived(useSvelteFlow());
```

### Binding the viewport 

**Old API**

```js
const viewport = writable<Viewport>({ x: 100, y: 100, zoom: 1.25 });

<SvelteFlow {viewport} />
```

**New API**

```js
let viewport = $state<Viewport>({ x: 100, y: 100, zoom: 1.25 });

<SvelteFlow bind:viewport />
```

### Custom Connection Line

Using a custom Connection Line was possible before by passing it to a slot. In Svelte Flow 1.0 we introduced a `connectionLineComponent` prop for this:

**Old API**

```js
<SvelteFlow {nodes} {edges}>
  <ConnectionLine slot="connectionLine" />
  <Background variant={BackgroundVariant.Lines} />
</SvelteFlow>
```

**New API**

```js
<SvelteFlow {nodes} {edges} connectionLineComponent={ConnectionLine}>
  <Background variant={BackgroundVariant.Lines} />
</SvelteFlow>
```

### `onedgecreate` -> `onbeforeconnect`

We renamed `onedgecreate` to `onbeforeconnect` so it better aligns with onconnectstart, onconnected and onconnect. You can return an edge with a custom id from the function before a new edge is created via connecting two handles

### colorModeSSR
You can pass a fallback color mode for server side rendering when `colorMode` is set to `'system'`.

### elevateNodesOnSelect
Control if nodes should be elevated via `z-index` when selected.

### clickConnect
You can now connect two handles by clicking on them one after the other. This is also useful if you want to connect two handles programmatically by calling the [click methods](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click) of the handles.

### Reconnecting Edges
You can now mount `<EdgeReconnectAnchor />` components on your edges where users can reconnect an already connected edge.

### Z-Index works across viewport
You can now order everything freely via z-index in the viewport. Edges, nodes and edge labels are all in the same stacking context now.

## Blast radius
Only 2 non-destructive changes have been made to the `system` library that make internalNodes immutable. Svelte Flow 1.0 requires also the nodes in the nodeLookup to be recreated when something changes.

### Improvements to onlyRenderVisibleElements
We have to also render nodes that belong to a visible edge. Otherwise it can happen that edges cannot be rendered because the node that is offscreen is not rendered. We can solve this by unifying the step that calculates the visible nodes and edges.

### `<EdgeReconnectAnchor />`
We should rethink how this is done in React as well, because the contained approach from Svelte Flow is much more flexible.

### Z-Index
Render everything on the viewport without containers, so all elements get sorted into the same stacking context.

## Remaining steps

- [x] Check useNodesData
- [x] EdgeMarkers not mounting
- [x] Decide fate of defaultNodeProps
- [x] Port useNodeConnections
- [x] Release 1.0.0-next 
- [x] Write migration guide
- [x] Convert all Website examples and check migration guide in the process
- [x] Check current docs (improving the docs will be an ongoing task from 1.0.0-next until real release)
- [x] clsx
- [x] HTMLAttributes to main div
- [x] Should we create a type for `.current`? [Related Issue
](https://github.com/sveltejs/svelte/issues/15075)
- [x] Implement new fitView changes
- [x] How to handle Initialized?
- [x] onConnect  #4989
- [x] useNodeConnection: onConnect, onDisconnect
- [x] Click connect
- [x] Reconnect Edge
- [x] https://github.com/xyflow/xyflow/issues/3738
- [x] Expose `HTMLAttributes` as props wherever possible
- [x] https://github.com/xyflow/xyflow/pull/5151
- [x] use .container class instead of component styles
- [x] https://github.com/xyflow/xyflow/issues/5011
- [x] https://github.com/xyflow/xyflow/issues/5203
- [x] Key navigation https://github.com/xyflow/xyflow/issues/3798
- [x] improve setNodes, setEdges experience - especially with hooks
- [x] fix tests
- [x] reset store
- [x] Typing
- [x] https://github.com/xyflow/vite-svelte-flow-template/issues/42
- [x] Aria
- [x] https://github.com/xyflow/xyflow/issues/4077
- [x] z-index
- [x] Fix BackgroundVariants
- [x] noPanClassName, noDragClassName, noWheelClassName
- [x] fix minimap: zoomable, pannable
- [x] fix switching handle positions


dberlin:
@peterkogo FYI - on this branch,  defaultEdgeOptions and nodeTypes got dropped from SvelteFlow.svelte's explicit destructure'd props, along with a few other options.

While they are properly passed down to the store through the ...rest's that exist, and thus get used properly, being missing from the explicit destructured props causes them to render on the toplevel div because the div renders the rest props as attributes:
```html
<div
  bind:this={domNode}
  bind:clientWidth
  bind:clientHeight
  style:width
  style:height
  {style}
  class={cc(['svelte-flow', className, store.colorMode])}
  data-testid="svelte-flow__wrapper"
  role="application"
  {...props}
>
```

So you get something like this in the output

```html
<div style="" class="svelte-flow dark s-6lmXlvjJHzBC"
 data-testid="svelte-flow__wrapper"
 role="application" 
 defaultedgeoptions="[object Object]"
 colormode="dark"
 nodetypes="[object Object]"
 fitview="true"
 autopanonconnect="true"
 autopanonnodedrag="true">
```

You can see that it appears colorMode, fitView, autoPanOnConnect, and autoPanOnNodeDrag have the same issue.

Happy to send a PR if you tell me how you'd like it fixed  (IE remove from rest props if they get passed to store/etc, or add them back to explicit destructuring, or whatever)

peterkogo:
@dberlin good catch! Will take a look at it.

peterkogo:
@dberlin fixed in `@xyflow/svelte@1.0.0-next.3`, thanks again for raising the issue

dberlin:
Awesome, thank you!

dberlin (commented):
So i ran into this as well, and I am kind of curious why bind the store instead of just using useStore everywhere and not passing the store around?
(It looked like a bunch of components/hooks already do useStore anyway, hence the question)

peterkogo:
@dberlin useStore also supports accessing the store outside of the main `<SvelteFlow />` component via the `<SvelteFlowProvider />. For that purpose some additional checking logic is implemented and it uses a `$derived` store instance.

This way we can skip that whole pathway for all internal components.

PR #4880, opened 2024-12-10
Fix type check for all event targets

crimx:
Sorry I missed one part in the previous PR #4874.

https://github.com/xyflow/xyflow/blob/0f21ec21ebf01567ec1de4fad1d0c55f40db3463/packages/system/src/utils/dom.ts#L46

This PR adds type checks for all `event.target`.


moklick:
Thanks @crimx!

PR #4890, opened 2024-12-15
Pass through NodeType to connectionLineComponent ReactFlow prop

solastley:
## Summary
The `connectionLineComponent` attribute in `ReactFlowProps` is of type `ConnectionLineComponent` which is a generic type that takes a `NodeType` as a type argument.

The `NodeType` of the `ReactFlowProps` is not getting passed through to the `ConnectionLineComponent`, though. This PR fixes that.

## Testing
Successfully ran `pnpm build` and `pnpm dev` locally without errors. Also ran `pnpm test:react` and all tests pass except one (but I think it may be flakey - see below).

Small note for the xyflow team: I ran into a few issues running `pnpm test:react`. First, I had trouble running the tests at all because the correct playwright binaries were not installed by default. To install the correct binaries I had to do:

```
cd tests/playwright
pnpm exec playwright install --with-deps
```

I think this is possibly because the version of playwright is different in the root `package.json` than it is in the `package.json` within `tests/playwright`.

The other issue was that it seems like there is a flakey or bad unit test:

```
  ✘  75 [firefox] › pane.spec.ts:87:9 › Pane default › autoPan › autoPanOnNodeDrag (1.9s)
```

I can't seem to get this test to pass locally, but it seems definitely unrelated to my changes.

### DIFF
--- a/.changeset/happy-hats-lay.md
+++ b/.changeset/happy-hats-lay.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': minor
+---
+
+Pass `NodeType` type argument from `ReactFlowProps` to `connectionLineComponent` property.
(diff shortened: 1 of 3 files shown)


moklick:
Thanks for your help. we will look into it!

moklick:
Thanks @solastley !

Issue #4891, opened 2024-12-15
Keyboard stopped working in SvelteFlow


JDR42:
### What platform were you using when you found the bug?

- React Flow / Svelte Flow version: SvelteFlow
- Browser and version: Chrome & Firefox
- OS and version: Win11

### Live code example

https://stackblitz.com/edit/h9trywep?file=index.html

### Describe the Bug

Keyboard stopped working in SvelteFlow.

Ex:
* Backspace will not delete selected node
* Cannot use CTRL modifier to mutliselect

### Steps to reproduce the bug or issue

1. Go here: https://svelteflow.dev/examples/nodes/update-node
2. Select an edge and hit the `Backspace` key.  Try to multiselect with CTRL. Note all keyboard behavior is working fine here.
3. Open in StackBlitz with the link in the lower-right
4. Try step 2 here and it fails.

### Expected behavior

As a user, I expected to get keyboard functionality in the stackblitz demo as in the website demo.

peterkogo:
I can confirm this is happening. This seems to be caused by one of our dependencies.

JDR42:
@nicbaz Works great. You're amazing! Cheers

peterkogo:
This was fixed in Svelte Flow 0.1.26.

PR #4894, opened 2024-12-17
fix(svelte): temporary fix for keyboard shortcuts

nicbaz:
Exclude problematic version `3.1.2` of `@svelte-put/shortcut` which causes `Shift` and `Backspace` keys to stop working in svelteflow.

fixes: #4891


peterkogo:
Sorry, I saw this PR too late. Thanks for the help, this is fixed in the newest version.

PR #4910, opened 2024-12-26
chore(deps): lower React peer dependency requirements to >=16.12.0

yizhihuamao:
Practical testing shows that React version 16.12.0 is fully compatible as well. As an open-source project, we should be more lenient with peerDependencies, considering many legacy projects face difficulties with upgrading.


peterkogo:
Unfortunately, we do not have the capacity to extend the support to React 16.

PR #4937, opened 2025-01-11
expose ControlButton props to consumers

jrmoynihan:
I'm guessing the props accidentally had the export keywords removed.  This probably isn't caught by Typescript or CI because the props are passed in by a spread prop from the parent `Controls.svelte`.

Should just be a band-aid fix until the Svelte 5 release is ready to use the `$props` rune.


peterkogo:
Thanks for the PR. This is indeed an oversight!

PR #4957, opened 2025-01-20
Enhance/node props

peterkogo:
We currently pass undefined to the node for selected, selectable, draggable, deletable. This causes some weird issues down the line like in the [NodeResizer](https://reactflow.dev/examples/nodes/node-resizer) (See "NodeResizer when selected" node shows resizer when selected is undefined because the `NodeResizer` has default value for `isVisible = true`.
Additionally this also complicates things for the user, even though we do know precisely if a node is selected, selectable, draggable or deletable.

I also noticed the current Svelte Flow version is super messy in this regard - I fixed what I could but in the new version everything is so much better!

### DIFF
--- a/.changeset/rotten-moons-roll.md
+++ b/.changeset/rotten-moons-roll.md
@@ -1,0 +1,7 @@
+---
+'@xyflow/react': minor
+'@xyflow/svelte': minor
+'@xyflow/system': patch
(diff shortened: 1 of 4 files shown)


moklick:
good catch. Can you make it a patch release?

Issue #4976, opened 2025-01-24
Batching/etc in useReactFlow seems to interact badly with updateNodeInternals


dberlin:
### What platform were you using when you found the bug?

Reactflow latest

### Describe the Bug

I am using elk to do layout, have multiple handles, and sort the handles into the order that elk suggests in order to minimize edge crossings.
The state manager is a standard zustand store.
After laying out the nodes, and calling set nodes, i go through and call updateNodeInternals on every single node id since we moved the nodes around and the connection lines need to be redrawn.

The code looks like this:
``` if (nodesInitialized) {
        const layoutNodes = async () => {
            const allNodes = getNodes();
            const allEdges = getEdges()
            const {layoutedNodes} = await getLayoutedNodes(allNodes, allEdges);
            setNodes(layoutedNodes);
            for (const node of allNodes) {
                   updateNodeInternals(node.id)
            }
            setTimeout(() => fitView(), 0);
        };

        layoutNodes().catch(console.error);
```

If i use `setNodes`/`getNodes` from `useReactFlow`, about 1 in 3 times the connection lines will not rerender.
The component will rerender, move to the correct position, labels are correct, order is correct post-rerender, but the connection lines do not get redrawn and seem to be in a weird state (clicking them makes them disappear).
I attached a picture of what you get in the broken rendering picture.  In this picture, the connection lines have not been updated despite updateNodeInternals on every node being called as above (As mentioned, the edge array is correct and the edge is from water to water, iron ore to iron ore, etc)

I have verified that no ids have changed (keys in rendering are order dependent on purpose to force re-rerender when this occurs, but handle ids are stable, so no edges or port ids actually change).
I have verified all edges are correct, all nodes are correct, etc.  
I have verified that outside of the literal drawing of connection lines from/to the right places, everything is correct :)

What appears to be happening is that the batch store update that useReactFlow uses under the covers does not get applied all the time prior to the updateNodeInternals callback happening.    Obviously, updateNodeInternals uses a callback to try to ensure the timing of node internals updates happens later.  But it tries to use `requestAnimationFrame` as the callback trigger to do the  internals update before the next repaint, and there is no guarantee i can find that says this will occur *after* the batch store update has been applied (which uses useeffect).

[...]

PR #4982, opened 2025-01-26
Change hook used by updateNodeInternals to guarantee ordering vs batched store updates

dberlin:
Fixes Issue #4976
Under the covers, useReactFlow does batched store updates using batch contexts.
These are applied through a useIsomorphmicLayoutEffect callback.

updateNodeInternals also uses a callback to apply updates.  However, it uses
requestAnimationFrame to trigger the callback, which is a browser callback
unrelated to React, and has no ordering guarantees vs the batched store
updates.

The result is that whether the callback created by updateNodeInternals sees the
latest state or not currently depends on your current monitor refresh rate and
graphics card rendering speed, and other amusing things like whether the window
is covered or not.

This commit changes the hook used for callbacks by updateNodeInternals to be
the same as the hook used to apply batched store updates.  React guarantees
these will happen in the order you set them up, which in turn guarantees the
callback will happen after the batch update completes.

### DIFF
--- a/packages/react/src/hooks/useUpdateNodeInternals.ts
+++ b/packages/react/src/hooks/useUpdateNodeInternals.ts
@@ -2,6 +2,7 @@ import { useCallback } from 'react';
 import type { UpdateNodeInternals, InternalNodeUpdate } from '@xyflow/system';
 
 import { useStoreApi } from '../hooks/useStore';
+import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect';
 
 /**
  * Hook for updating node internals.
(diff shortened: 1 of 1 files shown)


dberlin:
I'll note - this approach will break anyone using updateNodeInternals outside of a component.  I'm not sure that is supposed to be legal in the first place, and none of the testsuite fails, but ...

moklick:
Thanks for your PR! You are calling `useEffect` inside the `useCallback` function. Are you sure that you wanted to suggest this?

dberlin:
No - I'm hoping if i do it wrong,  someone who knows the codebase will suggest the right way to fix it :)
Please feel free to suggest a better idea.

The code there right now is essentially equivalent to doing what this makes explicit.  
It simply doesn't get flagged by hook checkers/etc because it uses a browser callback instead of a react one.  But it has all the same practical problems as if it was a react one, as the issue shows.

PR #4990, opened 2025-01-31
fix(system): ensure auto-pan stops when state changes during drag

damianstasik:
This PR ensures that auto-panning is disabled if the `autoPanOnNodeDrag` state changes while dragging on the canvas. Currently, if this property is set to `true` before the drag event starts, the auto-panning logic will begin but won't stop until dragging ends.

In my use case, I'd like to stop auto-panning if I detect that a user has dragged a node over a specific area near the canvas edge.

Feel free to close this PR if this behavior is expected.

### DIFF
--- a/.changeset/serious-ducks-care.md
+++ b/.changeset/serious-ducks-care.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/system': patch
+---
+
+Make it possible to stop autoPanOnDrag by setting it to false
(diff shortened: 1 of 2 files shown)


moklick:
Hey @damianstasik . Thanks for your PR :) I think this is a good improvement! What do you think about calling the drag end handler too when you return early inside the autoPan function?

moklick:
I checked again and it wouldn't make sense to call the drag end handler programmatically because it gets called after user finishes the drag .. This will be published with the next release. Thanks @damianstasik

Issue #3483, opened 2023-10-05
Background and panel positions are not updated when tabbing to off-screen node


ahmacleod:
### Describe the Bug

Hi all, big fan of ReactFlow here. I noticed this bug this morning when working on some accessibility features and thought I should flag:

When navigating the graph with the tab button, if the focus shifts to a node that is off screen, the background and panels are not re-rendered in the correct new position. This can be seen on the website examples.

Thank you!

### Steps to Reproduce the Bug or Issue

1. Open https://reactflow.dev/docs/examples/overview/
2. Zoom in enough for some nodes to be off-screen
3. Click a node to provide focus
4. Tab until a node that is off-screen becomes focused
5. The background and panels will be in the wrong position

### Expected behavior

Background and panels should remain stationary throughout.

### Screenshots or Videos

https://github.com/wbkd/react-flow/assets/2594635/3535ae30-2943-43b9-971c-d4ee1a1c71cc

### Platform

- OS: macOS
- Browser: Chrome
- Version: 117.0.5938.92
- `reactflow`: 11.9.2

moklick:
It's not easy to fix, because it's a native browser behaviour but we will have a look for v12.

moklick:
A solution could be to set `.react-flow__pane` to `position: fixed`. We will investigate.

peterkogo:
Also happens when input fields extend off-screen #3598 #3354

PR #4991, opened 2025-02-01
Fix viewport shifting on node focus

### LINKED ISSUE #4667: Programmatic focusing [el.focus()] of a node breaks the viewport

- Browser and version: 
- OS and version:

### Live code example

_No response_

### Describe the Bug

The title is pretty self explanatory. The issue can be seen on reactflow.dev's homepage. At high zoom, using `tab` or `element.focus()` will disrupt the regular behavior of the viewport. 

https://github.com/user-attachments/assets/db5b27e6-97b4-44a0-94a8-06b6ecd69de0

1. Goto https://reactflow.dev/
2. Max zoom on the demo nodes 
3. Select a node and press tab multiple times

### Expected behavior

Viewport should adjust based on the focused node.

### Screenshots or Videos

https://github.com/user-attachments/assets/1285280f-74c5-443f-9531-787db7cf08d5

### Additional context

_No response_

waynetee:
Fixes #4667, #3483

When a node outside the viewport is focused (e.g. via tab or when typing in a previously selected input field), the browser automatically scrolls `rf_wrapper` to bring the node into view, breaking the viewport.

This PR undos that behavior, keeping the viewport stable.

### DIFF
--- a/.changeset/cyan-pianos-lie.md
+++ b/.changeset/cyan-pianos-lie.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
+Fix viewport shifting on node focus
--- a/.changeset/sour-jobs-fold.md
+++ b/.changeset/sour-jobs-fold.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
+Prevent viewport shift after using Tab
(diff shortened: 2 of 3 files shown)


moklick:
wow, thanks! I can't believe this actually works. 

We need to check if this could have any unwanted side effects for example on mobile or with `preventScroll={false}`. @peterkogo @bcakmakoglu what do you think about this?

moklick:
I can't see any downsides of this approach. We can still have scrollable elements inside custom nodes and preventScroll also still works. Thanks for your help here @waynetee !

PR #5002, opened 2025-02-08
types/nodes: Use NodeTypeString for string-extending generics

behnam-oneschema:
This should make it easier to differentiate generic parameters which take a `NodeBase`-/`Node`-based type vs those that take a `string`-based one.

### DIFF
--- a/packages/system/src/types/nodes.ts
+++ b/packages/system/src/types/nodes.ts
@@ -9,7 +9,7 @@ import { Optional } from '../utils/types';
  */
 export type NodeBase<
   NodeData extends Record<string, unknown> = Record<string, unknown>,
-  NodeType extends string = string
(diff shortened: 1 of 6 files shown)


packages/system/src/types/nodes.ts
```diff
@@ -9,7 +9,7 @@ import { Optional } from '../utils/types';
  */
 export type NodeBase<
   NodeData extends Record<string, unknown> = Record<string, unknown>,
-  NodeType extends string = string
+  NodeTypeString extends string = string
```
behnam-oneschema, line 12: Not sure if `NodeTypeString` is better than `NodeStringType`. 🤷

moklick:
We decided to keep the name since we couldn't find a better one for the generic.

PR #5014, opened 2025-02-12
fix(NodeResizer): handle undefined dimensions with default values

alvaromorais:
This attempts to fix the problem in which there sometimes the nodes have `position.x` and `position.y` as `NaN`

### DIFF
--- a/.changeset/weak-olives-confess.md
+++ b/.changeset/weak-olives-confess.md
@@ -1,0 +1,6 @@
+---
+'@xyflow/react': patch
+'@xyflow/system': patch
+---
+
+fix(NodeResizer): handle undefined dimensions with default values
--- a/packages/react/src/additional-components/NodeResizer/NodeResizeControl.tsx
+++ b/packages/react/src/additional-components/NodeResizer/NodeResizeControl.tsx
@@ -74,8 +74,8 @@ function ResizeControl({
 
           if (node && node.expandParent && node.parentId) {
             const origin = node.origin ?? nodeOrigin;
-            const width = change.width ?? node.measured.width!;
-            const height = change.height ?? node.measured.height!;
+            const width = change.width ?? node.measured.width ?? 0;
+            const height = change.height ?? node.measured.height ?? 0;
 
             const child: ParentExpandChild = {
               id: node.id,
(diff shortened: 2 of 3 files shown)


moklick:
Why does this happen? Is it possible that the issue occurs because of a broken `onNodesChange` function?

alvaromorais:
> Why does this happen? Is it possible that the issue occurs because of a broken

I believe this is when what you call a broken `onNodesChange` without using `applyNodeChanges`.
We are trying to upgrade from react 11 to 12 and our app doesn't use a internal reactflow node state. 
Instead is a direct transpilation of another state with a different coordinate system to react flow nodes and edges. 
Changing this would be a significant refactor on our part, so we are trying to solve the issue. 
This actually solves our problem. In the code the fact that `!` is currently being used suggests you don't expect `undefined`. However this is not the case. Indeed it returns `node.measured.width` as `undefined`.

moklick:
can't you apply changes and then do your custom stuff like:

```js
const onNodesChange = useCallback((changes) => {
  setNodes((nds) => {
    const nodesWithChanges = applyNodeChanges(changes, nds);
    const nextNodes = doCustomStuff(nodesWithChanges);

    return nextNodes;
  })
}, []);
```

moklick:
Got it! I will discuss with the team what we can do about it. Maybe we can show a proper warning if users work with nodes that are not initialized correctly and then get rid of those `!` here.

moklick:
I need some hints how I can reproduce this. Even if I do not implement an `onNodesChanges` and try to drag a node I don't get this error.

alvaromorais:
> I need some hints how I can reproduce this. Even if I do not implement an `onNodesChanges` and try to drag a node I don't get this error.

Here. If you don't use applyNodeChanges you will see the drag event NaN. 

https://codesandbox.io/p/sandbox/wonderful-currying-ww9nly?file=%2FApp.js%3A20%2C9-20%2C19

moklick:
this is "fixed" in 12.4.4. You will receive a warning, because you are working with uninitialized nodes. Doing this comes affects your performance, because React Flow will re-measure your nodes over and over again. I would re-consider to implement onNodesChanges as explained.

alvaromorais:
> this is "fixed" in 12.4.4. You will receive a warning, because you are working with uninitialized nodes. Doing this comes affects your performance, because React Flow will re-measure your nodes over and over again. I would re-consider to implement onNodesChanges as explained.

Thank you. We will refactor eventually. This will fix the issue for now for us.

PR #5033, opened 2025-02-20
lint: use `React.JSX` type instead of the deprecated global `JSX` namespace

dimaMachina:


moklick:
good catch! Thanks :)

PR #5048, opened 2025-02-23
lint: refactor from legacy eslint config to flat eslint config

dimaMachina:
also bumped eslint to the latest and use `eslint.config.ts`


moklick:
superseded by #5762

PR #5055, opened 2025-02-25
Always allow releasing keys

darabos:
Wow you have some complex keypress handling code! 😅 

### What happened to me

I bring up a node search bar when `/` is pressed. The XYFlow keydown handler first registers that `/` has been pressed. Then we jump to the node search bar. Then the XYFlow keyup handler runs. The keyup handler sees that we are now inside an input, so it returns early. Leaving `/` forever pressed in the eyes of XYFlow.

### About the fix

This is a niche case and I'm probably the only one hurt. But the fix is so simple. Why would we ignore a key getting released because of being inside an input or whatever the modifier keys are doing? If it's released it's released.

Thanks for considering this!

### DIFF
--- a/packages/react/src/hooks/useKeyPress.ts
+++ b/packages/react/src/hooks/useKeyPress.ts
@@ -106,13 +106,6 @@ export function useKeyPress(
       };
 
       const upHandler = (event: KeyboardEvent) => {
-        const preventAction =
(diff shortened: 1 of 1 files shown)


moklick:
haha you found the part of the library that I like the least. It's a shame that there is no native API for shortcuts :( however, thanks for your PR. We will have a look!

moklick:
Thanks again @darabos !

PR #5059, opened 2025-02-28
fix(react,svelte): prevent pane click when connection is in progress

bcakmakoglu:
# Closes

- [5057](https://github.com/xyflow/xyflow/issues/5057)


moklick:
thanks @bcakmakoglu !

PR #5065, opened 2025-03-03
fix(svelte): inverted attribution condition

SIMULATAN:
The condition was accidentally inverted in https://github.com/xyflow/xyflow/commit/28e4fad342845a05beac9bcc77eb8a5a14727806#diff-88df6697ff34fa2b48872b0220ca1fb63f6e912ec5fdc76c5f45e18fe4c36075


peterkogo:
Thanks, this is absolutely right!

PR #5067, opened 2025-03-05
Enhance fitView

peterkogo:
## Removed
- `fitViewOnInit`, `fitViewOnInitDone` and `fitViewInitOptions` from store
- `fitView` and `fitViewSync` from store actions
- Handling initial `fitView` in updateNodeInternals
- `fitView` from `useViewportHelper`

## Added
- `fitViewQueued`, `fitViewOptions` and `fitViewResolver` to store
- `adoptUserNodes` now returns nodesInitialized
- `fitViewQueued` is now handled directly after `adoptUserNodes`
- `fitView` from `useReactFlow` now only sets `fitViewQueued` in the store and triggers `setNodes` without any changes via the `batchProvider` to trigger adoptUserNodes
- Promise resolution now works via `fitViewResolver`, which is a [`Promise.withResolvers`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/withResolvers). Even though this is a Baseline 2024 feature we should check if there are any issues with the Polyfill

## Fixes
Adding new nodes and calling fitView right after, will now execute the fitView correctly. Additionally, the first rendered frame with the new nodes has the correctly fitted view already applied.

## New Features
You can now define paddings inside the fitViewOptions as pixel values ('30px'), as viewport percentages ('20%') and you can define different paddings for each side `{ x, y, top, bottom, left, right }` (top, bottom, left, right takes precedent over x and y)

### DIFF
--- a/packages/react/src/store/index.ts
+++ b/packages/react/src/store/index.ts
@@ -64,7 +64,6 @@ const createStore = ({
         fitViewResolver?.resolve(value);
         set({ fitViewResolver: null });
       });
-      set({ nodes, fitViewQueued: false, fitViewOptions: undefined });
     }
 
     return {
@@ -89,6 +88,7 @@ const createStore = ({
 
         if (fitViewQueued && nodesInitialized) {
           resolveFitView();
+          set({ nodes, fitViewQueued: false, fitViewOptions: undefined });
         } else {
           set({ nodes });
         }
@@ -138,6 +138,7 @@ const createStore = ({
 
         if (fitViewQueued) {
           resolveFitView();
+          set({ fitViewQueued: false, fitViewOptions: undefined });
         } else {
           // we always want to trigger useStore calls whenever updateNodeInternals is called
           set({});
--- a/packages/react/src/store/index.ts
+++ b/packages/react/src/store/index.ts
@@ -62,10 +62,9 @@ const createStore = ({
           checkEquality: true,
         });
 
-        let viewportFitted = false;
         if (fitViewQueued && nodesInitialized && panZoom) {
-          const { fitViewOptions, width, height, minZoom, maxZoom } = get();
-          viewportFitted = fitViewport(
+          const { fitViewOptions, fitViewResolver, width, height, minZoom, maxZoom } = get();
+          const fitViewPromise = fitViewport(
             {
               nodes: nodeLookup,
               width,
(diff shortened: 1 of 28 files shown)


packages/react/src/store/index.ts
```diff
@@ -42,11 +41,35 @@ const createStore = ({
   nodeOrigin?: NodeOrigin;
   nodeExtent?: CoordinateExtent;
 }) =>
-  createWithEqualityFn<ReactFlowState>(
-    (set, get) => ({
+  createWithEqualityFn<ReactFlowState>((set, get) => {
```
moklick, line 44: can we use async here?

```ts
async function resolveFitView() {
      const { nodeLookup, panZoom, fitViewOptions, fitViewResolver, width, height, minZoom, maxZoom } = get();

      if (!panZoom || !fitViewResolver) {
        return;
      }

      await fitViewport(
        {
          nodes: nodeLookup,
          width,
          height,
          panZoom,
          minZoom,
          maxZoom,
        },
        fitViewOptions
      );

      fitViewResolver.resolve(true);
      set({ fitViewResolver: null });
    }
```

moklick (changes requested):

packages/svelte/src/lib/store/index.ts
```diff
@@ -156,52 +144,18 @@ export function createStore({
...
-        height: get(store.height),
-        minZoom: get(store.minZoom),
-        maxZoom: get(store.maxZoom),
-        panZoom
-      },
-      options
-    );
-
-    return fitViewNodes.size > 0;
+    console.log('fitView Store');
```
moklick, line 147: could you remove this?

moklick (changes requested):

packages/svelte/src/lib/store/utils.ts
```diff
@@ -148,14 +164,38 @@ export const createNodesStore = (
...
-    adoptUserNodes(nds, nodeLookup, parentLookup, {
+    const nodesInitialized = adoptUserNodes(nds, nodeLookup, parentLookup, {
       elevateNodesOnSelect,
       nodeOrigin,
       nodeExtent,
       defaults,
       checkEquality: false
     });
 
+    console.log(nodesInitialized);
```
moklick, line 175: and this :)

moklick (changes requested):

packages/svelte/src/lib/store/utils.ts
```diff
@@ -148,14 +164,38 @@ export const createNodesStore = (
...
+
+    if (get(fitViewQueued) && nodesInitialized && get(panZoom)) {
+      console.log('trying');
```
moklick, line 178: 👆

moklick (changes requested):

PR #5102, opened 2025-03-21
feat(system): adds generic node/edge typings to xyhandle

thedanchez:
## Summary

This PR adds generic typings to `XYHandle`.

## Comments

Let me know if this is undesirable. First time making a PR against the repository and not trying to introduce any breaking changes 🙂

PR #5105, opened 2025-03-21
feat(system): updates generic typings for xydrag & xyhandle

thedanchez:
## Summary

Similar to https://github.com/xyflow/xyflow/pull/5102 -- this PR updates the existing generic typing for XYDrag as I saw there were gaps in it. This one could perhaps be more sensitive than the other change as it was more additive in nature compared to this one. I'm uncertain about the sensitivity of the change, but just wanted to make sure that the `NodeType`/`EdgeType` generics get passed throughout the system's type definitions.

Feel free to poke holes in this PR or reject it all together. Hoping it can be helpful 🙂


packages/system/src/xydrag/XYDrag.ts
```diff
@@ -29,17 +29,24 @@ import type {
...
-  nodes: NodeBase[]
+  node: NodeType,
+  nodes: NodeType[]
 ) => void;
 
-type StoreItems<OnNodeDrag> = {
-  nodes: NodeBase[];
-  nodeLookup: Map<string, InternalNodeBase>;
-  edges: EdgeBase[];
+type OnNodeDrag<NodeType extends NodeBase = NodeBase> = (e: MouseEvent | TouchEvent, node: NodeType, nodes: NodeType[]) => void | undefined;
```
thedanchez, line 39: This is what I converted the original `OnDrag` type generic to that was originally defined using `any`.

packages/system/src/xydrag/XYDrag.ts
```diff
@@ -87,14 +94,13 @@ export type DragUpdateParams = {
   nodeClickDistance?: number;
 };
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function XYDrag<OnNodeDrag extends (e: any, nodes: any, node: any) => void | undefined>({
+export function XYDrag<NodeType extends NodeBase = NodeBase>({
```
moklick, line 97: It would be better to pass EdgeType here too, so that we can pass it to XYDragParams instead of having to type `getStoreItems` directly.

packages/system/src/xydrag/XYDrag.ts
moklick: Interesting. It seems there is a mismatch with the current OnNodeDrag type that is defined in the react package. There the `event` is just a MouseEvent. That could be a bug in the library 🤔

moklick (changes requested):

packages/system/src/xydrag/XYDrag.ts
```diff
@@ -87,14 +94,13 @@ export type DragUpdateParams = {
   nodeClickDistance?: number;
 };
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function XYDrag<OnNodeDrag extends (e: any, nodes: any, node: any) => void | undefined>({
```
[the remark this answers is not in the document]
moklick: The system package is not documented and I assume that no one besides us is using it currently, so this should be fine. For a React Flow user this should not have any effect.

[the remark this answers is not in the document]
thedanchez: Yeah, I think this would be helpful to others who want to build on the underlying system to build Flow wrappers for other libraries. Would be good for system to have a good generic foundation 🙂

packages/system/src/xydrag/XYDrag.ts
thedanchez: Good call -- I pushed up the changes

packages/system/src/xydrag/XYDrag.ts
```diff
@@ -52,23 +59,23 @@ type StoreItems<OnNodeDrag> = {
...
-  onSelectionDrag?: OnSelectionDrag;
-  onSelectionDragStop?: OnSelectionDrag;
-  updateNodePositions: UpdateNodePositions;
+  onNodeDragStart?: OnNodeDrag<NodeType>;
+  onNodeDrag?: OnNodeDrag<NodeType>;
+  onNodeDragStop?: OnNodeDrag<NodeType>;
+  onSelectionDragStart?: OnSelectionDrag<NodeType>;
+  onSelectionDrag?: OnSelectionDrag<NodeType>;
+  onSelectionDragStop?: OnSelectionDrag<NodeType>;
+  updateNodePositions: UpdateNodePositions<InternalNodeBase<NodeType>>;
```
thedanchez, line 73: Hopefully this wont be a problem as `UpdateNodePositions` accepts a generic type based on `InternalNodeBase` and this time we're passing the `NodeType` info through along with it.

packages/system/src/xydrag/XYDrag.ts
```diff
@@ -87,14 +94,13 @@ export type DragUpdateParams = {
...
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export function XYDrag<OnNodeDrag extends (e: any, nodes: any, node: any) => void | undefined>({
+export function XYDrag<NodeType extends NodeBase = NodeBase, EdgeType extends EdgeBase = EdgeBase>({
   onNodeMouseDown,
   getStoreItems,
   onDragStart,
   onDrag,
   onDragStop,
-}: XYDragParams<OnNodeDrag>): XYDragInstance {
+}: XYDragParams<NodeType, EdgeType>): XYDragInstance {
```
thedanchez, line 106: Now we're passing both NodeType + EdgeType to `XYDragParams` which will type the underlying store getter as well.

peterkogo:
@thedanchez just took us a little over a year, would you look at that! 🫣 

But this actually uncovered a wrong type, thank you!

peterkogo (approved):

peterkogo:
This PR now also includes the changes from #5102. Merge conflicts were a little bit severe for my taste and it's easier to check everything this way.

moklick:
Nice! Could you add a changeset for the system package too @peterkogo and check the other changesets?

peterkogo:
@moklick Done, its ready to merge from my side

PR #5115, opened 2025-03-25
fix: improve TSDoc comments for `EdgeLabelOptions` and `EdgeTextProps`

dimaMachina:


packages/react/src/components/Edges/EdgeText.tsx
```diff
@@ -8,9 +8,9 @@ function EdgeTextComponent({
   x,
   y,
   label,
-  labelStyle = {},
+  labelStyle,
   labelShowBg = true,
-  labelBgStyle = {},
```
dimaMachina, line 13: is ok to left them as `undefined`

packages/react/src/components/Edges/EdgeText.tsx
```diff
@@ -34,7 +34,7 @@ function EdgeTextComponent({
     }
   }, [label]);
 
-  if (typeof label === 'undefined' || !label) {
+  if (!label) {
```
dimaMachina, line 37: can be simplified

packages/react/src/components/Edges/EdgeText.tsx
moklick: 👍

PR #5124, opened 2025-03-29
chore: add missing type re-export for NodeConnection

bjornosal:
Missing the NodeConnection type export for the new hook `useNodeConnections`. 

I mainly use React, and doing guesswork on the Svelte part.


moklick:
Thanks @bjornosal !

PR #5142, opened 2025-04-01
fix: improve TSDoc comments for HandleProps, NodeBase and `InternalNodeBase`

dimaMachina:


packages/system/src/types/nodes.ts
```diff
@@ -71,7 +81,7 @@ export type NodeBase<
   };
 };
 
-export type InternalNodeBase<NodeType extends NodeBase = NodeBase> = NodeType & {
+export type InternalNodeBase<NodeType extends NodeBase = NodeBase> = Omit<NodeType, 'measured'> & {
```
dimaMachina, line 84: I omitted, because we respecify below

packages/system/src/types/nodes.ts
```diff
@@ -99,11 +109,11 @@ export type NodeProps<NodeType extends NodeBase> = Pick<
...
   Required<Pick<NodeType, 'type' | 'dragging' | 'zIndex' | 'selectable' | 'deletable' | 'selected' | 'draggable'>> & {
-    /** whether a node is connectable or not */
+    /** Whether a node is connectable or not. */
     isConnectable: boolean;
-    /** position absolute x value */
+    /** Position absolute x value. */
     positionAbsoluteX: number;
-    /** position absolute x value */
+    /** Position absolute y value. */
     positionAbsoluteY: number;
```
dimaMachina, line 117: typo :)

PR #5143, opened 2025-04-01
fix: improve TSDoc comments for `ReactFlowInstance` and `GeneralHelpers`

dimaMachina:


packages/react/src/types/instance.ts
```diff
@@ -241,6 +252,10 @@ export type ReactFlowInstance<NodeType extends Node = Node, EdgeType extends Edg
   NodeType,
   EdgeType
 > &
-  Omit<ViewportHelperFunctions, 'initialized'> & {
+   ViewportHelperFunctions & {
```
dimaMachina, line 255: there is no `initialized` on  `ViewportHelperFunctions` type

moklick:
good findings again!

PR #5151, opened 2025-04-04
fix: minimize inline style usage

leejuyuu:
In some environments, inline styles are not permitted due to security issues (e.g. strict CSP). Some of the current SvelteFlow components set string values to `style` attribute to apply dynamic styles, which makes part of the functionalities not available in these environments.

This commit tries to minimize usage of inline styles by

- Use `style:` directive where possible. These are compiled to setting properties of the CSSOM, which is permitted by CSP.
- For props that can be set to `style` attribute, default to `undefined` instead of empty string to avoid unnecessarily setting the attribute.

### DIFF
--- a/packages/svelte/src/lib/components/ConnectionLine/ConnectionLine.svelte
+++ b/packages/svelte/src/lib/components/ConnectionLine/ConnectionLine.svelte
@@ -10,8 +10,8 @@
     getStraightPath
   } from '@xyflow/system';
 
-  export let containerStyle: string = '';
-  export let style: string = '';
+  export let containerStyle: string | undefined = undefined;
+  export let style: string | undefined = undefined;
   export let isCustomComponent: boolean = false;
 
   const { width, height, connection, connectionLineType } = useStore();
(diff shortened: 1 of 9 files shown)


moklick:
Thanks! @peterkogo is working on the Svelte Flow 1.0 release and will have a look.

leejuyuu:
Thanks for looking into this.

Some organizations require CSP to be set. You could try setting `Content-Security-Policy: style-src 'self'; ` when serving a SvelteFlow built app. Some functionalities, such as dragging or zooming the viewport, would break.

For the width and height, could it be done with something like this? If all of the inline-related props are not set, then we skip setting the style attribute.

```js
let nodeStyle = $derived.by(()=>{
  const w = measuredWidth === undefined ? (width ?? initialWidth) : width
  const h = measuredHeight === undefined ? (height ?? initialHeight) : height
  if (w === undefined && h === undefined && style === undefined) {
    return undefined
  }
  return `${style};width:${toPxString(w)};height:${toPxString(h)}`
})
```

leejuyuu:
Thanks! Looking forward to the 1.0 release.

peterkogo:
This has landed in Svelte Flow 1.0 🚀 Take a look at our [migration guide](https://svelteflow.dev/learn/troubleshooting/migrate-to-v1) for more infos.

PR #5153, opened 2025-04-04
When horizontal, remove the border from all control buttons

nick-lawrence-ctm:
I found a small styling issue while creating a Reactflow component in our application. See discussion: https://github.com/xyflow/xyflow/discussions/5152

With the proposed change, horizontal buttons would have no bottom border by default.
image

closes #5152


moklick:
Thanks @nick-lawrence-ctm! Good catch :)

PR #5154, opened 2025-04-05
Fixed onNodesChange docstring in component-props.ts

ibagov:
Removed all mentions of "edges" in the docstring of onNodesChange and replaced them with mentions of "nodes"


moklick:
Thanks @ibagov !

PR #5160, opened 2025-04-05
Improve TSDoc comments for `type UseHandleConnectionsParams` and `useHandleConnections` hook

dimaMachina:


packages/react/src/hooks/useHandleConnections.ts
```diff
@@ -10,11 +10,16 @@ import {
 import { useStore } from './useStore';
 import { useNodeId } from '../contexts/NodeIdContext';
 
-type useHandleConnectionsParams = {
+type UseHandleConnectionsParams = {
```
dimaMachina, line 13: type should be starts with upper case

PR #5161, opened 2025-04-05
Improve TSDoc comments for `type UseKeyPressOptions` and `useKeyPress` hook

dimaMachina:


packages/react/src/hooks/useKeyPress.ts
```diff
@@ -6,7 +6,16 @@ type PressedKeys = Set<string>;
 type KeyOrCode = 'key' | 'code';
 
 export type UseKeyPressOptions = {
+  /**
+   * You may want to listen to key presses on a specific element. This field lets you configure
+   * that!
```
dimaMachina, line 11: copied from https://reactflow.dev/api-reference/hooks/use-key-press#options-target

I think we can simplify?

```suggestion
   * Listen to key presses on a specific element.
```

packages/react/src/hooks/useKeyPress.ts
```diff
@@ -40,11 +47,17 @@ const defaultDoc = typeof document !== 'undefined' ? document : null;
...
+   * The key code (string or array of strings) specifies which key(s) should trigger
+   * an action.
+   *
+   * A **string** can represent:
+   * - A **single key**, e.g. `'a'`
+   * - A **key combination**, using `'+'` to separate keys, e.g. `'a+d'`
+   *
+   * An  **array of strings** represents **multiple possible key inputs**. For example, `['a', 'd+s']`
+   * means the user can press either the single key `'a'` or the combination of `'d'` and `'s'`.
+   * @default null
```
dimaMachina, line 60: The original description with website's description are different https://reactflow.dev/api-reference/hooks/use-key-press#key-code

I used ChatGPT to generate new from both versions, I think this is pretty good now?

PR #5169, opened 2025-04-06
Improve TSDoc comments for `type ReconnectEdgeOptions` and `getViewportForBounds`

dimaMachina:


packages/system/src/utils/general.ts
```diff
@@ -275,16 +275,16 @@ function calculateAppliedPaddings(bounds: Rect, x: number, y: number, zoom: numb
 }
 
 /**
- * Returns a viewport that encloses the given bounds with optional padding.
+ * Returns a viewport that encloses the given bounds with padding.
```
dimaMachina, line 278: padding is not optional 
image

PR #5170, opened 2025-04-06
Improve TSDoc comments for `interface GetSimpleBezierPathParams` and `getSimpleBezierPath`

dimaMachina:


packages/react/src/components/Edges/SimpleBezierEdge.tsx
```diff
@@ -85,8 +95,8 @@ function createSimpleBezierEdge(params: { isInternal: boolean }) {
       sourceY,
       targetX,
       targetY,
-      sourcePosition = Position.Bottom,
-      targetPosition = Position.Top,
+      sourcePosition,
+      targetPosition,
```
dimaMachina, line 99: these optional values are redundant because we provide them in

https://github.com/dimaMachina/xyflow/blob/eb2a33c6dfb0629e851ab3a0f2cd70eca92efc42/packages/react/src/components/Edges/SimpleBezierEdge.tsx#L47-L53

PR #5172, opened 2025-04-06
Improve TSDoc comments for `useNodesData`, `useReactFlow`, `isNode` and `isEdge`

dimaMachina:


packages/react/src/hooks/useReactFlow.ts
```diff
@@ -31,8 +31,6 @@ const selector = (s: ReactFlowState) => !!s.panZoom;
  * This hook returns a ReactFlowInstance that can be used to update nodes and edges, manipulate the viewport, or query the current state of the flow.
  *
  * @public
- * @returns ReactFlowInstance
```
dimaMachina, line 34: redundant - we already know from type definition that return type is `ReactFlowInstance`

PR #5271, opened 2025-05-16
refactor(svelte): change a11y inline styles to classes

leejuyuu:
In some environments, inline styles are not permitted due to security issues (e.g. strict CSP). The a11y messages will be visible in these environments due to browsers blocking inline styles.

Change the inline styles to use classes instead.


moklick:
@peterkogo wdyt? seems reasonable to me.

peterkogo (approved):

PR #5274, opened 2025-05-19
docs(CONTRIBUTING): update links

printerscanner:
Reviewing the contributing guide, a few of the links were outdated. Links going to the old wbkd url were redirecting, but I updated those as well.


moklick:
good finding! thanks

PR #5277, opened 2025-05-19
chore(a11y): add static support for a11y message localization

printerscanner:


packages/react/src/components/NodeWrapper/index.tsx
```diff
@@ -143,9 +144,12 @@ export function NodeWrapper<NodeType extends Node>({
...
       store.setState({
-        ariaLiveMessage: `Moved selected node ${event.key
-          .replace('Arrow', '')
-          .toLowerCase()}. New position, x: ${~~internals.positionAbsolute.x}, y: ${~~internals.positionAbsolute.y}`,
+        ariaLiveMessage:
+          labelConfig['a11yDescription.ariaLiveMessage']?.(
+            event.key.replace('Arrow', '').toLowerCase(),
+            ~~internals.positionAbsolute.x,
+            ~~internals.positionAbsolute.y
+          ) || '',
```
moklick, line 152: to avoid this, the store labelConfig should have a slightly different type (all fields required)

packages/react/src/types/component-props.ts
```diff
@@ -666,4 +667,9 @@ export interface ReactFlowProps<NodeType extends Node = Node, EdgeType extends E
    * @default false
    */
   debug?: boolean;
+  /**
+   * Custom accessibility messages for screen readers and a11y features.
```
moklick, line 671: could you generalize this comment a bit more?

moklick (changes requested):

packages/system/src/constants.ts
```diff
@@ -36,3 +36,29 @@ export const infiniteExtent: CoordinateExtent = [
 ];
 
 export const elementSelectionKeys = ['Enter', ' ', 'Escape'];
+
+export const defaultLabelConfig = {
```
moklick: I just talked with @peterkogo and he suggested to rename it to "ariaLabelConfig" . We wanted to keep it more flexible in case we would like to add other labels but right now I can think of any other ones. 

Let's rename it and also rename the keys with "...title" to "...ariaRole" (controls.zoomin.title -> controls.zoomin.ariaLabel for example). Let's also rename all keys that start with "a11yDescription" (a11yDescription.node.default -> node.a11Description.default for example). For the users it's easier to understand I think.

moklick (changes requested):

packages/system/src/constants.ts
printerscanner: so defaultLabelConfig becomes `ariaLabelConfig` or should it be: 

LabelConfig -> AriaLabelConfig
defaultLabelConfig -> defaultAriaLabelConfig

moklick: LabelConfig -> AriaLabelConfig
defaultLabelConfig -> defaultAriaLabelConfig

this one!

moklick (approved):

moklick:
Awesome :)

PR #5299, opened 2025-05-28
feat: add `ariaRole` prop to nodes

printerscanner:


moklick:
Awesome! Could you rename `role` to `ariaRole`? It's not that I like it more but we already have `ariaLabel` so it's more consistent.

moklick:
This looks great :) Do you think we could do the same thing for edges in this PR?

packages/system/src/types/edges.ts
```diff
@@ -40,6 +40,17 @@ export type EdgeBase<
...
   interactionWidth?: number;
+  /**
+   * The ARIA role attribute for the edge, used for accessibility.
+   * @default "group"
+   */
+
+  ariaRole?: AriaRole;
+  /**
+   * A description of the edge's, used for accessibility.
+   * @default "node"
```
moklick: this should be "edge"

packages/system/src/types/utils.ts
```diff
@@ -56,3 +56,8 @@ export type Transform = [number, number, number];
  * to represent an unbounded extent.
  */
 export type CoordinateExtent = [[number, number], [number, number]];
+
+/**
+ * The `AriaRole` type is used to define the role of an element in the accessibility tree.
+ */
+export type AriaRole = 'button' | 'group' | 'listitem' | 'application' | 'region' | 'none' | null;
```
moklick: I am not sure about svelte, but react exports an `AriaRole` type that we could use for this.

packages/react/src/components/NodeWrapper/index.tsx
```diff
@@ -192,7 +191,8 @@ export function NodeWrapper<NodeType extends Node>({
       onDoubleClick={onDoubleClickHandler}
       onKeyDown={isFocusable ? onKeyDown : undefined}
       tabIndex={isFocusable ? 0 : undefined}
-      role={isFocusable ? 'button' : undefined}
+      role={node.ariaRole === null ? undefined : node.ariaRole || (isFocusable ? 'group' : undefined)}
```
moklick: do we still need this null check?

moklick (approved):

moklick:
Thanks! This is great :)

PR #5308, opened 2025-06-02
feat(NodeWrapper): focus nodes in the viewport on tab

printerscanner:


packages/react/src/components/NodeWrapper/index.tsx
```diff
@@ -155,6 +155,23 @@ export function NodeWrapper<NodeType extends Node>({
...
+    const zoom = panZoom?.getViewport().zoom ?? 1;
+    panZoom?.setViewport(
+      {
+        x: -(internals.positionAbsolute.x + nodeDimensions.width / 2) * zoom + window.innerWidth / 2,
+        y: -(internals.positionAbsolute.y + nodeDimensions.height / 2) * zoom + window.innerHeight / 2,
+        zoom: zoom,
+      },
+      { duration: 100 }
+    );
+  };
```
bcakmakoglu: Maybe I'm overlooking sth but couldn't one just call `fitView({ nodes: [{ id: nodeId }], duration: 100, minZoom: zoom, maxZoom: zoom })`? 🤔

printerscanner: Ah you might be right. I had a feeling I was overcomplicating it.

moklick:
this is such a cool feature 🤩 maybe we need a prop to turn it off?

packages/react/src/components/NodeWrapper/index.tsx
```diff
@@ -155,6 +155,23 @@ export function NodeWrapper<NodeType extends Node>({
     }
   };
 
+  const onFocus = () => {
+    if (disableKeyboardA11y) {
+      return;
```
moklick, line 171: We should also return early if the user clicks a node (because we only want to focus on tab). For this you could use:

`nodeRef.current?.matches(':focus-visible')`

moklick (changes requested):

packages/react/src/components/NodeWrapper/index.tsx
moklick: if it matches :focus-visible, we should also check if the node is already inside the viewport and only fit the view if that's not the case.

printerscanner:
@moklick could you take a look?

PR #5311, opened 2025-06-03
update readme team and website

omesome:
@moklick do we even need the team section in the readme? 
Creates another place where it has to be updated, and is kind of redundant, as you can see the team on the [xyflow](https://github.com/xyflow) org page?


moklick:
Good catch but there is already an open PR for this https://github.com/xyflow/xyflow/pull/5297

PR #5312, opened 2025-06-03
perf: improve pane dragging render performance

dxsun97:
Currently, when dragging/clicking a pane, the pane element will add a `dragging` class to change the cursor style property from grab to grabbing. As the number of nodes (in other words: the amount of DOM) increases, the time it takes to recalculate the style increases, especially on Chromium-based browsers. There is a known chromium issue here that reveals that modifying the cursor style can cause this problem: https://issues.chromium.org/issues/40493007

So I applied this style change to a single div to avoid recalculating styles for a large number of elements.

This might help resolve the issue: https://github.com/xyflow/xyflow/issues/4527

### DIFF
--- a/packages/react/src/container/Pane/index.tsx
+++ b/packages/react/src/container/Pane/index.tsx
@@ -265,7 +265,7 @@ export function Pane({
 
   return (
     <div
-      className={cc(['react-flow__pane', { draggable, dragging, selection: isSelecting }])}
+      className={cc(['react-flow__pane', { selection: isSelecting }])}
       onClick={hasActiveSelection ? undefined : wrapHandler(onClick, container)}
       onContextMenu={wrapHandler(onContextMenu, container)}
       onWheel={wrapHandler(onWheel, container)}
@@ -278,6 +278,7 @@ export function Pane({
       style={containerStyle}
     >
       {children}
+      {draggable && <div className={cc(['react-flow__drag', 'react-flow__container', { dragging }])}></div>}
       <UserSelection />
     </div>
   );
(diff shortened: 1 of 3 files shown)


moklick:
Thanks! This is interesting and we will have a look!

dxsun97:
> Thanks! This is interesting and we will have a look!

Hello, how is it going? I have a project that is troubled by this problem. I hope to solve this problem as soon as possible.

moklick:
Unfortunately this would be a breaking change. Some users might use the classes to change the cursor or do other things. We will have another look this week!

dxsun97:
> Unfortunately this would be a breaking change. Some users might use the classes to change the cursor or do other things. We will have another look this week!

Hi, I found another way to solve this problem: instead of changing the DOM, provide a new prop: paneClassName to override the className of the Pane component, which can avoid breaking changes. What do you think?

moklick:
I am not sure if I understand that solution. How would it solve the issue? Can't you overwrite the existing styles if you don't want to change the cursor?

dxsun97:
> I am not sure if I understand that solution. How would it solve the issue? Can't you overwrite the existing styles if you don't want to change the cursor?

Sorry, I provided some misleading information. In fact, the style recalculation here is because the class attribute of the pane element is changed, and the descendant elements need to be recalculated. Even if its style is overwritten to keep it consistent, the style recalculation still occurs. At present, it seems that the only way to avoid it is to keep the class of the pane element unchanged. This is what I actually did in my PR

moklick:
Thanks again! I put it on the React Flow 13 list since it's a breaking change.

peterkogo:
Hey, thanks for your contribution. We found a bit simpler solution for this but your optimization will be part of our next major release!

Closing in favor of #6029

PR #5316, opened 2025-06-05
Use CSS Variables for viewport & CSS only viewport invariance

peterkogo:


peterkogo:
@moklick We need to decide if we want to still apply the olds cool transform string instead of using the css variables for the viewport itself. If we don't then I go ahead and fix the failing tests.

peterkogo:
This is unfortunately not feasible for performance reasons...

PR #5321, opened 2025-06-06
Fix bug where non-child node positions aren't clamped

### LINKED ISSUE #5320: Node: position doesn't clamp within extents

- React Flow / Svelte Flow version: `12.3.0` - preset

### Live code example
https://codesandbox.io/p/sandbox/heuristic-tharp-fpzgsz

Note: to view the actual vs. expected position, open the console.

### Describe the Bug

As of `@xyflow/react@12.3.0` non-child node positions (i.e., `node.parentId === undefined`) aren't clamped when a node's position is calculated. This causes `getNode("<node-id>").position` to be set as the last position of the node was held instead of the actual position of the node, bound by the node's extent. This doesn't affect the UI, since a node's internal absolute position isn't affected; however, this affects applications that depend on the position to be bound within the extents.

1. Create a node with extents
2. Drag the node past left-x extent
  - Node should be stopped at the extent
3. View the position compared to the absolute position.

### Expected behavior

As a user, you would expect `getNode("<node-id>").position` to contain the position of the node on the plane _not_ the position where the node was last held. 

> Position of a node on the pane.
Source: https://reactflow.dev/api-reference/types/node#position

### Screenshots or Videos

_No response_

### Additional context

- This bug is related to `calculateNodePosition` in `@xyflow/system`
- Introduced in https://github.com/xyflow/xyflow/pull/4572

dylanmiddendorf:
Closes #5320


dylanmiddendorf:
@peterkogo @moklick

peterkogo:
Hi @dylanmiddendorf, nice find! Thanks for the PR.

moklick:
Good catch! But the solution seems a bit more complicated than this. We will have a look.

dylanmiddendorf:
@moklick

Could you clarify what the proposed solution is not addressing? This PR focuses on clamping node positions to their extents within `calculateNodePosition`, aligning behavior with both the documentation and user expectations. Let me know if there's anything I can do to help expedite the review.

moklick:
The issue is that this would overwrite the adjusted extent that we are calculating if a selection is being dragged https://github.com/xyflow/xyflow/blob/f0ce2c876d8688e13632bc86286cf857f86dead6/packages/system/src/xydrag/XYDrag.ts/#L173-L177

dylanmiddendorf:
Thanks for the input. However, it seems unrelated due to the mutually exclusive conditions. Specifically, `adjustedNodeExtent` only diverges from the flow extent when the node has no defined extent, as shown on line 165:

```ts
if (dragItems.size > 1 && nodeExtent && !dragItem.extent)
```

In `graph.ts`, the node’s extent is only used if it's explicitly defined; otherwise, it defaults to the provided parameter. Since one case requires an undefined extent and the other requires a defined one, they don’t conflict.

moklick:
Yep, you are right! Thanks for your help here 👍

PR #5325, opened 2025-06-10
fix(a11y): remove outdated `ariaRoleDescription` field

printerscanner:
Removes the `ariaRoleDescription` field we caught earlier. It's already removed everywhere else.


moklick:
👍

PR #5326, opened 2025-06-10
correct node resizer control size for viewport zoom

peterkogo:


packages/system/src/styles/node-resizer.css
```diff
@@ -28,12 +28,12 @@
...
 /* handle styles */
 .xy-flow__resize-control.handle {
-  width: 4px;
-  height: 4px;
+  width: 5px;
+  height: 5px;
   border: 1px solid #fff;
   border-radius: 1px;
   background-color: var(--xy-resize-background-color, var(--xy-resize-background-color-default));
-  transform: translate(-50%, -50%);
```
moklick, line 36: Let's keep this in case someone overwrites it.

moklick: ok, we need to keep it like this

PR #5332, opened 2025-06-11
docs(CONTRIBUTING): add changeset style guide

printerscanner:
Added only one introductory sentence compared to the original in Notion.


moklick:
Good idea 👌

PR #5339, opened 2025-06-13
fix(svelte): move events not firing

jrmajor:


peterkogo:
Good catch! Thanks for the PR!

Issue #5315, opened 2025-06-05
Edge: clicking reconnect region affects selection and rendering


ntkibm:
### What platform were you using when you found the bug?

- React Flow version: 12.6.4
- Browser and version: Chrome 135.0.7049.85
- OS and version: Mac Sonoma 14.4

### Describe the Bug

On single clicking the reconnect region of an edge, if the edge was previously selected, its `selected` property changes to `false` for a split second and then back to `true`. Also, you can notice a flicker/flash effect on the edge label on clicking the reconnect region.

https://github.com/user-attachments/assets/8169ddff-4073-4226-955c-ec21628af455

### Steps to reproduce the bug or issue

This can be reproduced on the example page https://reactflow.dev/examples/edges/reconnect-edge

### Expected behavior

If the edge was previously selected, it's value should just stay `true` on clicking the reconnect region, and the flickering effect shouldn't occur on the label.

peterkogo:
It seems like we need to adjust for the click threshold on reconnect, too. Thanks for the heads up!

corbinklett:
related: single-clicking the reconnect area forces the edge to disconnect from the original handle in the "ghost node" implementation: https://reactflow.dev/examples/edges/temporary-edges

This means that when a user simply clicks on the edge near the connection, the connection comes undone unintentionally. 

The user should have to start dragging in order to reconnect events to occur. I'm trying to figure out how to make this happen, such as by toggling the edge reconnectable prop, but if that is false then the reconnectable circle element disappears.

moklick:
A fix will be published with the next minor release. For this we added a new `connectionDragThreshold` that lets you configure when a connection should start.

corbinklett:
After trying version 12.8.0 of xyflow, the problem still persists. 

After setting a `connectionDragThreshold`, the `onReconnectStart` function is no longer called when I click on the reconnect area of an edge. However, the `onReconnectEnd` and `onConnectEnd` functions are still called. 

The edge still "jumps" out of place when you click on the reconnect area. Such behavior should only happen after the user drags beyond the threshold - not on a click. You can see the jumping behavior by clicking on the reconnect area in the example: https://reactflow.dev/examples/edges/temporary-edges

moklick:
Thanks @corbinklett this is fixed in 12.8.1 - however I am not sure how we can fix the jumping. Please open a new issue for this, since the original issue is closed with the latest version.

corbinklett:
[...]

Issue #4217, opened 2024-04-29
ConnectionDragThreshold


peterkogo:
Implement a connectionDragThreshold, so onClick and onDrag events can be distinguished better.

moklick:
Related: we added a `paneClickDistance` in 12.0.0 which lets you configure the max distance between mousedown/up that will trigger a click.

c0d3ster:
@moklick is the `paneClickDistance` prop on ReactFlow instance the one that should be used if we want to still trigger clicks when users accidentally make a small drag movement? We are experiencing an issue with our users where they click but it is not treated as a click on inner elements within a node because of a slight drag movement.
We are on `"@xyflow/react": "^12.0.0-next.16",` using Typescript, but that prop is not being recognized as a valid prop option for some reason.

c0d3ster:
I see `nodeDragThreshold` as a prop option, but unfortunately that still doesn't allow it to trigger the `onClick` of an inner element whenever a small drag movement is made under the threshold.

moklick:
You need to update to 12.0.3 to use `paneClickDistance`. If you have interactive elements inside your nodes, you might want to use the class name "nodrag" in order to prevent dragging at all.

c0d3ster:
ah okay I thought the 12.0.0 "next" build was newer than the 12.0.3. The problem with nodrag is that we do in fact want the whole node to be draggable, but we would just like to have it trigger normal click events when the drag amount is below the drag threshold.

c0d3ster:
Updating and adding `paneClickDistance` did not trigger either a `onNodeClick` action or the native `onClick` when doing a slight drag movement below the specified distance. Am I missing something here?

c0d3ster:
https://github.com/xyflow/xyflow/issues/2341 seems related but still not clear to me how we can achieve the native `onClick` event to trigger for small drag movements.

moklick:
In v12.0.2 we added `nodeClickDistance`. Does that solve your issue @c0d3ster ?

c0d3ster:
`nodeClickDistance` resolved our issue, but it is not yet in the documentation. This combined with `nodeDragThreshold` set to the same pixel value (10 in our case) allows small drag movements to be treated as clicks still and ignores any custom drag behavior below that threshold 👍

moklick:
That's great to hear @c0d3ster . I just added the missing docs!

moklick:
We added a `connectionDragThreshold` prop in 12.8.0

PR #5344, opened 2025-06-16
Add connectionDragThreshold

moklick:
#5315 #4217

### DIFF
--- a/.changeset/twelve-llamas-sort.md
+++ b/.changeset/twelve-llamas-sort.md
@@ -1,0 +1,7 @@
+---
+'@xyflow/react': minor
+'@xyflow/svelte': minor
+'@xyflow/system': minor
+---
+
+Add connectionDragThreshold prop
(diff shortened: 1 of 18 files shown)


moklick:
I will keep this open until the end of the dev cycle in case there will be more minor release changes.

TrueBoxGuy:
@moklick I'm not sure how many people have moved to 12.8 yet, but there is potentially an argument that a small change needs to be made before incorporating this into a minor release. 

From my understanding, to get equivalent behaviour to versions pre-12.8, `connectionDragThreshold` should be 0 (but it is currently 1). 

Although not documented, the behaviour that clicking on handles causes an `onConnectStart` event is in the public API. It is also a fairly significant behaviour. So I feel like making `connectionDragThreshold` 1 classifies as a breaking change and shouldn't be done until the next major release (according to semantic versioning).

moklick:
Hey @TrueBoxGuy thanks for your feedback here. We decided to change the behaviour slightly because we thought that this is the better default but I will have another look if we keep it like this.

PR #5345, opened 2025-06-17
chore(react): add batching for default nodes

### LINKED ISSUE #4593: V12 migration: Node jitter when snapping to Helper Lines in uncontrolled flow

- React Flow / Svelte Flow version: 12.1.1
- Browser and version:  Brave Browser (1.68.141)
- OS and version: macOS Sonoma (14.6.1)

### Live code example

https://codesandbox.io/p/github/philipburleigh/reactFlowBug/main?import=true

### Describe the Bug

I am in the midst of migrating from v11 to v12, but I am struggling with a jittering  node, when it should snap to the helper lines. My flow is uncontrolled, which forces me to manipulate my nodes through a ReactFlowInstance. Using ReactFlowInstance.setNodes allowed me to snap the node without any jitter in v11, but this is not the case in v12.

1. Go to the codesandbox (turn on the server if needed)
2. Drag a node around, and notice how it jitters when snapping to the helper lines.

### Expected behavior

A clean node snap and no jitter.

### Screenshots or Videos

_No response_

### Additional context

_No response_

moklick:
closes #4593

### DIFF
--- a/examples/react/src/examples/Stress/index.tsx
+++ b/examples/react/src/examples/Stress/index.tsx
@@ -221,10 +221,10 @@ const StressFlow = () => {
   return (
     <ReactFlow
       key={key}
-      nodes={nodes}
+      defaultNodes={initialNodes}
       edges={edges}
       onConnect={onConnect}
-      onNodesChange={onNodesChange}
+      // onNodesChange={onNodesChange}
       onEdgesChange={onEdgeChange}
       minZoom={0.2}
       fitView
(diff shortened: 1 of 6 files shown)


moklick:
If we decide to handle default nodes like this, we should drop the BatchProvider and all the queue stuff to the store. We could initialize the store with the correct queue, but it's currently not possible because the BatchProvider depends on the store.

PR #5346, opened 2025-06-17
Add `group` classname to node wrapper

0x0f0f0f:


0x0f0f0f:
cc @moklick

moklick:
This looks good! Could you add this for edges and handles and Svelte Flow too?

moklick:
This PR is obsolete because we found a way to style nodes based on the selected class name without using groups: https://play.tailwindcss.com/1pmCeLiCIX

PR #5359, opened 2025-06-24
set devEngines in package.json

omesome:
https://github.com/xyflow/internal/issues/81

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,15 +35,15 @@
     "turbo": "^2.0.3",
     "typescript": "5.4.5"
   },
-  "packageManager": "pnpm@9.2.0",
-    "devEngines": {
+  "packageManager": "pnpm@10.12.4",
+  "devEngines": {
     "runtime": {
       "name": "node",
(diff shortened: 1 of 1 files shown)


moklick:
If we add this, I think the "packageManager" field can be removed?

package.json
```diff
@@ -35,5 +35,15 @@
     "turbo": "^2.0.3",
     "typescript": "5.4.5"
   },
-  "packageManager": "pnpm@9.2.0"
+  "packageManager": "pnpm@9.2.0",
```
moklick: If we add this, I think the "packageManager" field can be removed?

moklick (changes requested):

package.json
omesome: Ok :)

omesome: I checked this again and we actually should keep both of these fields. 
updated to the latest `pnpm` version though.

omesome:
seems like this is causing problems in deployment, so closing

Issue #5360, opened 2025-06-24
Edges are rendered below group nodes


chrtze:
### What platform were you using when you found the bug?

- React Flow / Svelte Flow version: latest 
- Browser and version: Chrome, latest
- OS and version: MacOS latest

### Live code example

https://reactflow.dev/examples/grouping/sub-flows

### Describe the Bug

When rendering a sub flow without any further customization, the edges that are connected to the child nodes are rendered behind the group node.

### Steps to reproduce the bug or issue

1. Go to https://reactflow.dev/examples/grouping/sub-flows
2. Set a non-transparent background for one of the group nodes
3. You can see that the edges are behind that group node

### Expected behavior

The edges should be rendered on the same level as the nodes they are connected to.

### Screenshots or Videos

<img width="877" alt="Image" src="https://github.com/user-attachments/assets/9facf441-e93f-404b-acd5-a54247fa30db" />

moklick:
We changed the default behaviour in 12.8.0. From now on edges that are connected to a node with a parent are rendered above nodes.

PR #5361, opened 2025-06-24
Fix/edge elevation

peterkogo:
potential fix for #5360

### Behavior before
- `elevateEdgesOnSelect` controls both if an edge should be elevated on select AND when either of the connected nodes is selected
- `zIndex` of child nodes is the same as the parent
- When child nodes gets elevated, the edges do not change their zIndex

### Behavior after
- `elevateEdgesOnSelect` only controls if an edge should be elevated when it's selected
- `zIndex` of child nodes is `parent.zIndex + 1`
-  Edges receive the the `zIndex` of the connected nodes

### Question
Should we add a new prop that controls whether the edges automatically receive the zIndex of their parent? If you don't want the edges to ever be above the nodes this is important.

### DIFF
--- a/.changeset/real-ties-travel.md
+++ b/.changeset/real-ties-travel.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/system': patch
+---
+
+Render edges above nodes when they are within a subflow


chrtze:
Is it possible that whenever a zIndex is passed to the edge, it should overwrite the new default behaviour?

peterkogo:
Here is an updated solution. I incorporated the suggestion from @chrtze 

### New behavior
- `zIndex` of child nodes is `parent.zIndex + 1`
- `elevateEdgesOnSelect` only controls if an edge should be elevated when it's selected
-  Edges receive the the `zIndex` of the connected nodes when they have parentNodes
- passing zIndex to edge trumps all of the above

XiaoRongwen:
什么时候发新的npm包呢12.8.0

PR #5365, opened 2025-06-25
Fix/initial fit view

### LINKED ISSUE #5352: Fit view with a flow without nodes

If you use the `fitView` prop on a flow without nodes, it gets executed when a node is added to the flow. 

### Live code example

_No response_

### Describe the Bug

.

.

### Expected behavior

.

### Screenshots or Videos

_No response_

### Additional context

_No response_

peterkogo:
Closes  #5352


moklick:
let's keep the current behaviour and have another look in the next cycle. It seems that the behaviour is a bit different for Svelte Flow and React Flow in the current version and we should decide which one we prefer.

PR #5368, opened 2025-06-26
chore(store-updater): cleanup

moklick:


moklick:
thanks for the hint @peterkogo

PR #5376, opened 2025-07-03
feat(smoothstep-edge): add bendPosition parameter to control bend location

kennyjwilli:
- Added bendPosition parameter to getSmoothStepPath function to control where the bend occurs along smoothstep edges
- Updated React and Svelte SmoothStepEdge components to pass through the new parameter
- Added examples with different bend positions (0.2, 0.5, 0.8)

 The bendPosition parameter accepts values from 0 to 1. e.g., 
- 0 = bend at source position
- 1 = bend at target position
- 0.5 = bend at midpoint (default behavior)

This provides more granular control over smoothstep edge routing, especially useful for layouts where the default midpoint bend may not be desired.

**Example Screenshot**

Screenshot 2025-07-02 at 8 15 40 PM

### DIFF
--- a/examples/react/src/examples/EdgeRouting/index.tsx
+++ b/examples/react/src/examples/EdgeRouting/index.tsx
@@ -69,6 +69,55 @@ const nodes: Node[] = [
     targetPosition: Position.Bottom,
     style: { background: 'rgba(255,255,255,0.5)' },
   },
+  // Bend Position Examples
+  {
+    id: '9',
+    position: { x: 300, y: 0 },
+    data: { label: 'Source' },
+    sourcePosition: Position.Right,
+    targetPosition: Position.Right,
+    style: { background: 'rgba(255,255,255,0.5)' },
+  },
+  {
+    id: '10',
+    position: { x: 600, y: 150 },
+    data: { label: 'Target' },
+    sourcePosition: Position.Left,
+    targetPosition: Position.Left,
+    style: { background: 'rgba(255,255,255,0.5)' },
+  },
+  {
+    id: '11',
+    position: { x: 300, y: 300 },
+    data: { label: 'Source' },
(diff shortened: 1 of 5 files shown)


moklick:
Hey Kenny, this is a great addition! I would just like to think about the naming before this gets merged.

moklick:
@kennyjwilli I renamed bendPosition to stepPosition. Another thing that I can't get my head around is the issue that currently only one step position is getting passed which is used for x and y. That's why the edge labels in your screenshots also change their y position, even if they should stay in the center of the step. Do you have an idea for a solution for this? I could imagine to pass an array `stepPostion={[0.2, 0.5]}` but I am not sure if that would fix it 🤔

kennyjwilli:
@moklick Great call out. I think the issue is that the current implementation applies stepPosition as a linear interpolation between source and target for both X and Y coordinates, which causes the bend point (and consequently the label) to move diagonally. We can instead have the stepPosition only control where the bend occurs along the primary direction of the path. Check out https://github.com/xyflow/xyflow/pull/5376/commits/630b2385e24a5db66295e09eedaadd8941c3790e and the below updated screenshot after applying the change. 


Screenshot 2025-07-09 at 6 08 26 PM

moklick:
Thanks @kennyjwilli ! The `stepPosition` now only affects the default case (opposite handle positions) but I think that's fine.  I will check with the team but I think we can merge it :)

moklick:
Merged! Thanks @kennyjwilli 👌

kennyjwilli:
Welcome @moklick! Also, I must applaud the team -- it was so easy to get a local setup running. Literally just `install` & `dev` and it works. Great job.

moklick:
Thanks, that's great to hear @kennyjwilli ! `stepPosition` option is released in 12.8.2

PR #5384, opened 2025-07-05
fix(react): respect custom default node/edge type when falling back from unknown type

### LINKED ISSUE #5383: default nodeType overriding not working as expected

- React Flow / Svelte Flow version: 12.8.1
- Browser and version: Ungoogled chromium 137.0.7151.119
- OS and version: macOS Sequoia 15.5

### Live code example

https://codesandbox.io/p/sandbox/focused-glade-6wvq2j?file=%2Fsrc%2Fnodes%2Findex.ts%3A21%2C42

### Describe the Bug

I learned from [this discussion](https://github.com/xyflow/xyflow/discussions/4407) that I can override the `default` type of node with a custom Node component so that I can control how the fallback node will be rendered if its type is not registered via the `nodeType` prop. However, I tested the [mentioned approach](https://github.com/xyflow/xyflow/discussions/4407#discussioncomment-9941289) in my local project and the fallback node still uses the internal default node. I can also reproduce this behavior in the [provided codesandbox](https://codesandbox.io/p/sandbox/focused-glade-6wvq2j?file=%2Fsrc%2Fnodes%2Findex.ts%3A21%2C42).

One thing I noticed though is that if I don't set the `type` key in the initial node list, it will fallback to use the overridden default node component for rendering as expected (for example, the node with id "c" in the codesandbox). I wonder if this behavior should also apply to nodes with unregistered types (instead of only no types)?

Just fork the starter codesandbox and change the nodeTypes to:
```ts
export const nodeTypes = {
  default: PositionLoggerNode,
  // Add any of your custom nodes here!
} satisfies NodeTypes;
```

### Expected behavior

React flow should fallback to use the overridden default node component instead of the internal default node component for rendering when the type of the node is not registered via `nodeTypes`. In this example, the "drag me!" node should display x and y positions just like how "your ideas" node does.

### Screenshots or Videos

Image

### Additional context

_No response_

Sec-ant:
Fixes #5383.

Tell me if I do anything wrong, I'd be happy to fix them :). I didn't touch the svelte package because I have almost zero knowledge of it.

### DIFF
--- a/.changeset/forty-dryers-cry.md
+++ b/.changeset/forty-dryers-cry.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
+Fix node fallback to respect custom default node type when unknown node type is encountered
--- a/.changeset/purple-donuts-fail.md
+++ b/.changeset/purple-donuts-fail.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
+Fix edge fallback to respect custom default edge type when unknown edge type is encountered.
--- a/examples/react/src/App/routes.ts
+++ b/examples/react/src/App/routes.ts
@@ -9,6 +9,7 @@ import ControlledViewport from '../examples/ControlledViewport';
 import CustomConnectionLine from '../examples/CustomConnectionLine';
 import CustomMiniMapNode from '../examples/CustomMiniMapNode';
 import CustomNode from '../examples/CustomNode';
+import DefaultNodeOverwrite from '../examples/DefaultNodeOverwrite';
 import DefaultNodes from '../examples/DefaultNodes';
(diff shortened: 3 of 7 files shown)


Sec-ant:
I noticed the Edge has the [similar issue](https://github.com/xyflow/xyflow/blob/549ec62d6d667bb356e17dc6b7315c9b70d5b588/packages/react/src/components/EdgeWrapper/index.tsx#L49), if it is preferred I can also add a similar fix for the Edge to this PR.

Edit:

Similar fix for edge committed: ab05d00

moklick:
Thanks @Sec-ant !

peterkogo:
@moklick is this also relevant for Svelte Flow?

moklick:
@peterkogo it was already correct for Svelte Flow :)

PR #5388, opened 2025-07-07
Fix incorrect d3 event unbinding format

LuciferHuang:
## Issue Description
In the destroy method, an incorrect d3 event unbinding format `.on('.drag', null)` was used, causing the following error during execution:
```
TypeError: Cannot read properties of undefined (reading 'name')
```
The .on() method of d3-selection should use 'drag' instead of '.drag' to reference the drag event.

## Fix
Changed `d3Selection?.on('.drag', null)` to the correct syntax `d3Selection?.on('drag', null).`


moklick:
I think we are using it correctly. This is what the docs say:

> The listeners use the name .drag, so you can subsequently unbind the drag behavior as follows..

 https://d3js.org/d3-drag#_drag

PR #5390, opened 2025-07-08
docs(edges): update marker documentation to include formatting info

printerscanner:
From this issue in the web repo: https://github.com/xyflow/web/issues/477

I think the best place to update this is inside the api-reference. It seemed too specific to put in the example pages.


moklick:
Perfect! Could you add the same thing for Svelte Flow?

printerscanner:
@moklick updated

moklick (approved):

PR #5399, opened 2025-07-16
feat(examples): sync dropdown with current example on refresh

Amr-Shams:
- The dropdown now **correctly reflects the current example after a page refresh**.
  - Achieved by **parsing the example name from the URL** and using it to initialize the dropdown state.
- Added an `aria-label` to the dropdown for **improved accessibility**.


moklick:
Thanks @Amr-Shams . That's better :)

PR #5419, opened 2025-07-28
Style Arrow Heads Edge Markers with CSS Variables

0x0f0f0f:
Close https://github.com/xyflow/xyflow/issues/5366

### DIFF
--- a/packages/svelte/src/lib/store/initial-store.svelte.ts
+++ b/packages/svelte/src/lib/store/initial-store.svelte.ts
@@ -365,7 +365,9 @@ export function getInitialStore<NodeType extends Node = Node, EdgeType extends E
 
     selectNodesOnDrag: boolean = $derived(signals.props.selectNodesOnDrag ?? true);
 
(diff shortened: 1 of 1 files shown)


0x0f0f0f:
TODO: 
- [ ] Styling the whole example flow is not the best to test, find a solution

moklick:
This looks good already but the behaviour is slightly different for Svelte and React. If you pass "none" the arrow head is invisible in Svelte but it uses the new default css var in React. What do you think about adding the possibility to pass `null` to use the css var?

Ah and could you add a changeset?

0x0f0f0f:
TODO: 
- [x] Allow `null` in marker color (and default). use undefined as default.
- [x] `null` should be the option to use css variables. `undefined` should use the hardcoded default, or the `defaultEdgeOptions`.
- [x] Adjust comment in the prop for ReactFlow in tsdoc.
- [x] Add another node in example to test the behavior.
- [x] Check svelte flow behavior when `defaultColorMarker` is `null`

0x0f0f0f:
image

Behavior is now consistent between react and svelte

moklick:
thanks @0x0f0f0f  👍

Issue #5418, opened 2025-07-28
Type error with Node type and domAttributes.defaultValue incompatibility


ShlomoGalle:
### What platform were you using when you found the bug?

xyflow/react: 12.8.2
xyflow/system: 0.0.66
@types/react: 18.2.8

### Describe the Bug

After upgrading from xyflow/react 12.1.1 to 12.8.2, I’m encountering a TypeScript build error related to the 'Node' type. 
The 'Node' type appears to have been modified, specifically around the domAttributes property.
When I try to save a node in the Redux store, I get the following error:
```
Type 'Node' is not assignable to type 'WritableDraft<Node>'.
The types of 'domAttributes.defaultValue' are incompatible between these types.
Type 'string | number | readonly string[]' is not assignable to type 'string | number | string[]'.
```
It seems there is a type incompatibility on defaultValue between 'readonly string[]' and 'string[]'.

### Steps to reproduce the bug or issue

1. Attempt to save a Node object with domAttributes into a Redux store
2. Build the project and observe the TypeScript error.

### Expected behavior

The Node type should be compatible with WritableDraft<Node> without any type errors, allowing nodes with domAttributes (including defaultValue) to be saved in the Redux store without issues.

### Additional context

I think the root cause might be that defaultValue should not be included in the domAttributes type for Node, or should at least be omitted to avoid this incompatibility.
A potential workaround could be to define Node like this:
```
export type Node = {
  ...
  domAttributes?: Omit<HTMLAttributes<HTMLDivElement>, 'id' | ... | 'defaultValue' | keyof DomAttributes<HTMLDivElement>>;
}
```
Also, if possible, I’d looooove to open a PR myself to fix this (and the second issue I opened #5411 ) — just let me know if that would be welcome! :D
(I have other issues coming up if you're interested :D)

moklick:
Good catch! Could you check if that fixes the issue? Would be fine for me to adjust it. The other one will be fixed with the next release.

ShlomoGalle:
Thanks for the quick reply!
Yes, I’ve tested the change, the build (and the IDE errors) are resolved using that fix. Everything seems to work  🙌

moklick:
Cool! I think we can omit this field. If you want, you can create a PR for this.

moklick:
Thanks for your help here. Released in 12.8.3

PR #5420, opened 2025-07-28
chore(types): omit defaultValue from Node domAttributes to fix Redux typing issue #5418

ShlomoGalle:
This PR fixes a TypeScript incompatibility between the Node type and WritableDraft<Node> from immer, caused by the readonly string[] typing of defaultValue in domAttributes.

The fix omits defaultValue from Node['domAttributes'], restoring compatibility with Redux/Immer stores.

https://github.com/xyflow/xyflow/issues/5418

### DIFF
--- a/.changeset/chilly-donkeys-relax.md
+++ b/.changeset/chilly-donkeys-relax.md
@@ -1,5 +1,6 @@
 ---
 '@xyflow/react': patch
+'@xyflow/svelte': patch
 ---
 
-Omit `defaultValue` from `Node`'s `domAttributes` to fix type incompatibility with `WritableDraft<Node>`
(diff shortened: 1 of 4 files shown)


moklick:
Thanks for your help and bringing this up! I also adjusted the edge type and did the same thing for Svelte Flow.

ShlomoGalle:
I also wanted to add it for the Svelte package, but I realized that there's no defaultValue parameter in the HTMLAttributes type provided by the Svelte type library @moklick

moklick:
that makes sense .. thanks for checking

moklick:
released in 12.8.3! Thanks :)

PR #5427, opened 2025-07-31
Fix 5392 Intermediate zoom % are skipped when mouse scrolling on minimap in Windows OS

0x0f0f0f:
Fixes https://github.com/xyflow/xyflow/issues/5392

### DIFF
--- a/.changeset/chilly-donkeys-relax.md
+++ b/.changeset/chilly-donkeys-relax.md
@@ -1,6 +1,5 @@
 ---
 '@xyflow/react': patch
-'@xyflow/svelte': patch
 ---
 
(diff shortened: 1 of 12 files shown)


0x0f0f0f:
> The zoomStep variable is not used anymore. I think it should be 10 for macOs and 1 for Windows. Could you check?

It's handled by the `wheelDelta` function in `system/src/xypanzoom/utils.ts`, but it's called `factor`. It's `10` on `macOs` on `ctrl` key pressed, as the default zoom handling in the flow does.

moklick:
closed in favour of #5443

Issue #5273, opened 2025-05-19
Starting a connection by clicking a non-handle element


Karl255:
### Please describe the feature that you want to propose

In short: Introduce a function we can call to initiate the connection process.

In our use case we have a stylized button next to an unconnected handle the user can click on to start creating a connection. It essentially makes it easier to grab onto the handle. Image of what we have:

![Image](https://github.com/user-attachments/assets/dfbeb37d-7424-463c-9b43-b426d5ea9dad)

Currently we have a hacky implementation that dispatches a mousedown event on the handle when the user click down on the button:

```tsx
    const handleMouseDown = useCallback(() => {
        if (!handleRef.current) return;
        const bounds = handleRef.current.getBoundingClientRect();
        const eventInit: MouseEventInit = {
            clientX: (bounds.left + bounds.right) / 2,
            clientY: (bounds.top + bounds.bottom) / 2,
            bubbles: true,
        };

        handleRef.current.dispatchEvent(new MouseEvent('mousedown', eventInit));
    }, []);
```

But this still requires the actual handle to be at least 1x1 pixels in size due to the underlying Handle implementation (calls `document.elementFromPoint()` at the event's coordinates). This in the end doesn't exactly fit our intended design. **Ideally there should be a function we can call to initiate this connection process.** For our use case it'd be acceptable if we still need to pass in a real mouse event if that's needed, since we capture one anyway.

bcakmakoglu:
Can't you just put your `<button>` inside the `<Handle>`? 🤔

For example:
```tsx
      <Handle
        type="source"
        position={Position.Bottom}
        style={{
          display: "flex",
          justifyContent: "center",
          alignItems: "center",
          width: 16,
          height: 16,
          background: "transparent",
        }}
      >
        <button
          style={{
            display: "flex",
            justifyContent: "center",
            alignItems: "center",
            border: "1px solid black",
            borderRadius: 9999,
            backgroundColor: "pink",
            fontSize: 8,
            width: "100%",
            height: "100%",
            pointerEvents: "none",
          }}
        >
          X
        </button>
      </Handle>
```

Here's an example: https://codesandbox.io/p/sandbox/distracted-hoover-ftj2yx?file=%2Fsrc%2Fnodes%2FPositionLoggerNode.tsx%3A22%2C1-50%2C16

Karl255:
[...]

PR #5428, opened 2025-07-31
Allow detached handles 

Karl255:
This change enables having detached elements (outside the handle's bounding box) in handles which will still trigger edge connections.


Bqnic (approved):
LGTM

moklick:
#5273

moklick:
This is a very good idea! I adjusted the variable names slightly and added it for Svelte Flow too. Thanks for your help here :)

PR #5438, opened 2025-08-08
docs: add link to Chinese React Flow docs in README

dowithless:
Source code for the translated documentation is available at: https://github.com/readme-i18n/xyflow-zh


moklick:
Hey! Sorry but we won't add this. It would be nice to have a Chinese translation but we won't link to a source that we don't know and that's not from us. We might add translation to our docs at some point!

PR #5443, opened 2025-08-11
chore(minimap): fix zooming on windows 

moklick:


packages/system/src/xyminimap/index.ts
```diff
@@ -38,18 +38,19 @@ export function XYMinimap({ domNode, panZoom, getTransform, getViewScale }: XYMi
     zoomable = true,
     inversePan = false,
   }: XYMinimapUpdate) {
+    const _zoomStep = zoomStep / 10;
```
0x0f0f0f: @moklick as discussed today, this should be multiplied by 10 if `event.ctrlKey && isMacOs()`, as in `wheelDelta` function for pinch zoom

PR #5444, opened 2025-08-11
Export MiniMapNode

paula-stacho:
It's already possible to use a custom `nodeComponent`, this enables the users to used the original `MiniMapNode` with a custom wrapper


moklick:
Thanks! Available in 12.8.3

PR #5452, opened 2025-08-13
Set unique id in connectionLookup to fix multi selection if there are multiple edges between two nodes

Ni2Be:
Fix for issue https://github.com/xyflow/xyflow/issues/5451
With this change the source and target id should always be unique so no edges get lost in the connectionLookup map.


peterkogo:
The id is unique.

PR #5455, opened 2025-08-16
fix: skip resize handler updates when DOM node is not visible

Sec-ant:
Prevent unnecessary resize calculations and error messages for invisible elements by checking visibility before processing dimensions.

We use `display:none` in our project to hide a react flow canvas on demand without unmounting or destroying its internal states. This patch prevents unnecessary calculations, and also suppresses the "The React Flow parent container needs a width and a height to render the graph." error message in this case.

### DIFF
--- a/.changeset/stupid-lizards-wash.md
+++ b/.changeset/stupid-lizards-wash.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
+Skip resize handler updates when DOM node is not visible
--- a/.changeset/stupid-lizards-wash.md
+++ b/.changeset/stupid-lizards-wash.md
@@ -2,4 +2,4 @@
 '@xyflow/react': patch
 ---
 
-Skip resize handler updates when DOM node is not visible
+Fix warning when display is set to none on the wrapper div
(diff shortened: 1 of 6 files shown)


Sec-ant:
Hi @moklick, would love to hear your thoughts on this one.

moklick:
Hey @Sec-ant , first of all thanks for helping out here :) I need to check if this has any other effects. Are you aware that these updates are only happening when the user resizes the window?

Sec-ant:
> Hey @Sec-ant , first of all thanks for helping out here :) I need to check if this has any other effects. Are you aware that these updates are only happening when the user resizes the window?

Apart from window resize events, the `updateDimensions` is also called as a handler of the `resizeObserver` when the size of the `<ReactFlow/>` component is changed, [including visibility change](https://www.w3.org/TR/resize-observer/#:~:text=about%20these%20observations%3A-,Observation%20will%20fire%20when%20watched%20Element%20is%20inserted/removed%20from%20DOM,Observation%20will%20fire%20when%20watched%20Element%20display%20gets%20set%20to%20none.,-Observations%20do%20not) (therefore skipping resize handler updates when the component is not visible is okay because its size will be recalculated when it becomes visible again).

peterkogo:
Another stellar PR, thank you @Sec-ant!
`checkVisibility ` is a Baseline 2024 browser feature, however it is being checked explicitly.

I realized the style property was not being propagated correctly in Svelte Flow, so I included it in this PR.

PR #5459, opened 2025-08-18
Fix open arrow marker using fill color from edge instead of none

0x0f0f0f:
Fix https://github.com/xyflow/xyflow/issues/5456


moklick:
Now that I look at it again, I would prefer to add a class name to the marker to be able to distinguish between a closed and normal arrow. Then we could make the css more specific and won't need to overwrite the fill with fill none.

moklick:
Thanks @0x0f0f0f !

PR #5480, opened 2025-08-26
Fix connection pinch

### LINKED ISSUE #5475: Pinching can lead to dangling connection line

- React Flow version: 12.8.4
- Browser and version: Edge 139.0, Firefox 142.0
- OS and version: Windows 11 IOT Enterprise, Windows 11 Professional

### Live code example

_No response_

### Describe the Bug

While dragging a connection line on a touch screen, touching the screen with a second finger and moving the two fingers toward each other will start a new gesture (pinch to zoom), at which point the connection line is not canceled, but rather stays on the screen while no longer following the dragging finger. Even when both fingers stop touching the screen, the connection line persists. The video below demonstrates this using one of the examples on the reactflow website.

1. Go to https://reactflow.dev/ using a touch device
2. In the example flow on the homepage, start dragging an edge using a finger
3. Touch the screen with another finger
4. Move the two fingers toward each other a little bit
5. Remove your fingers from the screen
6. The connection line will still be there

### Expected behavior

There are two possible behaviours that one might expect when the second finger touches the screen

- Either the connectionline disappears because the drag action is canceled and the pinch to zoom action has started.
- Or the drag action is not canceled, the second finger does not initiate a new gesture, and the connectionline keeps following the first finger.

### Screenshots or Videos

https://github.com/user-attachments/assets/97b8872b-dadd-47ae-ab52-29f6a03c22c9

### Additional context

_No response_

peterkogo:
closes #5475

### DIFF
--- a/packages/system/src/xypanzoom/filter.ts
+++ b/packages/system/src/xypanzoom/filter.ts
@@ -12,6 +12,7 @@ export type FilterParams = {
   noWheelClassName: string;
   noPanClassName: string;
   lib: string;
+  connectionInProgress: boolean;
 };
 
 export function createFilter({
@@ -25,6 +26,7 @@ export function createFilter({
   noWheelClassName,
   noPanClassName,
   lib,
+  connectionInProgress,
 }: FilterParams) {
   return (event: any): boolean => {
     const zoomScroll = zoomActivationKeyPressed || zoomOnScroll;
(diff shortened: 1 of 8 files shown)


packages/system/src/xypanzoom/filter.ts
```diff
@@ -48,6 +50,11 @@ export function createFilter({
       return false;
     }
 
+    // we want to disable pinch-zooming while making a connection
+    if (connectionInProgress) {
```
moklick: would it be possible that we still allow to use the mousewheel here?

moklick (approved):

PR #5486, opened 2025-08-29
Add a NodeResizeControl identifier to NodeDimensionChange

0x0f0f0f:
Close https://github.com/xyflow/xyflow/issues/5485


peterkogo:
I would postpone this change for now, but it is indeed a great case for streamlining node changes to get the middlewares working.

Issue #5204, opened 2025-04-11
snapToGrid = true causes selection glitch during multiselect


corbinklett:
### What platform were you using when you found the bug?

- React Flow / Svelte Flow version: 12.5.5
- Browser and version: Chrome various versions
- OS and version: Mac OS 15.3.2

### Describe the Bug

When snapToGrid = true, and while holding the multiselect key,  the node selection behavior becomes glitchy. 

In this situation, clicking on a node can result in a node change of type 'position'. It can get pretty bad too, making it difficult to select or deselect a node.

### Steps to reproduce the bug or issue

In <ReactFlow> set

snapToGrid={true}

In a diagram with two or more nodes, hold down the multiselect key and repeatedly select and unselect nodes. If you look at the node changes events, sometimes you will get type: 'position' rather than type: 'selection'. 

### Expected behavior

I would expect for a 'position' event not to result from a click on a node.

### Screenshots or Videos

https://www.tella.tv/video/selection-glitch-gk57

In this video, I hold the multiselect key (cmd on a mac) and click on the device on the right. You will see that it flashes purple. Looking at the node changes in the console, you will see the following sequence of events:

1. Selection change: selected: true
2. Position change
3. Selection change: selected: false

#2 and #3 are unintended. If I set snapToGrid={false} this glitch does not occur.

corbinklett:
I do not have `selectOnDrag = false`. 

Increasing the `nodeDragThreshold` to the grid size (20px) works, but it is an undesirable workaround since it makes dragging feel clunkier. The problem even occurs with  a large threshold like 10px, even though when clicking I am definitely not moving the cursor 10 pixels while the mouse is down.

moklick:
Could you give us some more context? What are the initial positions of your nodes and what are your `snapGrid` values?

corbinklett:
The problem shown in the video can be easily reproduced with the simplest diagram:

```
      <ReactFlow
        nodes={currentNodes}
        onNodesChange={onNodesChange}
        fitView
        snapToGrid={true}
      >
```

where `onNodesChange` comes from `useNodesState` in React Flow. `currentNodes` is two default nodes (or maybe even just one). The problem occurs with `snapToGrid={true}` and does not occur when that line is commented out.

Happy to make another video with this case if that helps!

peterkogo:
I just tried again with `snapGrid={true}` which applies the default `snapGrid=[15, 15]`.

[...]

Issue #4996, opened 2025-02-05
Inconsistent Node Click Event Registration


ahmed-3li:
### What platform were you using when you found the bug?

@xyflow/svelte: 0.1.29

### Describe the Bug

When clicking nodes in a Svelte Flow canvas, some clicks are not being registered by the `nodeclick` event handler. This occurs even with a minimal implementation and with all other interactive features disabled.

## Reproduction
Here's a minimal example that demonstrates the issue:

```
<script>

  import {
    SvelteFlow,
    Background,
    BackgroundVariant,
  } from '@xyflow/svelte';
  import '@xyflow/svelte/dist/style.css';
  import { writable } from 'svelte/store';
 
  const nodes = writable([
    {
      id: '1',
      type: 'input',
      data: { 
        label: 'A',
        clicks: 0
      },
      position: { x: 0, y: 0 }
    },
    {
      id: '2',
      type: 'default',
      data: { 
        label: 'B',
        clicks: 0
      },
      position: { x: 0, y: 150 }
    }
  ]);
 
  const edges = writable([
    {
      id: '1-2',
      type: 'default',
      source: '1',
      target: '2',
      label: 'Edge Text'
    }
  ]);

  function handleNodeClick(event) {
    console.log('Click detected on node:', event.detail.node.id);
    
    nodes.update(currentNodes => {
      return currentNodes.map(node => {
        if (node.id === event.detail.node.id) {
          return {
            ...node,
            data: {
              ...node.data,
              clicks: (node.data.clicks || 0) + 1
            }
          };
        }
        return node;
      });
    });
  }
</script>
 
<div style:height="500px">
  <SvelteFlow
    {nodes}
    {edges}
    fitView
    on:nodeclick={handleNodeClick}
  >
    <Background variant={BackgroundVariant.Dots} />
  </SvelteFlow>
</div>

<div>
  <div>
    <h2>Node A</h2>
    <p>Clicks: {($nodes).find(n => n.id === '1').data.clicks}</p>
  </div>
  <div>
    <h2>Node B</h2>
    <p>Clicks: {($nodes).find(n => n.id === '2').data.clicks}</p>
  </div>
</div>
```

### Steps to reproduce the bug or issue

1. Create a new Svelte project with @xyflow/svelte
2. Implement the minimal example above
3. Test clicking pattern:
   - Click Node A, then Node B, then A, then B...
   - Do this alternating pattern about 10 times
4. Expected: Each click should increment the respective node's counter
[...]

PR #5489, opened 2025-09-01
fix: node selection instability by using client coordinates for drag threshold

### LINKED ISSUE #5412: Selecting nodes does not work reliably

- React Flow / Svelte Flow version: React Flow 12.8.2
- Browser and version: Chrome 138 (also not working in my Electron app)
- OS and version: MacOS Sequoia 15.0

### Live code example

https://stackblitz.com/edit/crekuenp?file=ResizableNodeSelected.jsx

### Describe the Bug

Selectable, resizable nodes do not get selected upon click reliably.

1. Create two or more selectable, resizable nodes
2. Keep selecting one or the other
3. Watch how after almost every 3rd time it stops working (unpredictable)

### Expected behavior

Nodes to be selected reliably with `selected` as `true` if the node is selected, and `selected` as `false`

### Screenshots or Videos

https://github.com/user-attachments/assets/f974321e-9956-43c9-bad8-c4ce39f8cb60

### Additional context

Please watch video and check out the StackBlitz for reference

Sec-ant:
This PR tries to fix #5412 and #4996 by using the client coordinates for `nodeDragThreshold`.

In #4996, [we learned that](https://github.com/xyflow/xyflow/issues/4996#issuecomment-2668547221) `nodeDragThreshold` uses flow coordinate system while `nodeClickDistance` uses client coordinate system, so it is impossible for a user to configure these two props simultaneously to get a correct configuration with a higher distance tolerance.

> [@moklick](https://github.com/moklick) I just found out that `nodeClickDistance` and `nodeDragThreshold` don't measure the same thing, which makes configuring when a node should be dragged vs clicked impossible.
> 
> `nodeDragThreshold` is measured in flow coordinates and `nodeClickDistance` in client coordinates.
> 
> How should we proceed with this?

And with the default low tolerance configuration (`nodeDragThreshold = 1` and `nodeClickDistance = 0`), many users complained that node selection is not very reliable.

We use ReactFlow in one of our internal platform and my co-worker who uses a trackpad instead of a mouse told me that he frequently encountered the unstable node selection issue (1 or more out of 10) on our platform. He noticed that whenever a failed selection is encountered, his cursor would transform [from `grab` to `grabbing`](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor) for a split second. Thus we came to a conclusion that it is highly probable that this issue is caused by the node drag threshold unit.

So this PR is a fix to the `nodeDragThreshold` to let it use the client coord sys instead of flow coord sys when interpreting the distance.

As you may have noticed, I first used copilot to do some vibe experimenting to find a solution in my fork: https://github.com/Sec-ant/react-flow/pull/2. It drew similar conclusions. And in that experiment, I considered about not bringing in any breaking changes by introducing a new prop `nodeDragDistance` (suits more as a counterpart for `nodeClickDistance`, wording-wise) which uses the client coord system, and takes precedence over `nodeDragThreshold` when present. However, I reviewed the code and I thought it would involve some refactoring on the svelte side which I have very little confidence in. Consider that @moklick [regards this as a bug](https://github.com/xyflow/xyflow/issues/4996#issuecomment-2792204191), I decided to take the simpler path without introducing any new props.

I also noticed that it was using `pointerPos.xSnapped` and `pointerPos.ySnapped` when calculating the distance. I'm not very sure about the intended behavior here, but to my understanding, the `nodeDragDistance` should only account for the drag intention instead of result, so I didn't consider any effect caused by the snapping when migrating to the client coordsys.

This can be considered as a bug fix or a breaking change, so please tell me what do you think of it, and hope this can bring some momentum to fix related issues like #5412, #4996 and  #5204 for good.

### DIFF
--- a/packages/system/src/xydrag/XYDrag.ts
+++ b/packages/system/src/xydrag/XYDrag.ts
@@ -336,8 +336,8 @@ export function XYDrag<OnNodeDrag extends (e: any, nodes: any, node: any) => voi
         if (!dragStarted) {
           // Calculate distance in client coordinates for consistent drag threshold behavior across zoom levels
           const currentMousePosition = getEventPosition(event.sourceEvent, containerBounds!);
-          const x = currentMousePosition.x - (mousePosition.x ?? 0);
-          const y = currentMousePosition.y - (mousePosition.y ?? 0);
+          const x = currentMousePosition.x - mousePosition.x;
+          const y = currentMousePosition.y - mousePosition.y;
           const distance = Math.sqrt(x * x + y * y);
 
           if (distance > nodeDragThreshold) {


Sec-ant:
I have tested that after this patch is applied, and using a same little higher value (like `5` or `10`) for both `nodeDragThreshold` and `nodeClickDistance`, I got a very reliable and stable node selection and dragging response. (The `example/switch` example is a good place to test it. EDIT: probably not, we need some custom css to distinguish `focused` from `selected`).

However, when only increasing `nodeDragThreshold` without changing `nodeClickDistance`, ~~you may end up selecting more than one nodes when clicking around~~ (this is probably incorrect, I cannot differentiate `focused` from `selected`). Personally I think these two props should be considered at the same time when one needs to change one of them, otherwise the events and node states may get confusing, and I believe this should be documented somewhere.

peterkogo:
Thanks you so much for this PR! This definitely resolves the issue. 
I have removed an unnecessary non-null assertion and changed up the patch notes. otherwise it's good to go!

I also agree with pretty much all your observations! And yes, the way `nodeClickDistance` and `nodeDragThreshold` are not combined is indeed unfortunate...

PR #5491, opened 2025-09-02
Fix visual artifacts around handles when zoomed out

PicoLeaf:
Hello, I noticed there was some visual artifacts around the handles when zoomed out, it seems that some pixel of the black background of the handles can be seen outside of the border due to some precision error both in chrome and firefox
This is easily fixed by using an outline instead of a border :)

Screenshot_20250902_134310


moklick:
Hey! This is a very good finding and we would like to merge it. Unfortunately this would break existing css overwrites from users and it would be a breaking change.

PR #5492, opened 2025-09-03
Preserve cursor during resizing

0x0f0f0f:
Closes https://github.com/xyflow/xyflow/issues/5471

### DIFF
--- a/packages/system/src/styles/node-resizer.css
+++ b/packages/system/src/styles/node-resizer.css
@@ -11,23 +11,9 @@
   cursor: ew-resize;
 }
 
-.xy-flow__pane:has(.xy-flow__resize-control.left.resizing),
-.xy-flow__pane:has(.xy-flow__resize-control.right.resizing),
-.xy-flow__node:has(.xy-flow__resize-control.left.resizing),
-.xy-flow__node:has(.xy-flow__resize-control.right.resizing) {
-  cursor: ew-resize !important;
-}
-
 .xy-flow__resize-control.top,
 .xy-flow__resize-control.bottom {
   cursor: ns-resize;
-}
-
-.xy-flow__pane:has(.xy-flow__resize-control.top.resizing),
-.xy-flow__pane:has(.xy-flow__resize-control.bottom.resizing),
-.xy-flow__node:has(.xy-flow__resize-control.top.resizing),
-.xy-flow__node:has(.xy-flow__resize-control.bottom.resizing) {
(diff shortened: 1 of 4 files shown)


0x0f0f0f:
TODO:
- [x] Make a solution for Svelte

packages/system/src/styles/node-resizer.css
```diff
@@ -26,6 +26,22 @@
...
+  cursor: ns-resize !important;
+}
+
+body:has(.xy-flow__resize-control:is(.bottom.left, .top.right).resizing) * {
+  cursor: nesw-resize !important;
+}
+
+body:has(.xy-flow__resize-control:is(.top.left, .bottom.right).resizing) * {
+  cursor: nwse-resize !important;
+}
```
0x0f0f0f: We can't just simply set the cursor on the pane, because it is often overridden by other selectors (e.g. other nodes will still have a `grabbing` cursor when you are resizing and you hover them, as their selector will be more specific). This makes sure that the cursor won't change, but it might be an aggressive solution.

0x0f0f0f:
TODO: 
- [x] try out more selective selectors to avoid potential performance issues

packages/system/src/styles/node-resizer.css
```diff
@@ -26,6 +26,26 @@
...
+body:has(.xy-flow__resize-control:is(.bottom.left, .top.right).resizing)
+  :is(.xy-flow__node, .xy-flow__resize-control, .xy-flow__pane, .xy-flow__handle) {
+  cursor: nesw-resize;
+}
+
+body:has(.xy-flow__resize-control:is(.top.left, .bottom.right).resizing)
+  :is(.xy-flow__node, .xy-flow__resize-control, .xy-flow__pane, .xy-flow__handle) {
+  cursor: nwse-resize;
+}
+
```
0x0f0f0f: @peterkogo do you think this selector might still have perf issues?

peterkogo:
Unfortunately after making this change the selector now takes up the majority of css selector time (like 95%).
Alessandro made the suggestion to add a global cursor override on the body when resizing. Which is pretty ugly but the only viable workaround.
The actual solution for this issue is to kick out d3-drag and use pointer capture. IMO this is doable for RF 13

moklick:
We leave the issue open (with a "next major" tag) and close this PR for now. Thanks for your ideas here @0x0f0f0f

PR #5497, opened 2025-09-08
skip force rendering nodes that have handles defined

peterkogo:


packages/system/src/utils/store.ts
```diff
@@ -109,13 +140,15 @@ export function adoptUserNodes<NodeType extends NodeBase>(
         ..._options.defaults,
         ...userNode,
         measured: {
-          width: userNode.measured?.width,
-          height: userNode.measured?.height,
+          width: userNode.width ?? userNode.measured?.width,
```
moklick: If we do this, RF won't trigger dimension changes after the first rendering and `measured` won't end up in the user nodes.

moklick (changes requested):

PR #5529, opened 2025-09-30
Fix svelte selection

peterkogo:


packages/svelte/src/lib/store/index.ts
```diff
@@ -235,12 +235,13 @@ export function createStore<NodeType extends Node = Node, EdgeType extends Edge
       const nodeWillBeSelected = ids.includes(node.id);
       const selected = isMultiSelection ? node.selected || nodeWillBeSelected : nodeWillBeSelected;
 
-      if (node.selected !== selected) {
+      console.log(selected);
```
syaltamimi: Is this log message meant to be cleaned up?

peterkogo: haha yes, thanks!

packages/svelte/src/lib/store/index.ts
```diff
@@ -235,12 +235,11 @@ export function createStore<NodeType extends Node = Node, EdgeType extends Edge
       const nodeWillBeSelected = ids.includes(node.id);
       const selected = isMultiSelection ? node.selected || nodeWillBeSelected : nodeWillBeSelected;
 
-      if (node.selected !== selected) {
+      if (!!node.selected !== selected) {
         // we need to mutate the node here in order to have the correct selected state in the drag handler
```
moklick: is the comment outdated?

peterkogo: you are correct!

PR #5531, opened 2025-10-01
fix(edges): add `type` property to `EdgeProps`

Sec-ant:
Adds the `type` property to `EdgeProps` to match `NodeProps` behavior. This allows custom edge components to access their edge type with proper TypeScript support.

The `type` property was already being passed [at runtime in `EdgeWrapper`](https://github.com/xyflow/xyflow/blob/4c2a3030575557b8270575e88fa5366bc6e1b476/packages/react/src/components/EdgeWrapper/index.tsx#L217) but was missing from the TypeScript type definition.

### DIFF
--- a/.changeset/curly-steaks-check.md
+++ b/.changeset/curly-steaks-check.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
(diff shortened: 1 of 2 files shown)


moklick:
good catch! Thanks :)

PR #5536, opened 2025-10-03
feat: Add ability to controll selection state externally

FredrikMWold:
## Summary
This PR introduces a new `selecting` prop to React Flow that allows developers to programmatically toggle selection mode without requiring users to hold down a key.

## Motivation
Currently, React Flow only supports box selection when the user holds a modifier key. While this works well on desktop, it can be cumbersome on laptops.  

We want to implement a toolbar with two buttons: **Pan** and **Box select**. Clicking "Select" enables selection mode in React Flow without requiring the user to hold `Shift`.

### DIFF
--- a/.changeset/chilled-oranges-turn.md
+++ b/.changeset/chilled-oranges-turn.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': major
+---
+
+Add ability to controll selection state externally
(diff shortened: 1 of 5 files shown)


moklick:
Hey! This is already possible by setting selectionOnDrag and panOnDrag: https://reactflow.dev/learn/concepts/the-viewport#design-tool-viewport-controls

FredrikMWold:
@moklick That does not work in the react basic example

FredrikMWold:
Ah, you need to set panOnDrag to false as well

moklick:
Yep! Bad API decisions :D We will hopefully change this with the next major release.

PR #5550, opened 2025-10-15
Feature/subflow sort order

### LINKED ISSUE #5400: The priority issue of child nodes of a parent node

- Browser and version: 
- OS and version:

Image

### Live code example

_No response_

### Describe the Bug

There are two loop nodes here. When they are placed together, the priority of the child nodes inside is higher than that of all the parent nodes. It seems that this should not be the case

1.create two parent nodes and add their respective child nodes
2. crag the parent node to slightly overlap them, and then select one of the child nodes

### Expected behavior

I hope that the priority of the unselected node cannot be higher than that of the selected parent node

### Screenshots or Videos

_No response_

### Additional context

_No response_

peterkogo:
closes #5400

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented child nodes from different parents from overlapping and improved stacking/elevation for clearer nested/grouped layouts.

* **Documentation**
  * Simplified the React example's initial graph to better reflect current subflow behavior.

* **Chores**
  * Bumped minor versions across affected packages.

### DIFF
--- a/.changeset/eighty-wolves-cheat.md
+++ b/.changeset/eighty-wolves-cheat.md
@@ -1,5 +1,5 @@
 ---
-'@xyflow/system': minor
+'@xyflow/system': patch
 '@xyflow/react': minor
 '@xyflow/svelte': minor
 ---


packages/system/src/utils/store.ts
```diff
@@ -120,6 +120,7 @@ export function adoptUserNodes<NodeType extends NodeBase>(
 ): boolean {
   const _options = mergeObjects(adoptUserNodesDefaultOptions, options);
 
+  let rootParentIndex = { i: -1 };
```
moklick, line 126: why is this an object?

packages/system/src/utils/store.ts
```diff
@@ -212,6 +214,17 @@ function updateChildNode<NodeType extends NodeBase>(
 
   updateParentLookup(node, parentLookup);
 
+  // We just want to set the rootParentIndex for the first child
+  if (rootParentIndex && !parentNode.parentId && parentNode.internals.rootParentIndex === undefined) {
+    parentNode.internals.rootParentIndex = ++rootParentIndex.i;
+    parentNode.internals.z = parentNode.internals.z + rootParentIndex.i * 10;
```
moklick: maybe add a comment about the "10" :D

moklick (changes requested):

packages/system/src/utils/store.ts
peterkogo: We only need to increment whenever a rootParent is found, which happens inside the function call - though the index needs to be tracked globally

Issue #5342, opened 2025-06-16
Start a selection above a node or group node


moklick:
Currently it's not possible to start a selection above a node. Especially if you are working with sub flows, this can be really annoying. It would be nice if we could support it somehow out of the box.

AquariusGodOfCookier:
I have the same problem now. Is there any way to implement this function first? Create a new Flow via ReactFlowProvider?

herbertpimentel:
@AquariusGodOfCookier for context, this is what we dicussed previously in discord. it does have some example that may interest you.

https://discord.com/channels/771389069270712320/1380353826984558723

--------

Herbert Pimentel
AP
 — 10/06/2025 23:32
@usmanabdurrehman thanks for you dedication man, you are amazing... 
I am trying to accomplish/mimic the similar behavior of comment node from unreal blueprints. 
I created a special node that groups other nodes, the groups can be some times full of nodes, so It will be useful to allow easily select them  with marquee selection as I already doing outside of the the grouping node. But I currently don't see a way to do that with the nodes in the group. 

I also implemented it  the same way as in your example , it was the most close I could  get so far

you may better understand watching this https://www.loom.com/share/f156938f3b154d1388a8a3920cf7847e 


usmanabdurrehman
[QT]
 — 07/06/2025 06:37
Hmn. Even with a drag handle for the group node, wasnt able to make a selection box inside a group in this [example](https://codesandbox.io/p/devbox/epic-visvesvaraya-g4xps7). 

You can initiate a selection box from outside a group and select nodes from within a group. You can do that thou.

ysds:
Actually, when `selectionMode` is set to `"partial"`, it's not possible to select only the nodes inside a group by starting the selection from outside. The group node itself gets selected instead.

It would be great if we could allow group nodes to have `"full"` selection behavior even when the global `selectionMode` is set to `"partial"`. This would enable users to:
- Start selection from outside a group
- Select individual nodes within the group without selecting the group node itself
- Have more granular control over selection behavior for different node types

This would particularly improve the UX when working with complex sub-flows and grouped nodes.

moklick:
We talked about this today and we are not sure if it's too risky to change the default behaviour. It's possible that users want a different behaviour for a node click (with pressed Shift) for example and we would change that. It's also not nice to introduce a new prop like `forceSelectionWithSelectionKeyCode` or something like that. We will have another look next week.

herbertpimentel:
@moklick @peterkogo 

[...]

PR #5551, opened 2025-10-15
refactor(selection): make it possible to start a selection above a node

moklick:
#5342

## Summary by CodeRabbit

* **New Features**
  * Start a drag multi-select even when beginning above or over a node.

* **Bug Fixes**
  * Elements marked to ignore selection (e.g., nokey) no longer start selection.
  * Improved pointer/click capture to prevent unintended clicks during/after drag selection.
  * Regression introduced to input detection that may affect selection around form fields — tracking for follow-up.

* **Chores**
  * Patch version bumps for multiple packages.

### DIFF
--- a/.changeset/fresh-groups-kneel.md
+++ b/.changeset/fresh-groups-kneel.md
@@ -1,7 +1,7 @@
 ---
-'@xyflow/svelte': patch
+'@xyflow/svelte': minor
 '@xyflow/system': patch
-'@xyflow/react': patch
+'@xyflow/react': minor
 ---
 
 Allow to start a selection above a node


peterkogo:
As per [PointerEvent specs](https://www.w3.org/TR/pointerevents/#mapping-for-devices-that-do-not-support-hover) we can expect the PointerDown to be always before the MouseDown. And doing e.preventDefault should also prevent the mousedown and click from firing.

herbertpimentel:
@moklick @peterkogo  I am not sure if this was the expected behaviour, or I am doing something wrong, but this selection on top of nodes works but looks to be depending of the "selection key (shift)"  to be pressed, not really obeying the already set behaviour when we have  panOnDrag={[1]} // LMB for selection

peterkogo:
@herbertpimentel and how do you drag the nodes? Or is this about completely static flows where you cannot click or drag the nodes / anything inside the nodes?

herbertpimentel:
> @herbertpimentel and how do you drag the nodes? Or is this about completely static flows where you cannot click or drag the nodes / anything inside the nodes?

@peterkogo I am not sure if is this your question, but In my graph, I am using the left mouse button to make the selection with no need to use  "selection key (shift)"

I am using a graph with the property `panOnDrag={[1, 2]}`

acoord to the docs, this settings allows me to:

1 - use the first mouse button as select (with no need to press shift to start selection)
2 - use second mouse button as drag of the graph

I am replicating the exact same behaviour of Unreal Blue Prints Engine. 
With this setting on panOnDrag, there's no need to press shit key to start a selection area 
This works perfect as intented the graph, but once we try to do that on top of nodes. It requires to press shift to start make the selection area. 

Let me know you need any clarification. 

in this view I show you now it works in unreal (the exact behavior I am trying to accomplish). 

https://www.loom.com/share/f156938f3b154d1388a8a3920cf7847e

peterkogo:
@herbertpimentel  Thanks for the clarification, I understand now! I will try to figure out how this can be handled.

herbertpimentel:
> @herbertpimentel Thanks for the clarification, I understand now! I will try to figure out how this can be handled.

@peterkogo thank yo very much, I was just wondering if you already have another issue/card or so that I can follow to know when it get done.

herbertpimentel:
@peterkogo sorry I belive that I get lost to track it ?! is there anyplace where I can get aware when it get done ?

PR #5560, opened 2025-10-21
Enhance: Svelte Flow with typed context

peterkogo:
## Summary by CodeRabbit

* **Chores**
  * Updated Svelte tooling and dev dependencies; bumped minimum Svelte version to >=5.40.0
  * Modernized TypeScript, Prettier, ESLint and build tooling versions

* **Style / Config**
  * Replaced example ESLint config with a new composed flat config; tightened example .gitignore/.eslintignore rules
  * Removed an unnecessary npm config line

* **Refactor**
  * Consolidated context handling across components and hooks into new type-safe context utilities
  * Removed a development-only runtime warning
  * Made nodeId optional on resize controls with documented guidance when used outside custom nodes

### DIFF
--- a/.changeset/nasty-clouds-shake.md
+++ b/.changeset/nasty-clouds-shake.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/svelte': patch
+---
+
+Remove `edgeId` from `EdgeToolbar` props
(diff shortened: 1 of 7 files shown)


.changeset/hungry-jobs-wave.md
```diff
@@ -0,0 +1,5 @@
+---
+'@xyflow/svelte': minor
+---
+
+Bump required Svelte version to >=5.40.0
```
moklick: should we say something like "to support typed contexts" in addition?

peterkogo: It's just for internal reasons. Nothing hinders a user to use the newest svelte version with typed contexts.

PR #5572, opened 2025-10-27
Fix/pane events

### LINKED ISSUE #5563: selectionOnDrag prevents pane mouse down events from triggering

- React Flow: 12.9.0
- Browser and version: 
- OS and version: mac

### Live code example

12.9.0: https://stackblitz.com/edit/wwkr61di?file=App.tsx
12.8.5: https://stackblitz.com/edit/wwkr61di-tjkmrdbu?file=App.tsx

### Describe the Bug

When selectionOnDrag is enabled, pane mouse events are no longer triggered. Additionally, when a custom Input node's input element is focused and the pane is clicked, the input element's onBlur event used to be triggered (removing focus), but now the focus is maintained.

This issue does not occur in version 12.8.5, but appears in version 12.9.0.

If this is an intentional change, please feel free to ignore this report. However, I am affected by this behavioral change and have not yet found a way to make the input element's onBlur trigger as expected.

1. Enable `selectionOnDrag` and disable `panOnDrag` on ReactFlow
  ```
  selectionOnDrag={true}
  panOnDrag={false}
```
2. Create a custom node with an input element
3. Focus the input element
4. Click on the pane area
5. Observe that onBlur is not triggered and focus remains on the input

### Expected behavior

- Pane mouse events should trigger same as 12.8.5
- Input elements should lose focus (onBlur should trigger) when clicking on the pane

### Screenshots or Videos

_No response_

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by peterkogo**
fixes #5563

### DIFF
--- a/.changeset/thirty-goats-invent.md
+++ b/.changeset/thirty-goats-invent.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/svelte': patch
+---
+
+Remove focus outline from selection box

### CODE REVIEW

**Comment by ysds** [comment]
Thank you for addressing this! I tested with the following code and it appears that the issue from #5563 has been resolved, and the behavior seems to have returned to match 12.8.5.

examples/react/src/examples/CustomNode/index.tsx
```
@@ -169,6 +169,14 @@ const CustomNodeFlow = () => {
       maxZoom={2}
       onBeforeDelete={onBeforeDelete}
       ref={ref}
+      selectionOnDrag={true}
+      panOnDrag={false}
+      onMouseDown={() => {
+        console.log('onMouseDowna');
+      }}
+      onMouseDownCapture={() => {
+        console.log('onMouseDownCapture');
+      }}
     >
       <MiniMap<MyNode>
         nodeStrokeColor={(n: MyNode): string => {
 ```
 I've also confirmed that the input blur issue has been resolved as well.

PR #5592, opened 2025-10-30
Avoid modifying `node.measured` object, create copy instead

svilen-ivanov-kubit:
When working with immutable `node` data, modifying in-place `node.measured` properties is not allowed. Since we perform a shallow copy, we need to create a new `measured` object to populate `width` and `height`.

I'm using Immer to layout the nodes, and I started getting the following error after upgrade to v12:
```
Uncaught TypeError: Cannot assign to read only property 'width' of object '#<Object>'
    at applyChange (index.js:685:34)
    at applyChanges (index.js:646:13)
    at applyNodeChanges (index.js:735:12)
```

### DIFF
--- a/.changeset/dry-seahorses-deliver.md
+++ b/.changeset/dry-seahorses-deliver.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
+Always create a new measured object in apply changes.


svilen-ivanov-kubit:
I was considering the following fix:

```javascript
element.measured = {
  ...(element.measured ?? {}),
  width: change.dimensions.width,
  height: change.dimensions.height,
};
```

But `Node['measured']` does not allow any other properties to justify the spread.

moklick:
Thanks @svilen-ivanov-kubit ! Good finding :)

PR #5617, opened 2025-11-11
Allow getEdgeId to be overridden

### LINKED ISSUE #5180: Feature request: Allow getEdgeId to be overridden

Hi all, thanks for the great work on this project.

I have my own custom implementation of `getEdgeId` that I need to remember to call anytime I add a new edge or update an existing one.

`reconnectEdge` and `addEdge` both call the internal `getEdgeId`, but `addEdge` will only call it if `isEdgeBase` is false. 

I'd like to either allow `reconnectEdge` to respect the existing edge id, or to allow passing a custom `getEdgeId` function.

Happy to mock up a PR if that's preferable.

akre54:
Fixes #5180

### DIFF
--- a/.changeset/custom-edge-id-generator.md
+++ b/.changeset/custom-edge-id-generator.md
@@ -1,5 +1,5 @@
 ---
-'@xyflow/system': minor
+'@xyflow/system': patch
 ---
 
 Allow custom `getEdgeId` function in `addEdge` and `reconnectEdge` options to enable custom edge ID schemes.


.changeset/custom-edge-id-generator.md
```diff
@@ -0,0 +1,5 @@
+---
```
akre54, line 1: I'm positive this naming convention is wrong (not enough lobsters 🦞 ) but happy to rename

packages/system/src/utils/edges/general.ts
```diff
@@ -190,10 +217,12 @@ export const reconnectEdge = <EdgeType extends EdgeBase>(
     return edges;
   }
 
+  const edgeIdGenerator = options.getEdgeId || getEdgeId;
+
   // Remove old edge and create the new edge with parameters of old edge.
   const edge = {
     ...rest,
-    id: options.shouldReplaceId ? getEdgeId(newConnection) : oldEdgeId,
+    id: options.shouldReplaceId ? edgeIdGenerator(newConnection) : oldEdgeId,
```
akre54, line 225: Another option is to do similar to above and switch based on `isEdgeBase`

akre54:
@coderabbitai generate docstrings

moklick:
Thanks for your PR. We will have a look :)

PR #5626, opened 2025-11-16
Clarify screen position wording

cygnuson:
Clarify that screen positions for this util function are relative to the viewport.


cygnuson:
I spent more than a few minutes thinking I needed window.screenX/Y so I thought I'd potentially save others the time.

peterkogo:
I think calling it viewport might be a little confusing because we also call the React Flow pane a "viewport". The function has "screen" in its name which is also the least ambiguous term in this case.

PR #5629, opened 2025-11-16
Prevent re-render on every store update by using shallow in FlowRendererComponent useStore

AlaricBaraou:
While doing perf optimization I noticed that FlowRenderer seemed to re-render for no apparent reason.
It's because on every store update, the current selector creates a new object which don't pass the equality check of zustand.
I fixed it the same way as other components in the app, by passing shallow as the equalityFunction parameter.


moklick:
hey @AlaricBaraou ! Good finding 🙏 Thanks!

PR #5642, opened 2025-11-26
Performance Improvement for Node Updates: Direct Array Mutation in `updateNodeInternals`

bnku:
**Description:**

This PR addresses a significant delay encountered when adding or updating nodes in Svelte Flow, particularly in scenarios with a large number of existing nodes. Users reported a 3-5 second delay after performing an action that modifies a node before the changes became visible on the canvas.

**Problem Cause:**

During the debugging process, it was discovered that the primary cause of the delay was an inefficient node update mechanism within the `updateNodeInternals` function of Svelte Flow. Every time this function was called (e.g., when adding a new node or changing the dimensions of an existing one), it ultimately led to the creation of a *new* `store.nodes` array by using the `map` method.

While creating a new array might be acceptable for a small number of elements, it introduces the following issues when dealing with hundreds or thousands of nodes:

1.  **Excessive Reactivity:** Svelte (especially in the context of Svelte 5 with its signals) perceives replacing the entire array as a change requiring a full re-evaluation and potential re-rendering of all its elements.
2.  **Repetitive Costly Computations:** The `adoptUserNodes` function in `@xyflow/system`, which is triggered by changes to the `nodes` array, performs resource-intensive operations (such as position, dimension, and handle bound calculations) for *each* node in the array. When the entire array is replaced, these computations are executed for all nodes, even if only one has changed.

Logging demonstrated that the `adoptUserNodes` function was called multiple times, with each call taking hundreds of milliseconds (e.g., 500+ ms for 355 nodes), cumulatively resulting in delays of several seconds.

**Debugging Process:**

1.  **Data Flow Investigation:** The initial investigation focused on node rendering components (`NodeRenderer.svelte`, `NodeWrapper.svelte`) and store operations related to node updates (`updateNodeInternals` in `@xyflow/svelte` and `adoptUserNodes` in `@xyflow/system`).
2.  **Logging Instrumentation:** Timestamps (`console.time`/`console.timeEnd`) and messages (`console.log`) were added to key functions:
    *   `updateNodeInternals` in `packages/svelte/src/lib/store/index.ts`.
    *   `adoptUserNodes` in `packages/system/src/utils/store.ts`.
3.  **Performance Analysis:** After reproducing the issue with activated logging, the browser console clearly showed that `adoptUserNodes` was the primary bottleneck, consuming significant time with each node array update. It was observed that this function was called for all nodes, even when only one had changed.

**Proposed Fix and Why It Helped:**

The fix involves modifying the `updateNodeInternals` function in the `packages/svelte/src/lib/store/index.ts` file. Instead of creating a new `store.nodes` array using `map`, we now directly mutate the existing `store.nodes` array.

```diff
<<<<<<< SEARCH
:start_line:122
-------
    store.nodes = store.nodes.map((node) => newNodes.get(node.id) ?? node);
    console.log(`updateNodeInternals: updated ${newNodes.size} nodes`);
  }
=======
    // Instead of creating a new array, directly update the nodes in the existing array
    for (const [id, updatedNode] of newNodes.entries()) {
      const index = store.nodes.findIndex((node) => node.id === id);
      if (index !== -1) {
        store.nodes[index] = updatedNode;
      }
    }
    console.log(`updateNodeInternals: updated ${newNodes.size} nodes`);
  }
>>>>>>> REPLACE
```

This change significantly improves performance for the following reasons:

*   **Svelte Reactivity Optimization:** In Svelte 5, with its signal-based reactivity model (`$state`), directly mutating array elements or objects allows Svelte to track changes more precisely. Instead of a full re-evaluation of the entire array, Svelte can focus only on the nodes that have been directly modified.
*   **Reduced Costly Operations:** Since `store.nodes` is no longer completely replaced, Svelte Flow is not forced to re-execute `adoptUserNodes` for all nodes. This minimizes the overhead of position, dimension, and handle calculations that were previously performed redundantly.

As a result of this change, when adding or updating a node, Svelte Flow now processes only the changed node, leading to a significant reduction in delay and providing a smoother user experience.

### DIFF
--- a/packages/svelte/src/lib/store/index.ts
+++ b/packages/svelte/src/lib/store/index.ts
@@ -119,7 +119,12 @@ export function createStore<NodeType extends Node = Node, EdgeType extends Edge 
       newNodes.set(change.id, node);
     }
 
-    store.nodes = store.nodes.map((node) => newNodes.get(node.id) ?? node);
+    for (const [id, updatedNode] of newNodes.entries()) {
+      const index = store.nodes.findIndex((node) => node.id === id);
+      if (index !== -1) {
+        store.nodes[index] = updatedNode;
+      }
+    }
   }
 
   function fitView(options?: FitViewOptions) {
@@ -414,3 +419,4 @@ export function createStore<NodeType extends Node = Node, EdgeType extends Edge 
 
   return storeWithActions;
 }
+


peterkogo:
I might have to shower after reading through this mess of oozing AI slop goo code.

## Before
```diff
-  store.nodes = store.nodes.map((node) => newNodes.get(node.id) ?? node);
```
- Iterate over the nodes and using the `newNodes` map as a lookup
- Complexity of O(n) with very fast array iteration and map lookups
- Sometimes life can be simple and make you happy at the same time, you know?

## After
```diff
+  for (const [id, updatedNode] of newNodes.entries()) {
+    const index = store.nodes.findIndex((node) => node.id === id);
+    if (index !== -1) {
+      store.nodes[index] = updatedNode;
+    }
+  }
```
- Iterate over `newNodes` map (!!!) and for each entry iterate over every node to see where it is in the array
- Update the node in-place and break reactivity
- Complexity of O(n^2) and using iterators and slow non-continuous array access
- Write long paragraphs over how **Excessive Reactivity** and **Repetitive Costly Computations** is ruining performance

PR #5643, opened 2025-11-26
Generate provenance statement on release to increase security

tnkuehne:
Following recent hacks on npm packages, it would be greatly appreciated if you could increase the trust level of the npm packages.

I added generation of provenance statements as outlined in this guide from npm: https://docs.npmjs.com/generating-provenance-statements


moklick:
Thanks for the PR. We will have a look.

moklick:
Thanks again! We will try this with the upcoming release.

PR #5682, opened 2026-01-22
Prevent unnecessary updates when selectNodesOnDrag = false

artemtam:
Updated `unselectNodesAndEdges` so it changes only those nodes and edges that are selected and skips those that are unselected. This prevents re-rendering of all nodes and edges on drag start when `selectNodesOnDrag` is set to `false`.

It fixes this performance issue – https://github.com/xyflow/xyflow/issues/1876#issuecomment-2778181736:

https://github.com/user-attachments/assets/af157363-0128-4fd3-8dfd-a27d7ce7265f

### DIFF
--- a/packages/react/src/store/index.ts
+++ b/packages/react/src/store/index.ts
@@ -307,12 +307,12 @@ const createStore = ({
 
         const nodeChanges: NodeSelectionChange[] = [];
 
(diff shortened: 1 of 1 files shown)


moklick:
Thanks @artemtam ! We will try it out.

PR #5684, opened 2026-01-23
fix: ensure drag handlers update when nodes selection changes programmatically

### LINKED ISSUE #4841: Rectangular selection blocks dragging new preselected nodes

- React Flow / Svelte Flow version: 12.3.5
- Browser and version: Firefox 132.0.2 (64-bit)
- OS and version: Windows 11 Pro 23H2

### Live code example

https://codesandbox.io/p/sandbox/upbeat-vaughan-6c4m56

### Describe the Bug

When an existing node (or nodes) is already selected via rectangular selection (i.e. holding shift and dragging the mouse to select nodes), adding new nodes with `selected: true` and setting `selected: false` on the old nodes causes the new nodes to be undraggable until they are reselected.

1. Rectangle select a node
2. Click the button
3. Attempt to drag the new nodes

### Expected behavior

I expected to be able to drag the newly created nodes immediately, but instead I am not able to drag the new nodes until I reselect them.

### Screenshots or Videos

https://github.com/user-attachments/assets/7f47b6d4-8b36-486d-81c3-171e3b74bc39

### Additional context

It looks like `NodesSelection` is rendered over the new nodes, but [`nodeRef`]`.current` is still `null` when `useDrag`'s effect runs to setup event handlers, so `NodesSelection` gets the mouse events and does nothing with them. I can work around this issue by calling `useStoreApi().setState({ nodesSelectionActive: false })` when the new nodes are added, but reaching into internals like that feels wrong.

[`nodeRef`]: https://github.com/xyflow/xyflow/blob/1e38733da5fc6591a787a01fa0a9853e9ee4339f/packages/react/src/components/NodesSelection/index.tsx#L44

I ran into this while implementing a function that duplicates the currently selected nodes. In that case, it's desirable to deselect the old nodes and select the new nodes so that the new nodes can be moved immediately after duplication.

ysds:
Fixes #4841

## Problem

When a selection rectangle was already displayed (e.g., after box selection), and then nodes were programmatically selected (e.g., adding a new node with `selected: true`), the selection rectangle would move to the new node's position but dragging it would not work.

## Root Cause

The previous implementation used `useEffect` to set up drag handlers, but when `nodeRef.current` became available after the initial render, the drag handlers were not properly initialized.

## Solution

- Replaced `useEffect` with `useIsomorphicLayoutEffect` to ensure drag handlers are set up immediately when the DOM element becomes available
- Added `disabled` parameter to `useDrag` to allow external control
- `NodesSelection` now passes `disabled: !shouldRender` to properly disable drag handlers when not rendered

## How to Test

You can test this fix using the UseNodesInit example. Update the `addNode` function to programmatically select the new node:

```typescript
const addNode = () =>
  setNodes((nodes) => [
    ...nodes.map((node) => ({ ...node, selected: false })),
    {
      id: `${Math.random()}`,
      data: { label: 'new node' },
      position: { x: Math.random() * 400, y: Math.random() * 400 },
      selected: true,
    },
  ]);
```

1. Perform a box selection (Shift + drag) → Selection rectangle should be draggable
2. Click "add node" button → New node is added with selected: true
3. Try to drag the selection rectangle → Should work correctly (this was broken before)

## Note

While working on this fix, I noticed that `nodeClickDistance` is missing from the effect dependencies array. Changes to this parameter won't trigger drag handler updates. This might be worth addressing in a follow-up PR.

### DIFF
--- a/packages/react/src/hooks/useDrag.ts
+++ b/packages/react/src/hooks/useDrag.ts
@@ -3,7 +3,6 @@ import { XYDrag, type XYDragInstance } from '@xyflow/system';
 
 import { handleNodeClick } from '../components/Nodes/utils';
 import { useStoreApi } from './useStore';
-import { useIsomorphicLayoutEffect } from './useIsomorphicLayoutEffect';
 
 type UseDragParams = {
   nodeRef: RefObject<HTMLDivElement>;
@@ -52,13 +51,8 @@ export function useDrag({
     });
   }, []);
 
-  useIsomorphicLayoutEffect(() => {
-    if (disabled) {
-      xyDrag.current?.destroy();
-      return;
-    }
-
-    if (!nodeRef.current || !xyDrag.current) {
+  useEffect(() => {
+    if (disabled || !nodeRef.current || !xyDrag.current) {
       return;
     }
 


moklick:
Thanks again for your help here @ysds - we will have a look 🙏

peterkogo:
@ysds Thanks for taking the time to investigate this! Especially finding the missing dependency....

Unfortunately, your fix does not seem to influence the bug for me. I have pushed a reproduction of the bug as an example (you can test it out by running `pnpm dev` and visiting `http://localhost:3000/examples/node-selection-bug`.

Might I be missing something?

Intuitively, I would have assumed the bug is caused by the way we handle selections. Once the selection box is drawn, explicit deselect or other select events need to happen to deactivate it again – otherwise it simply assumes all selected nodes to be part of the selection box. Need to investigate, though.

ysds:
@peterkogo 
Thanks for taking a look. In my environment (macOS 26.1, Chrome 144.0.7559.133), the issue seems to be resolved. However, I’m sorry that I haven’t been able to verify the behavior on other environments yet.



https://github.com/user-attachments/assets/e8228861-5ee8-4785-8827-ce41440327b3

peterkogo:
I'm sorry, I have misread the issue description! The fix does indeed seem to work.

peterkogo:
The isomorphicLayoutEffect is not needed in this case and I simplified the code a bit.

peterkogo:
@ysds Was there a specific reason you opted to use a `useLayoutEffect` instead of just and `useEffect`?

ysds:
My thinking was that since `NodesSelection` conditionally renders based on `shouldRender`, and `nodeRef.current` only becomes available after the DOM mounts, using `useLayoutEffect` would set up drag handlers synchronously before the browser paints — eliminating a theoretical single-frame gap where the selection box is visible but not yet draggable.
But in practice, that gap is imperceptible, so `useEffect` is sufficient here. Thanks for simplifying it!

moklick:
Thanks @ysds and @peterkogo

PR #5687, opened 2026-01-26
Optimize zooming when panOnScroll is enabled

### LINKED ISSUE #5683: Pinch-to-zoom lags in Safari if panOnScroll is enabled

- React Flow / Svelte Flow version: React Flow 12.10.0
- Browser and version: Safari 26.2
- OS and version: macOS Tahoe 26.2

### Live code example

https://codesandbox.io/p/devbox/fast-bird-xnskv8?workspaceId=ws_CGqKnD74e4ahgSLVoKKP1

### Describe the Bug

If `panOnScroll` is enabled then pinch-to-zoom lags a lot in Safari. But if it is disabled then it is really smooth.

1. Add a lot of nodes
2. Enable `panOnScroll`

### Expected behavior

Smooth zooming :)

### Screenshots or Videos

`panOnScroll` is enabled:

https://github.com/user-attachments/assets/3b69d195-c814-447c-bc45-23954784d268

`panOnScroll` is disabled:

https://github.com/user-attachments/assets/298eb6d4-30c8-4a69-840d-2db7ebbfc21e

### Additional context

~I suspect that the problem lies here: https://github.com/xyflow/xyflow/blob/cf2d4be3f57047ae02d08eed49dc640e89a2ef14/packages/system/src/xypanzoom/eventhandler.ts#L91~

vkrol:
It partially fixes #5683.

`store.setState({ paneDragging })` triggers all selector functions to run for each wheel event handler when pinch-to-zoom is used. We can avoid this by passing the updater function to setState and returning the previous state if nothing needs to be updated. 

Before:

https://github.com/user-attachments/assets/d85b1cb0-0386-4209-b4b2-50b3065ab13e

After:

https://github.com/user-attachments/assets/81a2d69e-00c4-4de7-9d1d-796ac8fba903

### DIFF
--- a/packages/react/src/container/ZoomPane/index.tsx
+++ b/packages/react/src/container/ZoomPane/index.tsx
@@ -74,12 +74,7 @@ export function ZoomPane({
         translateExtent,
         viewport: defaultViewport,
         onDraggingChange: (paneDragging) =>
-          store.setState((prevState) => {
-            if (prevState.paneDragging !== paneDragging) {
-              return { paneDragging };
-            }
-            return prevState;
-          }),
+          store.setState((prevState) => prevState.paneDragging === paneDragging ? prevState : { paneDragging }),
         onPanZoomStart: (event, vp) => {
           const { onViewportChangeStart, onMoveStart } = store.getState();
           onMoveStart?.(event, vp);


moklick:
This looks good! Thanks for your PR we will check it.

PR #5688, opened 2026-01-26
handle undefined node in NodeComponentWrapperInner

### LINKED ISSUE #5651: Error thrown on nodeLookup access when conditionally rendering a React Flow instance

- React Flow / Svelte Flow version: 12.10.0
- Browser and version: Chrome version 
- OS and version: Windows 11 Pro OS Build 26100.7171

### Live code example

https://github.com/valerio-guarnieri-one/xyflow-issue

### Describe the Bug

React Flow is throwing the following exception: `Exception has occurred: TypeError: Cannot destructure property 'internals' of 's.nodeLookup.get(...)' as it is undefined` when turning a state variable used to conditionally render a React Flow instance to false.

1. Download the [example repository](https://github.com/valerio-guarnieri-one/xyflow-issue)
2. Run `npm run dev` command
3. Open VS Code
4. Go to the "Run and Debug" sidebar button
5. Start Debugging (press F5)
6. Turn on Caught Exceptions in the Breakpoints section
7. In Chrome, turn off the switch in the top right corner
8. VS Code debugger will break on the following error thrown:
```
Exception has occurred: TypeError: Cannot destructure property 'internals' of 's.nodeLookup.get(...)' as it is undefined.
at http://localhost:5173/node_modules/.vite/deps/@xyflow_react.js?v=5776e29c:9368:13
    at memoizedSelector (http://localhost:5173/node_modules/.vite/deps/@xyflow_react.js?v=5776e29c:123:35)
    at http://localhost:5173/node_modules/.vite/deps/@xyflow_react.js?v=5776e29c:132:24
    at checkIfSnapshotChanged (http://localhost:5173/node_modules/.vite/deps/react-dom_client.js?v=3f65e72c:6095:27)
    at http://localhost:5173/node_modules/.vite/deps/react-dom_client.js?v=3f65e72c:6088:11
    at http://localhost:5173/node_modules/.vite/deps/@xyflow_react.js?v=5776e29c:6269:39
    at Set.forEach (<anonymous>)
    at setState (http://localhost:5173/node_modules/.vite/deps/@xyflow_react.js?v=5776e29c:6269:17)
    at reset (http://localhost:5173/node_modules/.vite/deps/@xyflow_react.js?v=5776e29c:8981:18)
    at http://localhost:5173/node_modules/.vite/deps/@xyflow_react.js?v=5776e29c:6543:7
```
at line 4493 in node_modules\@xyflow\react\dist\esm\index.js (`NodeComponentWrapperInner` function)

### Expected behavior

As a user, I expected no error being thrown.

### Screenshots or Videos

_No response_

### Additional context

The error appears only if a MiniMap component is present in the React Flow instance. If the MiniMap component is removed, no error occurs when turning off the switch to set the state variable to false.

filippo-fanni:
This PR attempts to fix #5651.
It seems that when ReactFlow unmounts, the store gets reset and nodeLookup is cleared. The MiniMap selector appears to still run one more time against the empty lookup, and the non-null assertion on nodeLookup.get(id)! throws.
The bug was likely already present, recent changes probably just shifted timing enough to expose it. (maybe the rerender optimization in #5629 or maybe some dependency update)

The fix:
Added optional chaining and an early return with safe defaults in the selector. The existing if (!node) guard already handles this case, it just wasn't being reached because the selector threw first.

### DIFF
--- a/packages/react/src/additional-components/MiniMap/MiniMapNodes.tsx
+++ b/packages/react/src/additional-components/MiniMap/MiniMapNodes.tsx
@@ -84,7 +84,10 @@ function NodeComponentWrapperInner<NodeType extends Node>({
   shapeRendering: string;
 }) {
   const { node, x, y, width, height } = useStore((s) => {
-    const { internals } = s.nodeLookup.get(id)!;
+    const internals = s.nodeLookup.get(id)?.internals;
+    if (!internals) {
+      return { node: undefined, x: 0, y: 0, width: 0, height: 0 };
+    }
     const node = internals.userNode as NodeType;
     const { x, y } = internals.positionAbsolute;
     const { width, height } = getNodeDimensions(node);


moklick:
Thanks @filippo-fanni !

Issue #5552, opened 2025-10-16
Add center-based (symmetric) resize mode to NodeResize component


bogy0:
### Problem
Currently, the NodeResize component in React Flow supports resizing nodes by dragging their edges or corners, but the resize behavior always expands or contracts from the dragged side. There is no built-in option (or at least im not aware, if yes sorry) to maintain the node’s center point while resizing symmetrically in both directions.

### Feature Request
Add a center-based resize mode (also known as symmetric resize) to the NodeResize component.
When enabled, dragging one edge (e.g., the left border) would resize the node equally in both directions — expanding or shrinking around its center point, keeping the node’s center coordinates fixed.

Example usage:
`<NodeResize symmetricResize />`

Maybe this can be even combined with an optional `axis` prop:
`<NodeResize symmetricResize axis="x" | "y" | "both" />`

### Use Case
In many diagramming tools (e.g., Figma, Lucidchart, draw.io), users expect an option to resize shapes symmetrically around their center.
This is particularly useful for:
- Keeping nodes visually aligned when resizing.
- Adjusting proportions of centered layouts or grid-based diagrams.
- Maintaining consistent relationships between nodes during dynamic layout adjustments.

### Additional Notes
- Could be implemented by adjusting x/y position offset inversely to the resize delta.
- Maybe support keyboard modifier (e.g., Alt/Option key for temporary symmetric resizing).

moklick:
@peterkogo do you have any thoughts about this?

peterkogo:
Its definitely a feature of most resizers, sounds like a good idea!

PR #5693, opened 2026-01-27
feat(node-resizer): add shift aspect lock and alt for symmetric resize

180107072:
Adds shift to lock aspect ratio (corner controls) and alt‑key symmetric resize for XYResizer (feature request #5552).


peterkogo:
Thanks for your contribution! It has helped us better grasp the viability of such a feature and how a possible implementation approach might look like.
We don't want any hardcoded key listeners controlling this behavior and there are a few more additional edge cases that need to be respected.
Because there are quite a few changes that need to be made here, I have created a separate branch & PR for this (#5700) – using your changes as a baseline. I will close this PR in favor of that.

PR #5703, opened 2026-02-10
Improve return type of useNodesData

### LINKED ISSUE #5657: useNodesData type not narrowed down properly

- React Flow version: 12.10.0
- Browser and version: Chromium 143.0.7499.40
- OS and version: Arch Linux
- Typescript 5.9.3; node 22.14

### Live code example

https://reactflow.dev/learn/advanced-use/typescript#hooks

### Describe the Bug

The `useNodesData` [example](https://reactflow.dev/learn/advanced-use/typescript#hooks) using a custom node type doesn't actually work due to the return type using `Pick<..., 'id' | 'type' | 'data'>`, which is [not distributive](https://github.com/microsoft/TypeScript/issues/28339). This results in the returned item being of type
```ts
{ type: 'a' | 'b'; data: A | B; id: string }
```
instead of
```ts
{ type: 'a'; data: A; id: string }
| { type: 'b'; data: B; id: string }
```

Just copy the example from [the docs](https://reactflow.dev/learn/advanced-use/typescript#hooks):
```ts
const nodesData = useNodesData<CustomNodeType>(connections?.[0].source);

nodeData.forEach(({ type, data }) => {
  if (type === 'number') {
    // This is type safe because we have narrowed down the type
    console.log(data.number);
  }
});
```

### Expected behavior

The node type should actually be narrowed down.

### Screenshots or Videos

_No response_

### Additional context

The linked TypeScript issue has a few options in the comments, one that worked for me was
```ts
type DistributivePick<T, K extends keyof T> = T extends unknown ? Pick<T, K> : never;
```
which I'm currently using for a workaround like this
```ts
const nodes = useNodesData<CustomNode>(nodeIds) as DistributivePick<CustomNode, 'id' | 'type' | 'data'>[];
```

peterkogo:
closes #5657

### DIFF
--- a/.changeset/old-laws-melt.md
+++ b/.changeset/old-laws-melt.md
@@ -1,0 +1,8 @@
+---
+'@xyflow/react': patch
+'@xyflow/svelte': patch
+'@xyflow/system': patch
+---
+
+Improve return type of useNodesData. Now you can narrow down the data type by checking the node type.  
+
--- a/examples/react/src/examples/UseNodesData/TextNode.tsx
+++ b/examples/react/src/examples/UseNodesData/TextNode.tsx
@@ -1,7 +1,8 @@
 import { memo, useState } from 'react';
 import { Position, NodeProps, Handle, useReactFlow } from '@xyflow/react';
+import type { TextNode } from '.';
 
-function TextNode({ id, data }: NodeProps) {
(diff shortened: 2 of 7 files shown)


moklick:
cool :)

PR #5707, opened 2026-02-16
fix(system): fire onNodeDrag even when nodeExtent clamps position

### LINKED ISSUE #5706: Bug: onNodeDrag not called when nodeExtent clamps position on both axes
## Describe the Bug

When using the `nodeExtent` prop, `onNodeDrag` stops firing if the node's position is clamped on **both X and Y axes simultaneously** (e.g., when a node is dragged into a corner of the extent boundary).

The user is still actively dragging the mouse — the cursor moves, the drag session is alive — but `onNodeDrag` receives no calls.

## Your Example Website or App

**Live reproduction:** https://xyflow-nodeextent-bug.vercel.app

**Source:** https://github.com/jiwonMe/xyflow-nodeextent-bug

## Steps to Reproduce

1. Set `nodeExtent={[[0, 0], [500, 500]]}` on a `<ReactFlow>` component
2. Place a node near the bottom-right corner `(490, 490)`
3. Drag the node toward the corner (past `(500, 500)`)
4. While holding the mouse button, continue moving the cursor further right/down

**Expected:** `onNodeDrag` continues to fire on every mouse move (with the clamped position).

**Actual:** `onNodeDrag` stops firing entirely once the position is clamped on both axes.

## Expected Behavior

`onNodeDrag` should always fire during an active drag operation, regardless of whether the final position was clamped by `nodeExtent`. The event object's `clientX`/`clientY` still changes, and consumers rely on it for:

- Cross-boundary drag detection (is the cursor outside a container?)
- Custom ghost/preview rendering at cursor position
- Drop target detection based on cursor position

This is consistent with how `onNodeDragStart` and `onNodeDragStop` always fire regardless of position changes.

## Root Cause

In [`packages/system/src/xydrag/XYDrag.ts`](https://github.com/xyflow/xyflow/blob/main/packages/system/src/xydrag/XYDrag.ts), the `updateNodes()` function:

```typescript
// Line ~192
hasChange = hasChange || dragItem.position.x !== position.x || dragItem.position.y !== position.y;

dragItem.position = position;
dragItem.internals.positionAbsolute = positionAbsolute;

// ...

// Line ~197
if (!hasChange) {
  return;  // ← Early return skips onNodeDrag callback
}
```

When `calculateNodePosition()` returns the same clamped coordinates as the previous frame, `hasChange` remains `false`, causing an early return that skips both `updateNodePositions()` AND `onDrag`/`onNodeDrag`/`onSelectionDrag` callbacks.

The `updateNodePositions` skip is correct (no state to update), but the callback skip is a bug — the drag event itself is still happening.

## Proposed Fix

Separate the `hasChange` guard so it only gates `updateNodePositions`, while `onNodeDrag` always fires during an active drag:

```typescript
if (hasChange) {
  updateNodePositions(dragItems, true);
}

// Always fire drag callbacks during active drag
if (dragEvent && (onDrag || onNodeDrag || (!nodeId && onSelectionDrag))) {
  const [currentNode, currentNodes] = getEventHandlerParams({
    nodeId,
    dragItems,
    nodeLookup,
  });

  onDrag?.(dragEvent, dragItems, currentNode, currentNodes);
  onNodeDrag?.(dragEvent, currentNode, currentNodes);

  if (!nodeId) {
    onSelectionDrag?.(dragEvent, currentNodes);
  }
}
```

I'm happy to submit a PR with this fix + tests if the team agrees with the approach.

## Screenshots

_N/A — behavioral bug, no visual artifact._

## Platform

- Browser: Chrome 131, Safari 18
- OS: macOS
- @xyflow/react: 12.6.4
- @xyflow/system: 0.0.52

jiwonMe:
## Summary

Fixes #5706

**Live reproduction:** https://xyflow-nodeextent-bug.vercel.app
**Source:** https://github.com/jiwonMe/xyflow-nodeextent-bug

When `nodeExtent` clamps a dragged node's position on both axes simultaneously (e.g., dragging into a corner), `onNodeDrag` stops firing entirely. This is because the `hasChange` early return in `updateNodes()` gates both the position update AND the drag callbacks.

## The Problem

In `XYDrag.ts`, `updateNodes()`:

```typescript
// Before (bug)
if (!hasChange) {
  return; // skips BOTH updateNodePositions AND onNodeDrag
}

updateNodePositions(dragItems, true);
onNodeDrag?.(dragEvent, currentNode, currentNodes); // never reached
```

When `calculateNodePosition()` returns the same clamped coordinates as the previous frame, `hasChange` stays `false` → early return → `onNodeDrag` never fires.

## The Fix

Split the guard so `updateNodePositions` is correctly skipped when there is no position change, but drag callbacks always fire during an active drag:

```typescript
// After (fix)
if (hasChange) {
  updateNodePositions(dragItems, true);
}

// Always fire - the drag event is still happening
onNodeDrag?.(dragEvent, currentNode, currentNodes);
```

This is a 3-line change.

## Why This Matters

Consumers use `onNodeDrag` for more than just position tracking:

- **Cross-boundary detection**: Is the cursor outside a container? (nested ReactFlow use case)
- **Ghost/preview rendering**: Rendering a drag preview at cursor position
- **Drop target detection**: Highlighting drop zones based on cursor, not node position

These all depend on the `event.clientX/clientY` from the MouseEvent, which keeps changing even when the node position is clamped.

## Consistency

This aligns with how `onNodeDragStop` already works - it fires regardless of whether positions changed (guarded by `nodePositionsChanged`, not `hasChange`).

## Testing

- Build passes (`pnpm build`)
- Existing Playwright tests: same results as `main` (43 pre-existing webkit failures unrelated to this change)

### DIFF
--- a/.changeset/fix-node-drag-on-clamped-extent.md
+++ b/.changeset/fix-node-drag-on-clamped-extent.md
@@ -1,0 +1,7 @@
+---
+'@xyflow/system': patch
+'@xyflow/react': patch
+'@xyflow/svelte': patch
+---
+
+Fix `onNodeDrag` not being called when `nodeExtent` clamps position on both axes
--- a/packages/system/src/xydrag/XYDrag.ts
+++ b/packages/system/src/xydrag/XYDrag.ts
@@ -197,11 +197,9 @@ export function XYDrag<OnNodeDrag extends (e: any, nodes: any, node: any) => voi
 
       nodePositionsChanged = nodePositionsChanged || hasChange;
 
-      if (!hasChange) {
-        return;
+      if (hasChange) {
+        updateNodePositions(dragItems, true);
       }
-
-      updateNodePositions(dragItems, true);
 
       if (dragEvent && (onDrag || onNodeDrag || (!nodeId && onSelectionDrag))) {
         const [currentNode, currentNodes] = getEventHandlerParams({


jiwonMe:
Closing this PR after reconsideration. The current behavior appears to be intentional — the comment in the source (`// we want to make sure that we only fire a change event when there is a change`) suggests `onNodeDrag` is designed to fire only when the node position actually changes, not on every mouse move. Our use case (cross-boundary detection in nested ReactFlow) is better served by a separate `mousemove` listener. Sorry for the noise!

PR #5708, opened 2026-02-16
fix(react): allow pointer events on viewport and nodes for interactive custom nodes

cap-jmk-real:
## Problem

When using custom nodes with interactive content (inputs, selects, buttons), the default library CSS sets `.react-flow__viewport` and `.react-flow__nodes` to `pointer-events: none`. Hit-testing then falls through to the pane, which causes:

- The pane/viewport to receive hover and click instead of the node
- Dropdowns and inputs inside nodes to be hard or impossible to click
- The cursor to show grab over the whole canvas instead of default/pointer over controls

## Solution

In the shared style source used by `@xyflow/react` (`packages/system/src/styles/init.css`), both selectors are set to `pointer-events: auto`:

- **`.xy-flow__viewport`** → `pointer-events: auto` — so the viewport can receive events for panning and the empty canvas keeps the grab cursor
- **`.xy-flow__nodes`** → `pointer-events: auto` — so nodes and their children (inputs, buttons, etc.) receive clicks and hover correctly

## Backward compatibility

Apps that depend on the previous behavior can override back to `pointer-events: none` on these classes in their own CSS.

### DIFF
--- a/packages/system/src/styles/init.css
+++ b/packages/system/src/styles/init.css
@@ -85,7 +85,7 @@
 .xy-flow__viewport {
   transform-origin: 0 0;
   z-index: 2;
-  pointer-events: none;
+  pointer-events: auto;
 }
 
 .xy-flow__renderer {
@@ -191,7 +191,7 @@ svg.xy-flow__connectionline {
 }
 
 .xy-flow__nodes {
-  pointer-events: none;
+  pointer-events: auto;
   transform-origin: 0 0;
 }
 


moklick:
Hey @cap-jmk-real . Thanks for your contribution but we can't merge this because it would be a breaking change. If you want to display interactive elements inside your custom nodes, you need to add the ["nodrag" class name](https://reactflow.dev/learn/customization/utility-classes#nodrag).

cap-jmk-real:
Hey no worries, if I patch it for my app it works well without braking anything, however the ["nodrag" class name](https://reactflow.dev/learn/customization/utility-classes#nodrag), does not do the job and the cursors get confused with the panning cursor and clicking on interactive elements does not reliably work. 

In the end its your decision, as I can just leave the patch for my app. But its a simple fix maybe worth considering.

moklick:
@cap-jmk-real could you create a codesandbox or do you have a repo I could check out? Maybe I don't understand the issue here.

cap-jmk-real:
No worries. Reproducing the error in a sandbox is rather difficult. You can try with the main of my [repo](https://github.com/cap-jmk-real/agentron.git). You launch from the root

`npm run dev:ui`

Then you can click on agents->visual then you can see the bug. I'll let you know when I can reproduce the bug in the sandbox.

PR #5722, opened 2026-03-09
fix: screenToFlowPosition add snapGrid types

dfblhmm:
# Changes

fix screenToFlowPosition type Definition, add `snapGrid` params in the options

# Summary

when I use `useReactFlow().screenToFlowPosition`, It give me a type error, Refer **useViewportHelper.ts**
image


moklick:
Thanks!

PR #5730, opened 2026-03-19
Fix missing border radius on smooth step edge bends

### LINKED ISSUE #5705: Not all edge breaks are rounded

- React Flow version: `"@xyflow/react": "^12.10.0"`
- Browser and version: Chrome latest stable
- OS and version: Mac OS

### Live code example

_No response_

### Describe the Bug

Hello

We are using custom edge component
where we render edge using `getSmoothStepPath`:
```
const [edgePath] = getSmoothStepPath({
  sourceX,
  sourceY,
  targetX,
  targetY,
  stepPosition: 1, // edge break closer to target node
  borderRadius: 7,
})
```
The end result looks like this:
Image

Is this intentional or this last line break should be also rounded?

Render custom edge using `getSmoothStepPath` like described above.

### Expected behavior

All edge line breaks are rounded without exceptions.

### Screenshots or Videos

_No response_

### Additional context

_No response_

justn-hyeok:
Fixes #5705

## Problem

When using `getSmoothStepPath` with extreme `stepPosition` values (e.g. `1` or `0`), some edge bends are rendered with sharp corners instead of rounded ones.

## Root Cause

Extreme `stepPosition` values cause intermediate path points to overlap with the gapped source/target points. When `getBend` receives consecutive duplicate points, `distance(a, b)` becomes 0, making `bendSize` 0 — resulting in no rounding.

## Fix

Filter out consecutive duplicate points before building the SVG path.

## Before / After

before

after

### DIFF
--- a/packages/system/src/utils/edges/smoothstep-edge.ts
+++ b/packages/system/src/utils/edges/smoothstep-edge.ts
@@ -107,12 +107,12 @@ function getPoints({
   if (sourceDir[dirAccessor] * targetDir[dirAccessor] === -1) {
     if (dirAccessor === 'x') {
       // Primary direction is horizontal, so stepPosition affects X coordinate
-      centerX = center.x ?? (sourceGapped.x + (targetGapped.x - sourceGapped.x) * stepPosition);
+      centerX = center.x ?? sourceGapped.x + (targetGapped.x - sourceGapped.x) * stepPosition;
       centerY = center.y ?? (sourceGapped.y + targetGapped.y) / 2;
     } else {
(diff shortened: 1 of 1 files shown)


moklick:
Thanks for your PR! I cleaned up the code a bit so that points get filtered out before so that we don't need to go over all points twice. Could you double check if it still works as expected @justn-hyeok ?

justn-hyeok:
Looks awesome! I tested with various 'stepPosition' values (0, 0.5, 1) and different handle positions - all bends are rounded correctly.

moklick:
cool! Thanks for your help here :)

PR #5733, opened 2026-03-20
fix(react): prevent empty store during ReactFlow remount

AlaricBaraou:
## Problem

When `<ReactFlow>` unmounts, `StoreUpdater`'s cleanup calls `reset()` which empties the store (nodes, edges, nodeLookup, etc.). On the subsequent mount, custom node effects that read from the store see empty data because `StoreUpdater` hasn't repopulated it yet.

Remounts can happen in real applications for many reasons — conditional rendering changes in ancestor components, React Suspense boundaries, HMR during development, route transitions, or parent components restructuring their tree. Making the store reliably available after a remount prevents subtle bugs in custom nodes that read store data in effects.

This only affects remounts, not initial mounts — on initial mount the store is empty but GraphView renders no nodes (no data yet), so no custom node effects fire against empty state. On remount, custom nodes render during the render phase using pre-reset data, so their effects exist and fire during the commit phase when the store has already been emptied by `reset()`.

The root cause was that `StoreUpdater` was rendered **after** `GraphView` in the JSX tree. React runs effects depth-first, so `GraphView`'s subtree effects (where custom nodes live) fired before `StoreUpdater`'s effect repopulated the store.

## Fix

1. **Move `StoreUpdater` before `GraphView`** in the JSX tree so its effects run before any custom node effects.
2. **Switch from `useEffect` to `useIsomorphicLayoutEffect`** so the store is repopulated synchronously before any child `useEffect` or `useLayoutEffect` fires.

The reorder is the primary fix — it ensures correct effect ordering within the same effect type. The layout effect switch provides an additional guarantee that the store is always populated before any child effect type fires.

## Test plan

- Added Cypress component test (`storeAvailabilityOnRemount.cy.tsx`) that verifies custom nodes see populated store data after a remount
- Tests cover both `useEffect` and `useLayoutEffect` consumers inside custom nodes
- Tests cover both controlled (`nodes`/`edges` props) and uncontrolled (`defaultNodes`/`defaultEdges`) usage
- Verified tests fail without the fix (2/3 fail) and pass with it (3/3 pass)
- Verified existing component tests pass with no regressions

### DIFF
--- a/.changeset/fix-store-reset-timing.md
+++ b/.changeset/fix-store-reset-timing.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
+Fix empty store during ReactFlow remount by reordering StoreUpdater before GraphView and using layout effects
--- a/examples/react/cypress/components/reactflow/storeAvailabilityOnRemount.cy.tsx
+++ b/examples/react/cypress/components/reactflow/storeAvailabilityOnRemount.cy.tsx
@@ -49,7 +49,9 @@ describe('<ReactFlow /> remount: store availability', () => {
     );
 
     cy.get('@spy').should('have.been.calledWithMatch', { lookupSize: 2 });
+    cy.then(() => spy.resetHistory());
     cy.get('[data-testid="remount"]').click();
+    cy.get('@spy').should('have.been.called');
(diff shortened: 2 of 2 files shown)


moklick:
Thanks for your PR @AlaricBaraou ! I will have a look.

PR #5735, opened 2026-03-25
Explicitly allow missing `type` field in BuiltInNode type definition

nvie:
This improves the DX in cases like this:

```ts
const initialNodes: BuiltInNode[] = [
//                  ~~~~~~~~~~~~~
//                   TS error here currently, because these nodes
//                   don't explicitly set `type: "default"`
  { id: "1", position: { x: 0, y: 0 }, data: { label: "Node 1" } },
  { id: "2", position: { x: 100, y: 100 }, data: { label: "Node 2" } },
];
```


moklick:
awesome!

PR #5750, opened 2026-04-12
Keep handle interactive when `isConnectableEnd` is false

devkcud:
## Summary

- Fix handle with `isConnectableEnd={false}` becoming permanently unusable after a connection starts and ends on the same handle
- Add missing `isConnectableStart` guard in Svelte Handle's `onpointerdown` for React parity

## Problem

When a handle has `isConnectableEnd={false}` and a user drags a connection that starts and ends on that same handle:

1. The drag cancels, then `click` fires on the same element
2. `clickConnectStartHandle` gets set to this handle
3. The `connectionindicator` class ternary switches to evaluate `isConnectableEnd` (which is `false`)
4. Handle loses `pointer-events: all` from CSS and becomes permanently unclickable
5. `clickConnectStartHandle` can only be cleared via click on a handle, but this handle is now dead

## Fix

The `connectionindicator` class now stays `true` for the click-connect source handle, so it retains `pointer-events: all` even when `isConnectableEnd={false}`.

**Svelte:** Added `isClickConnectSource` derived value used in `connectionindicator` expression.
**React:** Reused existing `clickConnecting` selector in `connectionindicator` expression (same bug existed).

Also added `isConnectableStart` check in Svelte's `onpointerdown` since React already had this guard.

### DIFF
--- a/examples/svelte/src/components/Header/Header.svelte
+++ b/examples/svelte/src/components/Header/Header.svelte
@@ -21,6 +21,7 @@
 		'node-toolbar',
 		'node-resizer',
 		'overview',
+		'self-attach',
 		'stress',
 		'subflows',
 		'two-way-viewport',
--- a/examples/svelte/src/routes/examples/self-attach/+page.svelte
+++ b/examples/svelte/src/routes/examples/self-attach/+page.svelte
@@ -1,0 +1,37 @@
+<script lang="ts">
+	import {
+		SvelteFlow,
+		Controls,
+		Background,
+		BackgroundVariant,
+		MiniMap,
+		type Node,
+		type NodeTypes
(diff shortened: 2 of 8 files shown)


moklick:
Thanks for the PR! There seems to be a bug around `isConnectableEnd` but it's still a bit buggy with this solution. When I start and end on the same handle and then try to connect from another handle it shows the crosshair cursor even if it's not possible to create a connection.

devkcud:
hey @moklick! sorry for the late reply

i'm still taking a look at these errors btw. i found another bug related to double clicking `isConnectableEnd` as soon as you let go of the connection, which can make it unusable as well (but i can't quite figure it out how to make it reproducible lol)

working on a small fix so the pr doesn't get huge

PR #5755, opened 2026-04-17
fix(react): add @types/react to peer dependencies

### LINKED ISSUE #5738: `@xyflow/react` fails with pnpm strict mode — missing peerDependency on `@types/react`
`@xyflow/react@12.10.1` ships `.d.ts` files that import from `react`:

```ts
import { ComponentType } from 'react';
```

TypeScript resolves these types through `@types/react`, but `@xyflow/react` only declares `react` and `react-dom` as peer dependencies — not `@types/react`. This works in flat `node_modules` layouts (npm, yarn) because `@types/react` is hoisted from the consuming project. When pnpm hoisting is disabled (`hoist: false`), TypeScript cannot find the React types and fails with:

```
error TS2307: Cannot find module 'react' or its corresponding type declarations.
```

**Workaround:**
```yaml
packageExtensions:
  "@xyflow/react@12":
    peerDependencies:
      "@types/react": "*"
```

**Suggested fix:** Add `@types/react` to `peerDependencies` in `@xyflow/react`'s `package.json`.

nielskaspers:
## Summary
Adds `@types/react` and `@types/react-dom` as optional peer dependencies of `@xyflow/react` so TypeScript can resolve the shipped `.d.ts` types under pnpm strict mode (`hoist: false`).

## Issue
Fixes #5738

## Changes
- `packages/react/package.json`: declare `@types/react` / `@types/react-dom` in `peerDependencies`, marked optional in `peerDependenciesMeta` so JavaScript consumers are not asked to install them.
- Added changeset (patch).

## Why optional
`@types/react` is only needed when the consumer is using TypeScript. Marking the peer deps as optional is the standard pattern across the ecosystem — strict pnpm installs resolve the types correctly, and plain JS consumers don't get a missing peer warning.

## Testing
- `.d.ts` resolution works in a pnpm project configured with `hoist: false` (the reported failure mode).
- No runtime code changes — this is purely a `package.json` metadata fix.

### DIFF
--- a/.changeset/react-types-peer-dep.md
+++ b/.changeset/react-types-peer-dep.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
+Add `@types/react` and `@types/react-dom` as optional peer dependencies so the package resolves React types under pnpm strict mode (`hoist: false`) without forcing JavaScript consumers to install the type packages.
--- a/.changeset/react-types-peer-dep.md
+++ b/.changeset/react-types-peer-dep.md
@@ -2,4 +2,4 @@
 '@xyflow/react': patch
 ---
 
-Add `@types/react` and `@types/react-dom` as optional peer dependencies so the package resolves React types under pnpm strict mode (`hoist: false`) without forcing JavaScript consumers to install the type packages.
+Add `@types/react` and `@types/react-dom` as optional peer dependencies to prevent issues with pnpm strict mode (`hoist: false`)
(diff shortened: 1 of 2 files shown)


peterkogo:
Thanks for the contribution! I can confirm that this is the best solution - several other React packages have taken the same route.

moklick:
Could you check if can remove the deps from devDependencies if it's a peer dep @peterkogo ?

peterkogo:
@moklick yes they are still needed

baraknaveh:
Thank you for the fix!

PR #5763, opened 2026-04-29
Feat/example apps

moklick:


moklick:
Let's not add this to this repo, but in a different one like xybench for testing against different branches, versions, local etc

PR #5772, opened 2026-05-07
fix: prevent infinite render loop by moving StoreUpdater component af…

### LINKED ISSUE #5760: Suspending inside of ReactFlow causes infinite render loop

- React Flow version: 12.10.2 (The bug does not appear on 12.10.1)!
- Browser and version: Firefox 150.0 (64-Bit)
- OS and version: Win10 Pro, 22H2, Build 19045.7184

### Live code example

_No response_

### Describe the Bug

When we have one of the following component trees:
```
<Suspense>
  <ReactFlowProvider>
    <ReactFlow>
```
or
```
<ReactFlowProvider>
  <Suspense>
    <ReactFlow>
```
and we use a custom component inside for nodes which suspends, an infinite render loop is generated inside and react crashes. This is particularly annoying if you use suspending things inside, like useSuspenseQuery from tanstackQuery.

I attached a minimal working example (mwe.zip). Simply run it and the error will appear on the console. I also included a screenshot of the console output in the zip.

[mwe.zip](https://github.com/user-attachments/files/27047084/mwe.zip)

### Expected behavior

A suspense boundary outside of the `ReactFlow` component should catch promises and replace them with the fallback component. This was the case in version 12.10.1.

### Screenshots or Videos

Image

### Additional context

_No response_

amans2003:
This PR fixes a regression introduced in v12.10.2 where using a suspending hook (like useSuspenseQuery) inside a custom node caused an infinite render loop when the <Suspense> boundary was placed outside of <ReactFlow> or <ReactFlowProvider>.

The fix involves moving the <StoreUpdater> component back to its original position after <GraphView> in the component tree.

Problem
In PR #5733, <StoreUpdater> was moved before <GraphView> and updated to use useIsomorphicLayoutEffect. When a node inside <GraphView> suspends:

React aborts the render of the tree.
If the Suspense boundary is outside the provider, the state (and store) is lost/reset on the next attempt.
Because <StoreUpdater> was before the suspending component, it would trigger state updates that, combined with the layout effect timing, forced React into an infinite loop of attempting to mount, suspending, and resetting.
Solution
By moving <StoreUpdater> back to its position after <GraphView> (as it was in v12.10.1), we ensure that GraphView renders first. This restores the stable behavior where the store is populated at a point that doesn't conflict with the Suspense orchestration.

Changes
Modified: packages/react/src/container/ReactFlow/index.tsx -> Moved <StoreUpdater> below <GraphView>.
Added: packages/react/tests/suspense.test.tsx -> New regression test using @testing-library/react.
Added: examples/react/cypress/components/reactflow/suspense.cy.tsx -> New Cypress component test for Suspense.
Related Issues
Fixes #5760 Reverts parts of #5733

Testing Done
Verified that the infinite loop no longer occurs using the provided reproduction steps.
Added a new Cypress component test that specifically covers the "Suspense outside Provider" scenario.
Added a unit test in the react package to prevent future regressions in mount timing.


moklick:
this is already fixed by #5769 . We decided to keep the previous behaviour. In order to fix this properly we would need to change how the store gets initialized and that's something for the upcoming major versions but not for 12 unfortunately. Thanks for your help!

PR #5777, opened 2026-05-11
style(base.css): add dark mode variables 

### LINKED ISSUE #5773: Base style buttons don't have dark mode
Svelte & React Flow

Image

Image

Possibly move to xyflow repo

printerscanner:
* I removed some dup imports in the examples of the style.css file
* Added some more variables to dark mode to match base styling we have set in light mode.

### DIFF
--- a/.changeset/sweet-mayflies-drop.md
+++ b/.changeset/sweet-mayflies-drop.md
@@ -2,4 +2,4 @@
 "@xyflow/system": patch
 "react-examples": patch
 ---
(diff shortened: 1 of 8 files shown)


moklick:
We should talk about this topic for the next major and might add a new theme. The base style was meant to be something you can import so that it just works. It already has too many styles in my view and it would be better to remove some variables. However this would be a breaking change and therefore it's not something we can do now.

PR #5779, opened 2026-05-12
feat: get connections from handleId without specifying handleType

printerscanner:
Added a `nodeId-handleId` key to `connectionLookup` so that useNodeConnections can now filter by handleId alone.

### DIFF
--- a/.changeset/six-pandas-lay.md
+++ b/.changeset/six-pandas-lay.md
@@ -1,0 +1,7 @@
+---
+"@xyflow/react": patch
+"@xyflow/svelte": patch
+"@xyflow/system": patch
+---
+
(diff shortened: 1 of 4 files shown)


peterkogo:
I think we cannot safely differentiate handles by `nodeId` & `handleId` alone. This is unfortunately a limitation based on the fact that we have just required `handleIds` to be unique per `handleType` & `nodeId`.

Though this solution would actually work as long as the user is using different ids across all handle types...

This actually makes me think... right now there will be collisions with the id key = `${nodeId}-${handleId}` but maybe there is a way to prevent collisions there and then just return whatever is in that map.

So we should still keep the restriction but you should be able to query the connection map just by `nodeId` and `handleId`

PR #5789, opened 2026-05-17
fix(system): cancel dangling connection line when originating touch is lifted while another touch is active

### LINKED ISSUE #5712: Multi-touch can cause dangling connection line

- React Flow version: 12.10.1
- Browser and version: Chrome 145.0, Firefox 147.0
- OS and version: Windows 11 Home

### Live code example

_No response_

### Describe the Bug

When dragging a connection line on a touch screen, if a second finger touches a node outside of the handle, the connection line won't be cancelled if the first finger is lifted from the screen before the second finger. I am adding a video demonstrating this on an example on the React Flow website where the handles are larger, which makes it easier to reproduce.

I am also attaching a screenshot of an error I see in the console when this happens. It does seem to me like this error is not coming directly from @xyflow/react, but rather from one of the dependencies of @xyflow/system, d3-drag. It might not be related, but thought it would be nice to add it.

Could this be related to what the issue was in #5475? 

1. Go to the [React Flow website ](https://reactflow.dev/examples/interaction/touch-device)
2. Start dragging an edge with one finger
3. Touch the center of the node with another finger
4. Lift the first finger (used to drag the connection line)
5. Lift the second finger
6. The connection line will be dangling

### Expected behavior

I would expect the connection line to disappear once the finger dragging the line has been lifted or after all fingers have been lifted.

### Screenshots or Videos

https://github.com/user-attachments/assets/23c4f744-b330-4075-a84f-59742f5f4a59

Image

### Additional context

_No response_

hari-h-k-k:
### Problem
When dragging a connection line on a touch screen, if a second finger touches anywhere on the canvas while the first finger is dragging, lifting the first finger does not cancel the connection. This leaves the connection line dangling on screen until the second finger is also lifted.

### Root cause:
In onPointerUp, the current guard bails out early if the touchId of the lifted finger doesn't match the one that started the connection.This guard was intended to ignore unrelated touches, but it also ignores the originating touch when other fingers are still on the screen. 

### Fix
Check whether the touchend event contains the originating touch (touchId). If it does, always proceed with cleanup ;do not bail out. Only bail out when the lifted finger is a different, unrelated touch.

Fixes #5712

### DIFF
--- a/packages/system/src/xyhandle/XYHandle.ts
+++ b/packages/system/src/xyhandle/XYHandle.ts
@@ -58,6 +58,7 @@ function onPointerDown(
   const handleType = getHandleType(edgeUpdaterType, handleDomNode);
   const containerBounds = domNode?.getBoundingClientRect();
   let connectionStarted = false;
+  const touchId = 'touches' in event ? event.changedTouches[0]?.identifier : undefined;
 
   if (!containerBounds || !handleType) {
     return;
@@ -196,9 +197,12 @@ function onPointerDown(
   }
 
   function onPointerUp(event: MouseEvent | TouchEvent) {
-    // Prevent multi-touch aborting connection
-    if ('touches' in event && event.touches.length > 0) {
-      return;
+    // Only handle the touch that started the connection
+    if ('changedTouches' in event && touchId !== undefined) {
+      const touch = Array.from(event.changedTouches).find((t) => t.identifier === touchId);
+      if (!touch) {
+        return;
+      }
     }
 
     if (connectionStarted) {
--- a/packages/system/src/xyhandle/XYHandle.ts
+++ b/packages/system/src/xyhandle/XYHandle.ts
@@ -197,10 +197,15 @@ function onPointerDown(
   }
 
   function onPointerUp(event: MouseEvent | TouchEvent) {
-    // Only handle the touch that started the connection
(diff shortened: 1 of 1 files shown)


hari-h-k-k:
Hi @moklick , Can we merge this?

moklick:
Hey @hari-h-k-k 

thanks for your PR! I see your point and it seems a bit buggy that the connection like dangles as soon as you lift the first finger. Your PR fixes this, but introduces a new issue. If you tap twice, the connection like gets aborted even if the first finger is still on the screen. If you can fix that issue, we can merge it :)

hari-h-k-k:
Hi @moklick ,
Thanks for the feedback! I wasn't able to reproduce the double-tap issue in my own testing, but I can see how the fix could potentially introduce it due to touch identifier recycling by the browser. I updated the guard to also check touches (currently active touches) in addition to changedTouches; so the connection now only cancels when the originating finger truly leaves the screen. 

Tested by holding Finger 1 on a handle, tapping Finger 2 multiple times rapidly, and confirming the connection line stays active throughout until Finger 1 is lifted.

moklick:
Thanks for the update, but I am still seeing the same behaviour. When I do a connection and then tap multiple times with another finger, the connection gets lost.

PR #5791, opened 2026-05-18
fix: add type error when handleId is used without handleType in useNodeConnections

### LINKED ISSUE #5701: Allow usage of handleId inside of useNodeConnections without specifiying handleType

Hi, thank you so much for this great library, just one small suggestion.

Currently it is not possible to get all connections to or from a handle if you do not know the handleType, since the handleId option is only evaluated if handleType is also specified.

For example this will give you a list of all connection to that node, not just to that handle:

`
 const connections = useNodeConnections({handleId: port.id})
`

I think it would be a better developer experience if you could provide just the handleId.

If this can't be implemented I think at least the docs for useNodeConnections should be updated to clearly state this limitation. I only really understood what they mean after looking at the source code. Something along the lines of "Filter by handle id (this is only needed if the node has multiple handles of the same type). This can only be used together with handleType".

printerscanner:
Peter and I chatted, and we decided the best solution to this would be to keep the functionality of useNodeConnections the same and just add a type error to point users in the right direction.

### DIFF
--- a/.changeset/clean-eyes-sleep.md
+++ b/.changeset/clean-eyes-sleep.md
@@ -1,0 +1,6 @@
+---
+"@xyflow/react": patch
+"@xyflow/svelte": patch
+---
+
+Adds a type error when `handleId` is used without `handleType` in `useNodeConnections`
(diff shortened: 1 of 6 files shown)


moklick:
Good idea to adjust the type here!

PR #5792, opened 2026-05-26
fix(controls): prevent white controls during color mode mismatch

printerscanner:
changes: 
* Use explicit light-mode icon colors instead of inherit
* CSS tweak that fixes the dark mode box shadow clipping.
* Add colorModeSSR prop to React Flow for parity with Svelte when using
colorMode="system".


moklick:
This seems to be a bigger topic. We can already specify which class should get added on the server with the `colorMode` prop but we can't set a default when "system" is being used. It seems `colorModeSSR` could be `colorModeFallback` in this case. 

Maybe it would make more sense here to think about our theming in general and might add another color mode that would allow ancestor class theming for example.

moklick:
this will be fixed in the next major React Flow 13

PR #5793, opened 2026-05-26
fix: svelte flow errors mention react flow

### LINKED ISSUE #5780: Svelte Flow errors mention React Flow
Image

printerscanner:


packages/svelte/src/lib/store/index.ts
```diff
@@ -298,7 +298,7 @@ export function createStore<NodeType extends Node = Node, EdgeType extends Edge
     const edge = store.edgeLookup.get(id);
 
     if (!edge) {
-      console.warn('012', errorMessages['error012'](id));
+      store.onerror('012', errorMessages['error012'](id));
```
moklick: this seems to be an error meant to be used for a node but it's about an edge in this case.

moklick (changes requested):

packages/svelte/src/lib/store/index.ts
printerscanner: Added a new error message and put a PR in the docs to add a new common error.

PR #5796, opened 2026-05-27
fix(system): prevent middle-mouse pan on nodes when panOnDrag excludes button 1

### LINKED ISSUE #5715: pan middlemouse

-     "@xyflow/react": "^12.10.0",
- Browser and version: 
- firefox/ chrome
- OS and version:
- windows11

### Live code example

https://stackblitz.com/edit/j8bdweiy?file=App.tsx

### Describe the Bug

When setting 
 panOnDrag={[ 2]}
you can still middlemouse pan if you start on a node

(i reused one of your example)
switch to selection mode
middlemouse dragging background does nothing
middkemouse dregging a node pans 

### Expected behavior

no panning expected

### Screenshots or Videos

_No response_

### Additional context

_No response_

printerscanner:


moklick:
good catch!

packages/system/src/xypanzoom/filter.ts
```diff
@@ -34,6 +34,8 @@ export function createFilter({
     const isWheelEvent = event.type === 'wheel';
 
     if (
+      isMiddleMousePan(panOnDrag) &&
```
moklick, line 37: now that I look at this again, I am wondering if this condition is more about a forcePan mouse button, because it works different than the other buttons defined in panOnDrag=[].

PR #5802, opened 2026-05-31
feat: add forcePanMouseButtons prop

printerscanner:


moklick:
I think this would be a good solution but unfortunately it doesn't work like this. We can't specify 0 or 2 as force pan mouse buttons that easily, so I would like to postpone this topic.

PR #5804, opened 2026-06-01
Enable all linting rules and fix

peterkogo:


packages/react/src/utils/general.ts
```diff
@@ -55,6 +55,5 @@ export const isEdge = <EdgeType extends Edge = Edge>(element: unknown): element
 export function fixedForwardRef<T, P = {}>(
```
moklick, line 55: we might be able to remove this. but we need to make a decision about React18/19

PR #5806, opened 2026-06-04
perf(system): avoid allocating a Rect per node in getNodesInside

AlaricBaraou:
While profiling node-heavy canvases I noticed `getNodesInside` is a source of GC churn while panning. On every call it builds a `paneRect` and then a `Rect` per node (via `nodeToRect`) just to measure overlap — so with `onlyRenderVisibleElements` it allocates one object per node on every pan/zoom frame.

I pulled the overlap math into an allocation-free `getRectsOverlappingArea` that takes scalars, compute the viewport bounds once, and test each node directly against them. `getOverlappingArea` now delegates to the same helper, so its behavior is unchanged.

Quick before/after on a 1000-node graph: GC events dropped from ~9,700 to ~45 and the call got ~35% faster.

### DIFF
--- a/.changeset/cull-no-rect-alloc.md
+++ b/.changeset/cull-no-rect-alloc.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/system': patch
+---
+
+Reduce allocations in `getNodesInside` by testing nodes against the viewport without building a `Rect` per node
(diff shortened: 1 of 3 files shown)


moklick:
This looks great! Thanks for your help again @AlaricBaraou 🙏 We will have a closer look in the upcoming week.

PR #5815, opened 2026-06-16
fix(types): make FinalConnectionState narrowable (#5810)

### LINKED ISSUE #5810: onConnectEnd: FinalConnectionState is not typed correctly
Why are all of the fields in the FinalConnectionState nullable?

0007aadil:
## Fixes #5810

`FinalConnectionState` was defined as `Omit<ConnectionState, 'inProgress'>`. Since `ConnectionState` is a discriminated union (`ConnectionInProgress | NoConnection`), applying `Omit` to the union collapses both arms into a single object type — so every field (e.g. `fromHandle`, `from`, `fromNode`) became nullable, even though they are always non-null while a connection is in progress.

This distributes the `Omit` over each arm so the discriminated union is preserved and consumers can narrow correctly:

```ts
export type FinalConnectionState<NodeType extends InternalNodeBase = InternalNodeBase> =
  | (Omit<ConnectionInProgress<NodeType>, 'inProgress' | 'toPosition'> & { toPosition: Position | null })
  | Omit<NoConnection, 'inProgress'>;
```

`toPosition` is overridden to `Position | null` because the runtime sets it to `null` on connection end when there is no target handle (see `XYHandle.onPointerUp`), so the in-progress arm of the *final* state must allow it.

### Changes
- `@xyflow/system`: distribute `Omit` across the union in `FinalConnectionState`.
- `@xyflow/react` / `@xyflow/svelte`: assert `FinalConnectionState` at the `onClickConnectEnd` callback boundary, where the final state is built imperatively.

### Verification
- `tsc --noEmit` passes for `@xyflow/system` and `@xyflow/react`.
- `svelte-check` passes for `@xyflow/svelte` (0 errors).

Approach approved by @moklick in #5810.

### DIFF
--- a/.changeset/final-connection-state-types.md
+++ b/.changeset/final-connection-state-types.md
@@ -1,0 +1,7 @@
+---
+'@xyflow/system': patch
+'@xyflow/react': patch
+'@xyflow/svelte': patch
+---
+
+Fix `FinalConnectionState` type so it preserves the discriminated union. Previously `Omit` was applied to the `ConnectionState` union, collapsing both arms and making every field nullable. It now distributes over the union, so fields like `fromHandle`, `from` and `fromNode` keep their non-null types in the connection-in-progress case and can be narrowed correctly.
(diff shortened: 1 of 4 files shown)


moklick:
Thanks @0007aadil !

PR #5817, opened 2026-06-17
perf(react): avoid allocating connection state per handle on every store update

AlaricBaraou:
While profiling pan/zoom on a canvas with a lot of handles, I noticed every `<Handle>` re-runs its `connectingSelector` on every store update. While no connection is in progress, each handle allocates a new 7-field object that then gets shallow-compared, for every handle on every frame, even though nothing about the connection changed.

Since this idle state is the same for every handle, I return a shared `idleConnectingState` reference when there is no connection in progress (`!fromHandle && !clickHandle`). `useStore`'s `shallow` can then bail out by reference (`Object.is`) instead of allocating and comparing per handle. The connecting path is untouched so the behavior stays the same. `isPossibleEndHandle: true` is exactly what the selector computed before when `fromHandle` is not set.

Quick before/after on the Stress example (625 nodes / 1250 handles), measuring the synchronous fan-out of a transform `setState`: around 25% faster per update, and 1250 fewer object allocations per frame (around 54 KB, so about 3.2 MB/s of garbage avoided at 60fps). It does not change the React re-renders. Nodes and handles already bail out on pan, this just makes the bail out cheaper.

### DIFF
--- a/.changeset/handle-idle-connection-state.md
+++ b/.changeset/handle-idle-connection-state.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
+Reduce per-handle work on every store update by returning a shared connection state from the `Handle` selector while no connection is in progress, instead of allocating a new object and shallow-comparing it for each handle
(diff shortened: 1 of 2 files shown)


moklick:
good finding 🙏 thanks for the PR!

PR #5818, opened 2026-06-17
perf(react): read shared handle config from context instead of subscribing per handle

AlaricBaraou:
While profiling a canvas with many handles I noticed every store update (pan, zoom, drag, selection) runs a selector once per handle. Each `<Handle>` subscribes to the store for `connectOnClick`, `noPanClassName` and `rfId`, but those are flow level config that is identical for every handle, so with N handles that is N selector runs that all return the same thing and bail out.

I moved these three values into a small context provided once near the top of the flow (in `ReactFlowProvider`, so it covers the nodes and any handle a user renders in `children`). A single subscription reads them (with `shallow`, so the reference stays stable) and `Handle` reads them from context, which removes the per-handle subscriptions from the store fan-out.

The behavior is the same. These values are global, so changing one of them already re-rendered every handle before (they all read it). The context value changes only when one of the three actually changes, so handles still re-render exactly then, and not on pan/zoom/drag. The way you set them is unchanged too, they are still `ReactFlow` props synced into the store by `StoreUpdater`.

Before/after on the Stress example (625 nodes / 1250 handles), measuring the synchronous fan-out of a transform `setState`: around 25% faster per update (698us to 523us here).

### DIFF
--- a/.changeset/handle-config-context.md
+++ b/.changeset/handle-config-context.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
+Provide the shared handle config (`connectOnClick`, `noPanClassName`, `rfId`) through context instead of subscribing to the store in every `Handle`, so a store update no longer runs a selector once per handle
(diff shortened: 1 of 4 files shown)


moklick:
thanks @AlaricBaraou !

PR #5820, opened 2026-06-17
feat(vue): setup vue package

bcakmakoglu:


pnpm-workspace.yaml
```diff
@@ -5,11 +5,25 @@ packages:
   - 'tests/*'
 
 catalog:
+  '@antfu/eslint-config': 8.2.0
```
moklick: @bcakmakoglu do we need all this stuff for the catalog?

moklick (changes requested):

PR #5825, opened 2026-06-22
perf(react): only create an XYDrag instance for draggable nodes

AlaricBaraou:
### What this does

`useDrag` (per node) built an `XYDrag` instance for every node on mount, including
non-draggable ones. The `.update()` that attaches the d3-drag listeners was already gated on
`!disabled`, but the instance itself was always constructed. This gates creation on `!disabled`
too, so a non-draggable node never allocates one.

The win is mount and init cost, not the drag hot path: read-only or mostly-static canvases, and
the non-draggable nodes in any graph, no longer build an `XYDrag` (and attach d3-drag listeners)
they will never use. The saving scales with the number of non-draggable nodes.

The create effect now owns the instance lifecycle: create when draggable, destroy on unmount or
when draggability toggles off. The update effect just reconfigures in place. `update()` re-binds
via `selection.call(drag)`, and d3 replaces the `.drag`-namespaced listeners rather than stacking
them, so the previous per-prop-change destroy and rebuild was redundant.

### No behavior change

Draggable nodes still drag. Non-draggable nodes still don't move and stay click-selectable.
Marquee and box-select (handled in `Pane` / `UserSelection`, not `useDrag`) and selection-box
drag (`NodesSelection`, which already gates on `disabled: !shouldRender`) are unaffected.

### Verification

`tsc` and eslint clean. e2e drag suite green (24/24): `draghandle.cy.ts` covers the
`handleSelector` reconfiguration path; `interaction.cy.ts` covers drag, the draggable toggle,
non-draggable nodes, selection, and connection.

### DIFF
--- a/.changeset/drag-instance-draggable-only.md
+++ b/.changeset/drag-instance-draggable-only.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
+Only create an `XYDrag` instance for draggable nodes. A non-draggable node no longer allocates one on mount, so static and read-only graphs pay nothing for drag setup (the saving scales with the number of non-draggable nodes). Draggable nodes are unaffected.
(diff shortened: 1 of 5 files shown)


moklick:
Good idea! I also passed the `nodesDraggable` prop from the `ReactFlow` component down to the `NodeWrapper`, so that the initial state is correct when the global prop is being used.

PR #5830, opened 2026-06-22
perf(react): take the renderers and selection off the global store fan-out

AlaricBaraou:

### What this does

The node and edge renderers, and the selection listener, each ran a `useStore` selector that re-ran
on every store emit. The renderers rebuilt and shallow-compared the whole visible id list on every
frame of a node drag (and under `onlyRenderVisibleElements` re-ran the cull each frame);
`SelectionListener` scanned every node and every edge on every emit just to notice that the selection
had not changed.

This routes all three onto coarse notify-only channels, keeping `set()` so every other `useStore`
consumer is unaffected:

1. A `nodesList` and an `edgesList` channel notify only when the id set or order changes. `setNodes`
   notifies `nodesList` when the adopt takes the full-rebuild path (the only path that changes the id
   set/order); a position-only drag takes the incremental path and notifies nothing. `setEdges`
   detects add/remove/reorder in the single pass it already walks for the per-edge channel (edge
   array order is SVG paint order, so a reorder has to repaint). `useVisibleNodeIds` /
   `useVisibleEdgeIds` recompute the id list only on that signal. When culling, they additionally
   subscribe to the store, because the visible set then genuinely depends on the viewport and node
   positions.
2. A `selection` channel notifies only when the selected set changes. `setNodes` / `setEdges` diff
   the incoming array against the store's own `selectedNodeIds` / `selectedEdgeIds` set (so a change
   is caught even though the user passes a fresh array each time). `SelectionListener` recomputes the
   selection only on that signal, and `onSelectionChange` fires exactly when it did before.

No breaking change, no opt-in. The win is biggest on large graphs that are not using
`onlyRenderVisibleElements`.

### React 17 compatibility

`useExternalSnapshot` reads through `useSyncExternalStore` from `use-sync-external-store/shim`, not
React's built-in hook. The built-in ships only from React 18, and `@xyflow/react` keeps `react >=17`
as a peer dependency, so the shim keeps 17 working (native hook on 18, backfill on 17). The
`use-sync-external-store` dependency was already added by the foundation PR.

### Verification

- `tsc` and eslint clean.
- A position-only drag notifies none of the three channels (the renderers and `SelectionListener`
  do not re-run). Adding/removing/reordering a node or edge notifies the matching list channel and
  the renderer mounts/unmounts the element in the DOM. Selecting a node notifies the selection
  channel, fires `onSelectionChange`, and does not notify `nodesList` (the id list is unchanged).
  `reset` notifies all three. Proven against the source-aliased build.
- The edge-channel, culling+subflow, and connection-highlight proofs are unchanged by this PR.
- Wake-completeness: every structural change funnels through `setNodes` (full rebuild) or `setEdges`
  (the add/remove/reorder detector); every selection change funnels through the `setNodes` / `setEdges`
  selection diff; `reset` notifies all three. So nothing the renderers or `SelectionListener` depend
  on can go stale now that they no longer ride the global emit.

### DIFF
--- a/.changeset/edge-off-fanout.md
+++ b/.changeset/edge-off-fanout.md
@@ -1,0 +1,11 @@
+---
+'@xyflow/react': patch
+---
+
+Take `EdgeWrapper` off the global store. Each edge now reads its data through a per-edge channel
+and gets its static config (`defaultEdgeOptions`, `connectionMode`, `elevateEdgesOnSelect`,
+`zIndexMode`) from a single provider subscription, so a node drag no longer re-runs a per-edge
+selector for every edge, and changing one edge re-renders only that edge. `setEdges` wakes the
+channel for the edges whose object actually changed.
+
+Behaviour is unchanged: same props, same edge rendering.
--- a/.changeset/granular-node-updates.md
+++ b/.changeset/granular-node-updates.md
@@ -1,0 +1,13 @@
+---
+'@xyflow/react': minor
+'@xyflow/system': patch
+---
+
+Make single-node updates scale to large graphs. Node components subscribe to their own
+node instead of the global store, so changing one node out of thousands no longer re-runs
+every node's selector. Edges re-path through a per-edge channel when an endpoint node
+moves, so a single-node move re-renders only the edges that touch it. `setNodes` also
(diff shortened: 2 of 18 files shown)


peterkogo:
superseded by #6025

PR #5831, opened 2026-06-22
perf(react): take the MiniMap off the global store fan-out

AlaricBaraou:

### What this does

The minimap mirrored the main node renderer's old shape: `MiniMapNodes` read the node id list with a
`useStore` selector, and each minimap node read its node with a per-node `useStore` selector. Both
re-ran on every store emit (each node-drag frame), so a graph with a minimap paid a second per-node
fan-out on top of the main renderer's.

This routes the minimap onto the same channels the main renderer now uses:

1. `MiniMapNodes` reads its node id list from the node-list channel (`useVisibleNodeIds(false)`), so
   it recomputes only on add/remove/reorder, not every emit.
2. Each minimap node subscribes to its own node through the per-node channel (`subscribeNode`) and
   recomputes its `{ x, y, width, height, ... }` snapshot only when that node changes, with the same
   shallow dedup as before.

No breaking change, no opt-in, no visible difference. It removes the minimap's per-node fan-out for
any flow that renders a `<MiniMap />`.

### React 17 compatibility

The per-node read here goes through `useExternalSnapshot`, which reads `useSyncExternalStore` from
`use-sync-external-store/shim` rather than React's built-in hook, so the `react >=17` peer range keeps
working (native hook on 18, backfill on 17). The `use-sync-external-store` dependency was added by the
foundation PR.

### Verification

- `tsc` and eslint clean.
- The minimap renders every node; moving a node updates that node's minimap rect (x went 0 -> 120
  through the per-node channel); selecting it adds the `selected` class to its rect; adding a node
  grows the minimap (through the node-list channel). Proven against the source-aliased build.
- Behavior-equivalent to the old version: `Array.from(nodeLookup.keys())` is the same id list and
  order as the old `nodes.map(id)` (adopt builds the lookup in array order); hidden nodes are still
  included in the list and filtered in the wrapper; the per-node snapshot is unchanged.

### DIFF
--- a/.changeset/edge-off-fanout.md
+++ b/.changeset/edge-off-fanout.md
@@ -1,0 +1,11 @@
+---
+'@xyflow/react': patch
+---
+
+Take `EdgeWrapper` off the global store. Each edge now reads its data through a per-edge channel
+and gets its static config (`defaultEdgeOptions`, `connectionMode`, `elevateEdgesOnSelect`,
+`zIndexMode`) from a single provider subscription, so a node drag no longer re-runs a per-edge
+selector for every edge, and changing one edge re-renders only that edge. `setEdges` wakes the
+channel for the edges whose object actually changed.
+
+Behaviour is unchanged: same props, same edge rendering.
--- a/.changeset/granular-node-updates.md
+++ b/.changeset/granular-node-updates.md
@@ -1,0 +1,13 @@
+---
+'@xyflow/react': minor
+'@xyflow/system': patch
+---
+
+Make single-node updates scale to large graphs. Node components subscribe to their own
+node instead of the global store, so changing one node out of thousands no longer re-runs
+every node's selector. Edges re-path through a per-edge channel when an endpoint node
+moves, so a single-node move re-renders only the edges that touch it. `setNodes` also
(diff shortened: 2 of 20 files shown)


peterkogo:
@AlaricBaraou Thank you so much for this PR! This has made us think quite a bit about the potential performance benefits of having more fine grained updates and especially about rethinking the way we use stores in the library!

I have done quite a bit some work that comes nicely together in #6025 which is very much inspired by your PR! 
Its a mixture of changes on how we initially process our nodes, splitting up the single store into multiple ones, relying more on the compiler instead of Zustand selectors and last but not least implementing a fine-grained subscription system for nodes, edges & connections.

We are still in the process of moving this further along to include selections as well.

I will leave your PRs open for now, but will close it in favor of #6025. Together with a bunch of other performance improvements, it should now be faster than your PR for flows with a big number of nodes  in various cases! (moving many nodes, moving few nodes, panning, making connections etc.)

AlaricBaraou:
thanks for taking a hard look at this and building something more comprehensive. Will keep an eye on it and happy to help benchmark once it's further along!
Feel free to close those whenever.

PR #5832, opened 2026-06-22
feat(react): add opt-in patchNodes for keyed node updates

AlaricBaraou:

### What this does

For high-frequency direct manipulation (live cursors, animation, collaborative drag) the full
`setNodes` round-trip - rebuild the array, run the change pipeline, re-adopt - is more than you need
when you just want to move or update a few known nodes. `patchNodes` is an opt-in fast path for that.

`patchNodes([{ id, ...partial }])` writes the patched nodes straight into the internal node store:
O(changed) plus any children whose absolute position moves, recomputed with the same derivation
`setNodes` uses (`@xyflow/system` now exports `getNodePositionAbsolute` for it). Only the changed
nodes wake, through the per-node channel, and their incident edges re-path. `useInternalNode` now
reads through that channel, so it reflects `patchNodes` writes; pair it (or `useNode`) with
`patchNodes` for reads.

It is a narrow tool and bypasses the rest of the pipeline by design. It does not update the nodes
array or fire `onNodesChange`, does not recompute z-index or selection, does not recompute the
visible set under `onlyRenderVisibleElements`, and does not survive a controlled `nodes` prop (the
next prop-driven `setNodes` reverts the patch). `id` and `parentId` are not patchable. Drive
re-parenting, selection and structural changes through `setNodes`. It is reached through the store
API (`useStoreApi().getState().patchNodes(...)`), deliberately kept off the friendly
`updateNode` / `updateNodeData` surface so its bypass semantics are explicit.

### React 17 compatibility

This builds on the stack's per-entity subscriptions, which read `useSyncExternalStore` from
`use-sync-external-store/shim` rather than React's built-in hook, so the `react >=17` peer range keeps
working (native hook on 18, backfill on 17). The `use-sync-external-store` dependency was added by the
foundation PR; this PR adds no new React-version surface of its own.

### Verification

- `tsc` and eslint clean for both `@xyflow/react` and `@xyflow/system`.
- A keyed patch moves the node in `nodeLookup` and wakes its `useInternalNode` consumer (absolute x
  0 -> 120 -> cascaded 210); a patch to an unrelated node does not wake a sibling's consumer; a parent
  patch cascades absolute position to its children; the user-facing `nodes` array is untouched;
  `useInternalNode` reflects the node from mount (the ref-based snapshot, see review notes). Proven
  against the source-aliased build.
- `getNodePositionAbsolute` is line-for-line equivalent to `createInternalNode` / `calculateChildXYZ`,
  so a patched node lands exactly where `setNodes` would place it.
- The node / edge / minimap / selection / nested / culling / connection proofs are unchanged.

### DIFF
--- a/.changeset/edge-off-fanout.md
+++ b/.changeset/edge-off-fanout.md
@@ -1,0 +1,11 @@
+---
+'@xyflow/react': patch
+---
+
+Take `EdgeWrapper` off the global store. Each edge now reads its data through a per-edge channel
+and gets its static config (`defaultEdgeOptions`, `connectionMode`, `elevateEdgesOnSelect`,
+`zIndexMode`) from a single provider subscription, so a node drag no longer re-runs a per-edge
+selector for every edge, and changing one edge re-renders only that edge. `setEdges` wakes the
+channel for the edges whose object actually changed.
+
+Behaviour is unchanged: same props, same edge rendering.
--- a/.changeset/granular-node-updates.md
+++ b/.changeset/granular-node-updates.md
@@ -1,0 +1,13 @@
+---
+'@xyflow/react': minor
+'@xyflow/system': patch
+---
+
+Make single-node updates scale to large graphs. Node components subscribe to their own
+node instead of the global store, so changing one node out of thousands no longer re-runs
+every node's selector. Edges re-path through a per-edge channel when an endpoint node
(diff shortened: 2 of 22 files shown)


peterkogo:
We will have some other ideas how to improve things in the future, but for now we wont be adding this.

PR #5836, opened 2026-06-23
refactor(vue): consolidate Handle's per-instance reactivity

bcakmakoglu:
## What

Trims the per-instance reactive surface of `@xyflow/vue`'s `<Handle>`. There are ~2 handles per node, so this setup runs a lot at scale.

- Collapses the seven separate connection-state `toRef`s (`connectingFrom`, `connectingTo`, `valid`, `connectionInProcess`, `clickConnectionInProcess`, `isPossibleEndHandle`, `isClickConnecting`) into a single `connectionClasses` computed.
- Drops the redundant `isConnectableStart` / `isConnectableEnd` refs — the props already default to `true`, so handlers and the template can use `connectableStart` / `connectableEnd` directly.

Net: ~13 → ~5 reactive effects per handle, with no behavior change to the emitted classes or connection logic.

## Notes

This is the `@xyflow/vue` port of the equivalent vue-flow change (bcakmakoglu/vue-flow#2164).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/vue/src/components/Handle/Handle.vue
+++ b/packages/vue/src/components/Handle/Handle.vue
@@ -82,7 +82,7 @@ const isHandleConnectable = computed(() => {
 // All connection-driven classes in one computed instead of ~7 separate refs: they all derive from the same
 // global `connection*` store state (so they toggle together during a connection) and are used only in the
 // class binding. Mirrors xyflow/react's `connectingSelector`.
-const connectionClasses = computed(() => {
+const connectionClasses = computed<Record<string, boolean>>((prev) => {
   const fromHandle = store.connectionStartHandle;
   const clickFromHandle = store.connectionClickStartHandle;
   const toHandle = store.connectionEndHandle;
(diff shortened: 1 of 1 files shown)


packages/vue/src/components/Handle/Handle.vue
```diff
@@ -123,6 +79,41 @@ const isHandleConnectable = computed(() => {
...
 
+// All connection-driven classes in one computed instead of ~7 separate refs: they all derive from the same
+// global `connection*` store state (so they toggle together during a connection) and are used only in the
+// class binding. Mirrors xyflow/react's `connectingSelector`.
+const connectionClasses = computed(() => {
+  const fromHandle = store.connectionStartHandle;
+  const clickFromHandle = store.connectionClickStartHandle;
+  const toHandle = store.connectionEndHandle;
+  const handleType = type.value;
+
```
moklick, line 90: could you check if you are able to return a stable ref here like https://github.com/xyflow/xyflow/blob/main/packages/react/src/components/Handle/index.tsx#L68-L70

moklick (changes requested):

PR #5845, opened 2026-06-26
fix(system): guard against missing parent node in updateNodeInternals

### LINKED ISSUE #5835: clampPositionToParent crashes when parent node is absent from nodeLookup

- React Flow / Svelte Flow version: 12.8.4 (@xyflow/system 0.0.68)
- Browser and version: 12.8.4 (@xyflow/system 0.0.68)
- OS and version: Windows / macOS (not OS-specific)

### Live code example

https://stackblitz.com/~/github.com/larsthorup/xyflow-parent-crash-repro?file=src/App.tsx

### Describe the Bug

updateNodeInternals passes nodeLookup.get(node.parentId)! to clampPositionToParent without a null check ([store.ts:455](https://github.com/xyflow/xyflow/blob/main/packages/system/src/utils/store.ts#L455)). When the parent is absent from nodeLookup, getNodeDimensions(undefined) throws:

```
TypeError: Cannot read properties of undefined (reading 'measured')
```

The crash is inside a ResizeObserver callback so React ErrorBoundaries cannot catch it.

1. Open the StackBlitz repro
2. Click "Remove parent → triggers crash"
3. The error appears in the HTML and browser console

### Expected behavior

When nodeLookup.get(node.parentId) returns undefined, the clamping should be skipped rather than crashing. Suggested fix:

```
const parent = nodeLookup.get(node.parentId);
if (parent) {
  positionAbsolute = clampPositionToParent(positionAbsolute, dimensions, parent);
}
```

### Screenshots or Videos

Image

Image

### Additional context

Bug is still present on main branch (0.0.78). The ! non-null assertion on line 455 masks the issue at compile time.

Jian-Zhang08:
## Summary

Fixes #5835.

`updateNodeInternals` crashed with `TypeError: Cannot read properties of undefined (reading 'measured')` when a node's `parentId` referenced a parent absent from `nodeLookup` (e.g. the parent was removed while the child's `ResizeObserver` callback still fired). The crash happens inside a `ResizeObserver` callback, so React error boundaries cannot catch it.

`store.ts` passed `nodeLookup.get(node.parentId)!` (a non-null assertion) into `clampPositionToParent`, which called `getNodeDimensions(undefined)`. This skips clamping when the parent is missing — the fix suggested by the issue author.

## Testing done

- Added Cypress regression test `examples/react/cypress/components/utils/update-node-internals.cy.ts` (asserts no throw when the parent is missing).
- Standalone reproduction against built `@xyflow/system`: `getNodeDimensions(undefined)` throws the exact reported error; the guarded path does not throw.
- `typecheck` clean, `build` success, `lint` no new issues.

## Notes

Live repro in the issue. _AI disclosure: prepared with help of an AI coding agent (Claude); verified locally._

### DIFF
--- a/.changeset/fix-update-node-internals-missing-parent.md
+++ b/.changeset/fix-update-node-internals-missing-parent.md
@@ -2,4 +2,4 @@
 "@xyflow/system": patch
 ---
 
-Fix a crash in `updateNodeInternals` when a node's `parentId` is absent from `nodeLookup` (for example the parent was removed while the child's ResizeObserver callback still fires). The position clamping is now skipped instead of throwing `TypeError: Cannot read properties of undefined (reading 'measured')`. (#5835)
+Check if parent node is in node lookup before clamp the position to prevent throwing an error.
--- a/.changeset/fix-update-node-internals-missing-parent.md
+++ b/.changeset/fix-update-node-internals-missing-parent.md
@@ -1,0 +1,5 @@
+---
+"@xyflow/system": patch
+---
+
+Fix a crash in `updateNodeInternals` when a node's `parentId` is absent from `nodeLookup` (for example the parent was removed while the child's ResizeObserver callback still fires). The position clamping is now skipped instead of throwing `TypeError: Cannot read properties of undefined (reading 'measured')`. (#5835)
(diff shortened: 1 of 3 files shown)


moklick:
Thanks @Jian-Zhang08 !

PR #5849, opened 2026-06-29
fix(system): always pair onPanZoomEnd with onPanZoomStart on pan-on-scroll (#5840)

### LINKED ISSUE #5840: [@xyflow/system] pan-on-scroll emits onPanZoomStart without a matching onPanZoomEnd on a lone wheel tick (isPanScrolling stuck true)
- Library: `@xyflow/system` (latest / `main`) — surfaced via `@xyflow/vue`. The handler lives in `@xyflow/system`, so this affects `@xyflow/react`'s `onMoveStart`/`onMoveEnd` and `@xyflow/svelte` equally.
- Config: `panOnScroll: true`

### Describe the Bug
In `createPanOnScrollHandler` (`packages/system/src/xypanzoom/eventhandler.ts`), the first wheel event sets `isPanScrolling = true` and fires `onPanZoomStart`, but the `onPanZoomEnd` timeout is only scheduled in the **`else`** branch (i.e. on a *subsequent* tick):

```ts
clearTimeout(zoomPanValues.panScrollTimeout);

if (!zoomPanValues.isPanScrolling) {
  zoomPanValues.isPanScrolling = true;
  onPanZoomStart?.(event, nextViewport);      // first tick: start, but no end armed
} else {
  onPanZoom?.(event, nextViewport);
  zoomPanValues.panScrollTimeout = setTimeout(() => {
    onPanZoomEnd?.(event, nextViewport);
    zoomPanValues.isPanScrolling = false;
  }, 150);
}
```

Consequences:
1. A **single, isolated wheel tick** (one mouse-wheel notch, or the only scroll event in a gesture) runs the `if` branch only → `onPanZoomStart` fires, no end timeout is ever armed, and `isPanScrolling` stays `true` indefinitely.
2. Because the flag never resets, the **next** pan-scroll gesture's first tick takes the `else` branch and **skips its own `onPanZoomStart`** — so starts and ends drift permanently out of pairing after a single lone tick.

### Steps to reproduce
1. Render a flow with `panOnScroll` and handlers for `onMoveStart` / `onMoveEnd` (react) — or `onPanZoomStart` / `onPanZoomEnd` at the system level.
2. Dispatch a single `wheel` event over the pane (or do one discrete mouse-wheel notch).
3. `onMoveStart` fires; `onMoveEnd` never does, and `isPanScrolling` remains `true`.

### Expected behavior
Every `onPanZoomStart` is eventually paired with an `onPanZoomEnd`, and `isPanScrolling` resets — even for a single wheel tick.

### Suggested fix
Arm (and re-arm) the debounced end on **every** tick, including the first, instead of only in the `else` branch:

```ts
clearTimeout(zoomPanValues.panScrollTimeout);

if (!zoomPanValues.isPanScrolling) {
  zoomPanValues.isPanScrolling = true;
  onPanZoomStart?.(event, nextViewport);
} else {
  onPanZoom?.(event, nextViewport);
}

// (re)arm on every tick so a lone first tick still ends and isPanScrolling resets
zoomPanValues.panScrollTimeout = setTimeout(() => {
  onPanZoomEnd?.(event, nextViewport);
  zoomPanValues.isPanScrolling = false;
}, 150);
```

Jian-Zhang08:
## Summary

Fixes #5840.

`createPanOnScrollHandler` handles the pan-zoom start/zoom/end callbacks itself (it can't rely on d3-zoom's events for pan-on-scroll). It scheduled the "end" timeout **only in the subsequent-tick (`else`) branch**:

```ts
if (!zoomPanValues.isPanScrolling) {
  zoomPanValues.isPanScrolling = true;
  onPanZoomStart?.(event, nextViewport);
} else {
  onPanZoom?.(event, nextViewport);
  zoomPanValues.panScrollTimeout = setTimeout(() => {   // ← only armed here
    onPanZoomEnd?.(event, nextViewport);
    zoomPanValues.isPanScrolling = false;
  }, 150);
}
```

So a **single isolated wheel tick** (one scroll notch with no follow-up) fires `onPanZoomStart` but never schedules the end timeout. `onPanZoomEnd` is therefore never called and `isPanScrolling` stays `true` forever. The next gesture then takes the `else` branch on its very first tick, skipping its own `onPanZoomStart` — permanently misaligning start/end pairs.

## Fix

Move the end-timeout scheduling out of the `else` branch so it is (re)armed on **every** wheel tick. `clearTimeout` is already called at the top of each tick, so multi-tick gestures still emit a single `onPanZoomEnd` 150 ms after the last tick, while a single tick now also gets its matching end.

## Verification

Driving the real `createPanOnScrollHandler` (mocked `d3Selection`/`d3Zoom`, real timers) with two separate single-tick gestures:

| | event sequence | `isPanScrolling` after |
|---|---|---|
| before | `START zoom END` | (first gesture never ended, second never started) |
| after  | `START END START END` | `false` |

The sequence is wrong before the change and correctly paired after it.

I did not add an automated test: `@xyflow/system` has no unit-test setup for the event handlers, and the existing cypress wheel-event tests are flagged as unreliable in the repo (`view-props.cy.tsx`: "the event is somehow broken… works fine when using the mouse"). Happy to add a test in whatever form you prefer.

### DIFF
--- a/.changeset/pan-on-scroll-end-event.md
+++ b/.changeset/pan-on-scroll-end-event.md
@@ -2,4 +2,4 @@
 "@xyflow/system": patch
 ---
 
-Fix `onPanZoomEnd` not firing (and `onPanZoomStart` being skipped on the next gesture) after a single pan-on-scroll wheel tick
+Always fire `onPanZoomEnd` when `onPanZoomStart` was fired due to a pan-on-scroll wheel tick
--- a/.changeset/pan-on-scroll-end-event.md
+++ b/.changeset/pan-on-scroll-end-event.md
@@ -1,0 +1,5 @@
+---
+"@xyflow/system": patch
+---
+
+Fix `onPanZoomEnd` not firing (and `onPanZoomStart` being skipped on the next gesture) after a single pan-on-scroll wheel tick
--- a/packages/system/src/xypanzoom/eventhandler.ts
+++ b/packages/system/src/xypanzoom/eventhandler.ts
@@ -130,13 +130,20 @@ export function createPanOnScrollHandler({
       onPanZoomStart?.(event, nextViewport);
     } else {
       onPanZoom?.(event, nextViewport);
-
-      zoomPanValues.panScrollTimeout = setTimeout(() => {
-        onPanZoomEnd?.(event, nextViewport);
-
-        zoomPanValues.isPanScrolling = false;
-      }, 150);
     }
+
+    /*
+     * (re)arm the end timeout on every wheel tick, not only on subsequent ones.
+     * Otherwise a single isolated wheel tick fires onPanZoomStart but never
+     * schedules the end, so onPanZoomEnd is never called and isPanScrolling
+     * stays true forever - which also makes the next gesture skip its own
+     * onPanZoomStart. (#5840)
+     */
+    zoomPanValues.panScrollTimeout = setTimeout(() => {
+      onPanZoomEnd?.(event, nextViewport);
+
+      zoomPanValues.isPanScrolling = false;
+    }, 150);
   };
 }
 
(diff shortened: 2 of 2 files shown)


moklick:
Thanks @Jian-Zhang08 and @bcakmakoglu !

PR #5866, opened 2026-07-08
React Flow 13 and Svelte Flow 2

### LINKED ISSUE #5305: Remove UMD builds and add example using ESM-based CDN such as esm.sh

With React 19,
[UMD builds have been removed](https://react.dev/blog/2024/04/25/react-19-upgrade-guide#umd-builds-removed).
The React team now recommends using ESM-based CDNs like
[esm.sh](https://esm.sh).

### LINKED ISSUE #5471: Preserve cursor during node resizing

When using React Flow’s NodeResizer, the cursor changes from an 'arrow' to the default 'drag' cursor when dragging a line or corner to resize a node, which creates a jarring user experience. 
Please see here for reference:

https://github.com/user-attachments/assets/9fff42a1-c978-4b90-b611-b6a159ce4f0b

As a workaround, I detect the target element being resized (line or handle) and its position (top, bottom, left, right) in onResizeStart, then set the pane and node cursor manually to the appropriate resize type (ns-resize, ew-resize, nwse-resize, nesw-resize). The cursor is reset on onResizeEnd. This preserves the correct directional cursor during resizing. It would be great if this behavior could be built-in to improve the overall UI experience. Please see the attached recording for reference.

https://github.com/user-attachments/assets/1f102b4e-b674-44fb-b518-0b7e16037b29

### LINKED ISSUE #5485: Add a `NodeResizeControl` identifier to `NodeDimensionChange`

Right now, `NodeDimensionChange` does only contain the dimensions and direction of the resize.

```ts
export type NodeDimensionChange = {
  id: string;
  type: 'dimensions';
  dimensions?: Dimensions;
  /* if this is true, the node is currently being resized via the NodeResizer */
  resizing?: boolean;
  /* if this is true, we will set width and height of the node and not just the measured dimensions */
  setAttributes?: boolean | 'width' | 'height';
};
```

If the node is currently being resized via the `NodeResizer`, it would be informative to know what handle or line (`NodeResizeControl`) triggered the resize. The default ones are identified by `key` and `position` props This is useful for the Helper Lines refactoring

### LINKED ISSUE #5549: Make ReactFlow's nodes property of type readonly NodeType[]

```
<ReactFlow nodes={nodes}
TS4104: The type readonly Readonly<Node>[] is readonly and cannot be assigned to the mutable type Readonly<Node>[]
```

oh no, cringe!

The same with edges.

### LINKED ISSUE #5685: "TypeError: ... is not a function" error, build conflict with zustand 4.x and 5.x versions in the same repo

- React Flow / Svelte Flow version: 12
- Browser and version: latest
- OS and version: macos

### Live code example

_No response_

### Describe the Bug

FYI got multiple runtime "TypeError: ... is not a function" issues with platejs used with xyflow. main cause is different zustand 4.x and 5.x versions, ended hardforking and patching to resolve

https://github.com/xyflow/xyflow/blob/cf2d4be3f57047ae02d08eed49dc640e89a2ef14/packages/react/package.json#L66

related
https://github.com/xyflow/xyflow/issues/5095 

it would be nice to use the latest zustand with react 19 in the latest versions of xyflow/xyflow

use platejs with zustand

### Expected behavior

no runtime crash

### Screenshots or Videos

Image

### Additional context

_No response_

### LINKED ISSUE #5695: Bug report: the movement of blocks using arrow keys is not working

- React Flow / Svelte Flow version: 12.7.0
- Browser and version:  Chrome 144.0.7559.110
- OS and version: Win10

### Live code example

https://reactflow.dev/examples/nodes/connection-limit

### Describe the Bug

The bug occurs in all React Flow projects and also appears on the official React Flow page. The essence of the bug is as follows: when multiple blocks are selected, they can be moved around the canvas using the arrow keys. However, if the selection is removed from any one block, the movement function via the arrow keys breaks, and the selected blocks can no longer be moved using the keyboard.

1. Go to a project with multiple blocks on the canvas.  
2. Select all the blocks.  
3. Try to move them across the field using the arrow keys and ensure that the movement works.  
4. Deselect one of the blocks.  
5. Try to move them across the field again using the arrow keys and confirm that the movement has stopped working.

### Expected behavior

As a user, I expect that after deselecting one of the blocks, the keyboard buttons for movement will continue to work and that the blocks will be able to move across the canvas using the arrow keys.

### Screenshots or Videos

_No response_

### Additional context

_No response_

### LINKED ISSUE #5734: Breaks when built using Vite 8 / rolldown

- React Flow / Svelte Flow version: React flow 12.10.1
- Browser and version: Chrome
- OS and version: Mac OS latest

### Live code example

_No response_

### Describe the Bug

When we build with Vite 8 we get an error "Dynamic require of "react" is not supported"

After analyzing this deeply we found that there is still a hard CJS dependency coming from zustand v4, specifically these

"use-sync-external-store",
"use-sync-external-store/shim"

We are trying to build ESM only however all downstream apps are not able to run due to this error, it seems the default rolldown polyfill or shim is not able to shim this particular CJS usage. 

One fix may be to migrate to zustand v5. 

We have tested many builds and many components and so far only this xyflow is causing us these issues with Vite 8, so it will be related to the change from rollup to rolldown I suspect and clearly a CJS downstream issue. 

Create a react project
Install latest @xyflow/react
Build using vite 
Inspect the dist output and see the usage of require in there
alternatively try to import and use the component in an ESM only react app, it will show the error

### Expected behavior

No require CJS usage

### Screenshots or Videos

_No response_

### Additional context

_No response_

### LINKED ISSUE #5767: Example apps for testing packages
Create some example apps for testing the packed versions of our packages. Frameworks we could check:

- nextjs
- tanstack start
- astro
- nodejs 
- waku
- sveltekit

I already started with a first draft #5763

### LINKED ISSUE #5807: Helper functions for changing parent node

- Svelte Flow version: 1.6.0
- Browser and version: Applicable to any browser. Webkit, Firefox, Chromium-based
- OS and version: Fedora 44

### Live code example

https://stackblitz.com/edit/svelte-flow-ts-p1v4ceyf?file=src%2FFlow%2Findex.svelte

### Describe the Bug

I would like to implement a "group" block type which automatically moves blocks with it that are visually inside it. For that, I use `parentId`, intersection detection methods, and adjust positions to account relative position when using `parentId`, though `updateNodeData`.

The problem: changing the `position` of the node does not work if set alongside the `parentId`. (This is very visible with the custom node in the stackblitz demo.)

👉 _Not_ changing `parentId` allows the `updateNodeData` method to set new `position`s. Additionally dirty hacks like splitting `updateNodeData`, using `tick()` or two do not work. It looks like there is an event ordering issue in Svelteflow that resets changes to nodes' position with `parentId` _after_ the `onnodedragstop`.

1. Open the demo
2. Drag the group node onto the custom node
3. Observe that the coordinates of the custom node _did not change_ despite the code that readjusts its position to relative coordinates.
4. Dragging the node again makes it jump unpredictably.

### Expected behavior

There should be a way to make dynamic parentId assignment work without special shenanigans and coordinate jumps. Setting parentId and new position must work.

### Screenshots or Videos

(Putting a node into a group / putting a group over nodes produces same results)

https://github.com/user-attachments/assets/922c6feb-7477-40fd-bc14-c7c28df4a369

### Additional context

_No response_

### LINKED ISSUE #5811: onconnectend: to is in screen coordinates
onconnectend should reflect what useConnection returns.

moklick:
# React Flow 13 and Svelte Flow 2

This is the PR for the next major versions of both libraries. We will collect the most important changes in this description, so that you can follow the process. Many updates affect both libraries. These are not migration guides, just WIP lists of the most important changes. 

## React Flow 13

For React Flow this is mostly a maintenance release. One of the driving factors for this was the need to update Zustand to the latest version. 

* [x] Set minimum Zustand version to `5`
* [x] Set minimum React version to `18` (we might even drop 18 this is TBD)
* [x] Use new `NodeChangeset` / `EdgeChangeset` class for `onNodesChange` / `onEdgesChange`  
* [x] rename `useStore` to `useReactFlowStore`
* [x] make "@xyflow/react/style.css" available 
* [x] drop `colorMode` prop but always use system/website mode via CSS only (overwritable via `forceColorMode`)
* [x] lib CSS is on "xyflow" layer
* [x] add hooks like `useNode` / `useEdge` 
* [ ] add a new theme
* [x] add helper for changing a parent node 

## Svelte Flow 2

* [x] change API for hooks that take params (from `$derived(useNodeConnections({ id: 'a' }))` to `useNodeConnections(() => { id : 'a' })`
* [x] drop `colorMode` prop but always use system/website mode via CSS only (overwritable via `forceColorMode`)
* [x] rename `useStore` to `useSvelteFlowStore`
* [x] make "@xyflow/svelte/style.css" available
* [x] lib CSS is on "xyflow" layer
* [x] add `onnodeschange` / `onedgeschange` 
* [x] add hooks like `useNode` / `useEdge` 
* [ ] add a new theme
* [x] add helper for changing a parent node 

## General updates

We wanted to use this release to do some general cleanup and maintenance. 

* Revise testing (WIP)
* Move from custom rollup script to vite library mode + tsc
* Add [publint](https://publint.dev/docs/) checking
* Cleanup eslint config & update to v10

https://github.com/xyflow/internal/issues/198

🤫 Psst — @xyflow/vue is in the works, and it's ready for early testing! [Take a look](https://github.com/xyflow/xyflow/blob/renovate/examples/vue/src/Basic/Basic.vue) and let us know what you think.


MisterCheesy:
Please dont drop React 18 support unless it's an absolute must.

AlaricBaraou:
I went through the `packages/react` and `packages/system` diff with a focus on the pan/zoom path and render behavior. I found a few things, I will post them separately as I finish verifying them. First one is a behavior regression in the zoom filter.

**Wheel zoom, pinch zoom and double click zoom stop working when `panOnDrag` is an array**

In `packages/system/src/xypanzoom/filter.ts` this check lost its `event.type === 'mousedown'` condition:

```diff
- if (Array.isArray(panOnDrag) && !panOnDrag.includes(event.button) && event.type === 'mousedown') {
+ if (Array.isArray(panOnDrag) && isMouseEvent(event) && !panOnDrag.includes(event.button)) {
```

`isMouseEvent` is an `instanceof MouseEvent` check, and wheel and dblclick events are `MouseEvent` instances too, with `button === 0`. d3-zoom runs the filter first in its `wheeled` and `dblclicked` handlers, and the `zoomOnScroll` path delegates to the captured native wheel handler, so the filter applies to all of them.

So with `panOnDrag={[1, 2]}` (or any array that does not include 0):

- `zoomOnScroll` (default true) does nothing
- `zoomOnPinch` via ctrl+wheel / trackpad pinch does nothing when `panOnScroll` is off
- `zoomOnDoubleClick` does nothing
- `panOnScroll` itself keeps working because its wheel handler bypasses the filter, but it breaks while `zoomActivationKeyCode` is held, since that falls back to the filtered zoom handler

I reproduced it in the browser on this branch with a minimal flow:

```tsx
<ReactFlow defaultNodes={nodes} defaultEdges={edges} panOnDrag={[1, 2]} fitView />
```

Scroll, pinch and double click do not zoom. Toggling `panOnDrag` back to `true` restores all three, and the same array config on `main` zooms fine. Right and middle button panning still work, it is only the zoom side that dies.

It looks like the condition was dropped in the lint cleanup commit (af9cf831e) where `isMouseEvent(event)` was added so that `event.button` type checks. It reads as redundant with the new guard but it is not, since the instanceof check also lets wheel and dblclick events through to the button check.

Fix is to keep the type guard and restore the event type condition:

```ts
if (
  Array.isArray(panOnDrag) &&
  isMouseEvent(event) &&
  event.type === 'mousedown' &&
  !panOnDrag.includes(event.button)
) {
  return false;
}
```

AlaricBaraou:
Second finding, this one is a render performance regression in `useConnection`.

**`useConnection` re-renders every subscriber on every pointermove during a connection, regardless of the selector**

In v12 the user selector is composed into the store subscription and the equality function runs on the selector output. So `useConnection((c) => c.inProgress)` only re-renders when that value actually changes.

On this branch the hook subscribes to the whole `s.connection` with `useShallow` and applies the user selector afterwards. `XYHandle.onPointerMove` builds a new connection object with fresh `from`, `to` and `pointer` objects on every pointermove, so the shallow check fails on every move while a connection is in progress. The selector cannot prevent the re-render anymore because it runs after the subscription.

I measured it with a flow where custom nodes call `useConnection((c) => c.inProgress)` and count their own renders, driving a connection drag with a fixed number of pointermoves (dev build, StrictMode, so raw counts are doubled). A node using `useConnection()` without a selector and a node without the hook are included as controls:

| build | pointermoves | selector node | no selector node | no hook node |
| --- | --- | --- | --- | --- |
| this branch | 20 | +46 | +44 | 0 |
| this branch | 40 | +86 | - | 0 |
| v12 | 20 | +8 | +46 | 0 |
| v12 | 40 | +8 | +86 | 0 |

The small constants come from the drag start and end transitions (v12 spends one more render there than this branch, per gesture). The regression is the slope: one render per pointermove per subscriber, which the selector should be able to prevent and cannot anymore. Handle highlighting in custom nodes is a common use of this hook, so on a large flow a connection drag becomes N re-renders per mousemove. The no selector case is fine on both, it re-renders per move because `to` actually changes.

A smaller side effect of the same change: the no-selector form now returns a fresh object on every render, so anything using the result in `useMemo`/`useEffect` deps re-runs every render.

I tested a fix on this branch that restores the v12 numbers exactly, including the boundary renders (+8 constant for the selector case, +46/+86 for the no selector case, matching v12 move for move, no loop warnings). It has two parts.

First, compose the selector back into the subscription like v12 did and diff its output:

```ts
export function useConnection<NodeType extends Node = Node, SelectorReturn = ConnectionState<InternalNode<NodeType>>>(
  connectionSelector?: (connection: ConnectionState<InternalNode<NodeType>>) => SelectorReturn
): SelectorReturn {
  const combinedSelector = useCallback(
    (s: ReactFlowStore) => {
      const connection = (
        s.connection.inProgress
          ? { ...s.connection, to: pointToRendererPoint(s.connection.to, s.transform) }
          : { ...s.connection }
      ) as ConnectionState<InternalNode<NodeType>>;

      return connectionSelector ? connectionSelector(connection) : connection;
    },
    [connectionSelector]
  );

  return useReactFlowStore(useCustomDiff(combinedSelector, shallow)) as SelectorReturn;
}
```

Second, and this part matters, `useCustomDiff` needs to memoize on state and selector identity, like the `use-sync-external-store` with-selector shim did in v12:

```ts
export function useCustomDiff<S, U>(selector: (state: S) => U, compare: (a: U, b: U) => boolean): (state: S) => U {
  const cache = useRef<{ state: S; selector: (state: S) => U; value: U } | null>(null);
  return (state) => {
    const cached = cache.current;

    /*
     * If neither the state nor the selector changed, return the cached value without
     * running the selector again. Repeated getSnapshot calls between store updates then
     * always return the same reference, which keeps selectors that build fresh objects
     * from re-render looping (React compares snapshots with Object.is).
     */
    if (cached && cached.state === state && cached.selector === selector) {
      return cached.value;
    }

    const next = selector(state);
    const value = cached && compare(cached.value, next) ? cached.value : next;
    cache.current = { state, selector, value };

    return value;
  };
}
```

Without the second part the fix render-loops: the selector re-runs on every getSnapshot call, builds a fresh `to` object, `shallow` fails on its identity and React logs "The result of getSnapshot should be cached to avoid an infinite loop". I hit this with a first version of the fix. The current `useCustomDiff` call sites do not hit it today only because their compare functions are fully content based (`areConnectionMapsEqual`, `shallowNodeData`, the minimap rect compare), so this change also makes the helper safe for any compare function, and it saves selector executions on the existing call sites as a bonus.

AlaricBaraou:
Last one from my pass, this is documentation and migration surface rather than a bug in the code itself.

**The `useReactFlowStore` JSDoc still documents the removed equality argument, and following it crashes the app**

The doc comment on `useReactFlowStore` says:

> Passing `zustand/shallow` as the second argument is still supported for backwards compatibility.

but the signature only takes a selector now. The sentence and the parameter removal both come from the same commit (57570a127), so it looks like a leftover from an intermediate state of the migration, maybe from a version that still routed through `zustand/traditional`. The changeset in this PR states the intended behavior ("Use useShallow instead of passing equalityFunction parameter"), and the doc contradicts it. The doc also still refers to the hook as `useStore` in the intro and in the `@example`.

The stale sentence matters because the failure mode is not graceful. TypeScript users get a compile error (TS2554), which is fine. Plain JS users who follow the doc keep passing `shallow` and it is silently ignored. With a selector that returns an object, the component then render-loops: I mounted a component using exactly the documented pattern, `useReactFlowStore((s) => ({ x: s.transform[0], y: s.transform[1], zoom: s.transform[2] }), shallow)`. On v12 it renders once and works. On this branch it rendered 106 times, React logged "The result of getSnapshot should be cached to avoid an infinite loop" and then threw "Maximum update depth exceeded", which unmounted the whole flow. Since this is the upgrade path for every v12 app that uses `useStore` with `shallow` today, I think the migration guide should call it out loudly next to the rename.

Related, from the same commit: the shallow comparison was dropped from `useNodes` and `useEdges` (plain `Object.is` on the array now). Neither the commit message nor a changeset mentions it, so I suspect it was a side effect of the migration rather than a decision. The impact is bigger than it sounds. When the nodes array is replaced by a new array containing the same node references, every component using `useNodes` re-renders, and each one receives a new array reference, so `useMemo`/`useEffect`/`memo` children keyed on the array invalidate below it as well. v12 returned the previous array reference in that case, which suppressed the re-render and kept downstream references stable. I measured one re-render per `setNodes([...nodes])` call per consumer on this branch and zero on v12.

This shape is exactly what external state integrations produce. Redux selectors and collaborative stores like Liveblocks re-derive the nodes array on every store update, and the more an app memoizes per node to keep element references stable, the more of those derives end up as a new array of unchanged nodes when something unrelated updates. Under v12 that case cost nothing. On this branch every `useNodes`/`useEdges` consumer re-renders and re-maps on each of those updates, and for a sidebar or list over a few thousand nodes that is real work on every remote update. So it is worth either restoring the shallow comparison (`useReactFlowStore(useShallow(nodesSelector))` keeps the v12 behavior) or adding a changeset line for it.

PR #5889, opened 2026-07-15
fix(system): clamp extent:'parent' child to an unmeasured parent with an intrinsic size

Jian-Zhang08:
## The bug

`calculateNodePosition` (the drag path) builds the "parent" extent for a child with `extent: 'parent'` from `parentNode.measured` directly:

```ts
// packages/system/src/utils/graph.ts
const parentWidth = parentNode.measured.width;
const parentHeight = parentNode.measured.height;
if (parentWidth && parentHeight) {
  extent = [[parentX, parentY], [parentX + parentWidth, parentY + parentHeight]];
}
```

If the parent declares an **intrinsic size** (`width` / `initialWidth`) but hasn't been measured yet, `measured.width`/`height` are `undefined`, the `if` is skipped, the extent stays undefined, and the child is **not clamped to the parent** — it can be dragged outside of it.

The reconcile path (`clampPositionToParent`, used by `adoptUserNodes` / `updateNodeInternals`) already sizes the parent with `getNodeDimensions(parent)` (the `measured → width → initialWidth` fallback), so the two paths were inconsistent.

## The fix

Use `getNodeDimensions(parentNode)` for the parent extent in `calculateNodePosition` too. `getNodeDimensions` is already imported in this file.

- Parent with `initialWidth`/`width` but no `measured` → **now clamped** (was: child escaped).
- Parent already measured → unchanged.
- Parent with no size at all → still not clamped (can't clamp to an unknown size; `getNodeDimensions` returns `0`, so the `if` stays false).

## Verification

Driving the real `calculateNodePosition` (built `@xyflow/system`) with a child at `extent: 'parent'` dragged to `{ x: 500, y: 500 }`:

| parent | before | after |
|---|---|---|
| `initialWidth: 200, initialHeight: 100`, unmeasured | `{ 500, 500 }` (escaped) | `{ 150, 50 }` (clamped) |
| `width: 200, height: 100`, unmeasured | `{ 500, 500 }` | `{ 150, 50 }` |
| `measured: { 200, 100 }` | `{ 150, 50 }` | `{ 150, 50 }` (unchanged) |
| no size | `{ 500, 500 }` | `{ 500, 500 }` (unchanged) |

Added a regression test (`examples/react/cypress/components/utils/calculate-node-position.cy.ts`) for the intrinsic-but-unmeasured parent case. `tsc --noEmit` and `eslint` pass.

Found by auditing the dimension/clamp helpers — same `measured`-fallback class as #5851.

### DIFF
--- a/.changeset/extent-parent-unmeasured-parent.md
+++ b/.changeset/extent-parent-unmeasured-parent.md
@@ -1,0 +1,5 @@
+---
+"@xyflow/system": patch
+---
+
+Fix a child node with `extent: 'parent'` not being clamped to its parent when the parent declares an intrinsic size (`width`/`initialWidth`) but hasn't been measured yet. `calculateNodePosition` now uses the same `getNodeDimensions` fallback as `clampPositionToParent`.
--- a/.changeset/extent-parent-unmeasured-parent.md
+++ b/.changeset/extent-parent-unmeasured-parent.md
@@ -1,5 +1,7 @@
 ---
+"@xyflow/react": patch
+"@xyflow/svelte": patch
 "@xyflow/system": patch
 ---
 
-Fix a child node with `extent: 'parent'` not being clamped to its parent when the parent declares an intrinsic size (`width`/`initialWidth`) but hasn't been measured yet. `calculateNodePosition` now uses the same `getNodeDimensions` fallback as `clampPositionToParent`.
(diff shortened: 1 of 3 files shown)


peterkogo:
Thanks for the contribution!

PR #5916, opened 2026-07-21
Remove expand parent

peterkogo:


moklick:
I am not sure how to move on with this. One idea was the support of groups with absolute positions and `autoSize` but that is still tbd. I don't want to remove a feature without a proper alternative.

PR #5918, opened 2026-07-21
Fix touch panning when drag selection is enabled

AlaricBaraou:
## Summary

- let the existing touch-pan gesture win over pane selection capture when drag panning is enabled
- preserve explicit selection-key behavior and primary-mouse marquee selection
- keep React and Svelte behavior in sync
- clarify that `panOnDrag` button arrays restrict mouse input, not touch input

## Why

When `selectionOnDrag` is enabled and `panOnDrag` is a mouse-button array such as `[1, 2]`, a primary touch currently starts both viewport panning and a marquee selection. The system filter intentionally treats button arrays as mouse-only for touch panning (see #4229), so the Pane should not capture the same touch as a selection unless the selection key is explicitly active.

This uses the gesture's actual `pointerType`, so hybrid touchscreen computers keep normal mouse and pen behavior.

## Behavior

- touch + enabled `panOnDrag` + no selection key: pan without marquee
- touch + selection key: marquee selection
- primary mouse button excluded from `panOnDrag`: marquee selection
- `panOnDrag={false}` and pen behavior remain unchanged

## Tests

- added a desktop-profile, touch-enabled Playwright regression for the conflict
- verified the regression fails on the pre-fix code with one unexpected selection marquee
- added coverage for selection-key touch behavior and primary-mouse behavior
- React pane suite: 29 passed, 4 skipped
- Svelte pane suite: 29 passed, 4 skipped
- React and Svelte package typechecks, builds, and changed-file lint/format checks pass

Related to #2709.

### DIFF
--- a/.changeset/tame-panes-touch.md
+++ b/.changeset/tame-panes-touch.md
@@ -3,4 +3,4 @@
 '@xyflow/svelte': patch
 ---
 
-Prefer touch panning over drag selection when `selectionOnDrag` is combined with mouse-button-specific `panOnDrag` settings.
+Fix selection box appearing when dragging the pane via touch. Prefer touch panning over drag selection when `selectionOnDrag` is combined with mouse-button-specific `panOnDrag` settings.
--- a/.changeset/tame-panes-touch.md
+++ b/.changeset/tame-panes-touch.md
@@ -1,0 +1,6 @@
+---
+'@xyflow/react': patch
+'@xyflow/svelte': patch
+---
+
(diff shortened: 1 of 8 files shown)


peterkogo:
Just to clarify, I am seeing panning + selection behavior only when zooming via multi-touch with `selectionOnDrag` and `panOnDrag = [1, 2]`. I cannot replicate with just a single finger.

None the less, fix is correct! Thank you for the contribution :)

PR #5920, opened 2026-07-23
Svelte onnodeschange, onedgeschange & new node/edge change class

peterkogo:
closes https://github.com/xyflow/xyflow/issues/5867


packages/svelte/src/lib/changes/apply.ts
```diff
@@ -0,0 +1,124 @@
...
+      break;
+    }
+
+    case 'dimensions': {
+      if (typeof change.dimensions !== 'undefined') {
+        node.measured = {
+          ...change.dimensions
+        };
+
+        if (change.setAttributes) {
```
hayleigh-dot-dev: we could just remove this outer check: 
```ts
if (change.setAttributes)
```
right?

packages/svelte/src/lib/changes/apply.ts
```diff
@@ -0,0 +1,124 @@
...
+        }
+      }
+
+      if (typeof change.resizing === 'boolean') {
+        node.resizing = change.resizing;
+      }
+
+      break;
+    }
+    case 'replace': {
```
hayleigh-dot-dev, line 98: nit but consistent spacing between cases pls

packages/svelte/src/lib/changes/index.ts
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    const index = elementChanges.indexOf(change);
+    if (index < 0) {
+      // TODO: change not found, warn?
+      return;
+    }
+
+    // swap the change with the last one & pop
+    elementChanges[index] = elementChanges[elementChanges.length - 1];
```
hayleigh-dot-dev: uh this feels a little clever? could we `elementChanges.splice` instead?

packages/svelte/src/lib/changes/index.ts
```diff
@@ -0,0 +1,304 @@
...
+  /**
+   * Iterates over all queued changes.
+   *
+   * @example
+   * ```ts
+   * for (const change of changes.all()) {
+   *   console.log(change.type, change.id);
+   * }
+   * ```
+   */
```
hayleigh-dot-dev, line 140: maybe we could just implement `[Symbol.iterator]` on this class, or alternatively just return `this.changes.values()` directly without the generator wrapping: it's already an iterator this is just extra shuffling.

packages/svelte/src/lib/changes/index.ts
```diff
@@ -0,0 +1,304 @@
...
+  /**
+   * Applies the queued changes to an elements array and returns the result.
+   *
+   * @example
+   * ```ts
+   * const nextElements = changes.applyTo(elements);
+   * ```
+   */
+  applyTo(elements: ElementType[]): ElementType[] {
+    return this.applyChanges(elements, this);
```
hayleigh-dot-dev, line 210: i dont have anything to say other than `this.function(this)` feels like a code smell to me

packages/svelte/src/lib/changes/index.ts
peterkogo: array will probably be super short anyway so I you convinced me

packages/svelte/src/lib/changes/index.ts
peterkogo: Yeah I get the smell, too! It is however because this.applyChanges is passed in by the constructor and applyChanges is just a pure function, so `this` has to be the second parameter because the `this` inside the function is something else

packages/svelte/src/lib/changes/index.ts
```diff
@@ -130,22 +125,22 @@ export class ElementChanges<
...
-   * for (const change of changes.all()) {
+   * for (const change of changes) {
    *   console.log(change.type, change.id);
    * }
    * ```
    */
-  *all() {
-    for (const elementChanges of this.changes.values()) {
-      yield* elementChanges;
+  *[Symbol.iterator](): Iterator<ChangeType> {
```
hayleigh-dot-dev, line 141: Because `this.changeMap.values()` already returns an iterator I think we might be able to just return it without the for loop/yield ceremony?

peterkogo: but this.changes.values() returns an iterator over arrays. But I want it to iterate over every change not every array.

PR #5936, opened 2026-08-03
Forward missing handle errors to the current onError callback

### LINKED ISSUE #5932: onError is never invoked for error-008 once an edge's handles measure, because EdgeWrapper's position selector captures a stale handler on the edge's first render, when <ReactFlow> has no ancestor <ReactFlowProvider>

- React Flow / Svelte Flow version: @xyflow/react@12.11.2 (@xyflow/system@0.0.79)
- Browser and version: n/a — reproduces independent of browser (React-level bug)
- OS and version: n/a

### Live code example

https://codesandbox.io/p/sandbox/jovial-waterfall-vcy3wy

### Describe the Bug

When <ReactFlow> is rendered without wrapping it in <ReactFlowProvider>, <Wrapper> creates its own store, seeded with edges before any effect runs. EdgeWrapper mounts and memoizes its handle-position selector on that very first render — before the caller's onError prop has reached the store — so the selector's useCallback closes over the store's default onError (devWarn) instead. On that first render the nodes are not yet measured, so getEdgePosition early-returns null and no error fires yet. The real onError prop reaches the store afterwards, via StoreUpdater's effect, but the selector's useCallback dependency array omits onError, so it never re-memoizes. Once measurement later lands handle bounds and that same stale selector re-runs, it detects the missing handle and calls the captured devWarn — which only logs in a development build (NODE_ENV === "development") — instead of the caller's onError. In a production build the error is therefore silent.
Root cause, verified against the installed dist/esm/index.js bundle: [paste the draft's "Root cause" section, steps 1–4, including the 4 code snippets and file:line references]

1. Render <ReactFlow nodes={nodes} edges={edges} onError={onError} /> with no ancestor <ReactFlowProvider>, where one edge references a handle ID that the actually-measured DOM does not contain (i.e. the node component renders no matching <Handle> for it).
2. React Flow's own measurement path detects the missing handle and would raise error-008, but the caller's onError is never called — only React Flow's own dev-only console warning fires (and only under NODE_ENV === "development").
3. To confirm the fix works: wrap the same tree in a bare <ReactFlowProvider> (no other props). onError now fires as expected, because <Wrapper> finds the existing (empty) store instead of creating a new one, so edges and onError both arrive together via StoreUpdater's effect, and EdgeWrapper's first render already has the caller's handler in the store.

### Expected behavior

As a user, I expected onError to be called with error-008 when an edge references a handle ID the DOM doesn't have — regardless of whether <ReactFlow> has an ancestor <ReactFlowProvider> — but instead the error is silently swallowed (routed to a dev-only console.warn that doesn't fire in production) when there is no ancestor provider.

### Screenshots or Videos

_No response_

### Additional context

Suggested fix: add `onError` to the useCallback dependency array in EdgeWrapper's handle-position selector (or, more robustly, read onError from the store inside the selector rather than closing over the prop directly).

sviat-barbutsa:
## Summary

Forward missing-handle errors to the current `onError` callback after nodes finish measuring.

## Problem

When `<ReactFlow />` creates its own provider, edges are available before `StoreUpdater` installs the caller's `onError` callback. `EdgeWrapper` can memoize its position selector during that first render and keep the default warning callback, so error `008` detected after handle measurement never reaches the caller.

## Changes

- Read `onError` from the current store state whenever the edge-position selector runs.
- Add a Playwright regression for a missing source handle without an external `<ReactFlowProvider />`.
- Add a patch changeset for `@xyflow/react`.

## Testing

- `@xyflow/react` typecheck
- React Chromium edge suite: 17 passed
- Focused ESLint check for `EdgeWrapper`
- Prettier check for all changed files

## Related

Fixes #5932

### DIFF
--- a/.changeset/fresh-edges-report.md
+++ b/.changeset/fresh-edges-report.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
+Forward missing-handle errors to the current `onError` callback after node measurement.
--- a/examples/react/src/generic-tests/edges/missing-handle.ts
+++ b/examples/react/src/generic-tests/edges/missing-handle.ts
@@ -1,0 +1,26 @@
+export default {
+  flowProps: {
+    nodes: [
+      {
+        id: 'source',
+        data: { label: 'Source' },
+        position: { x: 0, y: 0 },
+        type: 'input',
+      },
+      {
+        id: 'target',
+        data: { label: 'Target' },
+        position: { x: 200, y: 0 },
+      },
+    ],
+    edges: [
+      {
+        id: 'missing-handle',
+        source: 'source',
+        sourceHandle: 'missing',
+        target: 'target',
+      },
+    ],
+    onError: (code) => console.info(`react-flow-error:${code}`),
+  },
+} satisfies FlowConfig;
--- a/packages/react/src/components/EdgeWrapper/index.tsx
+++ b/packages/react/src/components/EdgeWrapper/index.tsx
@@ -85,7 +85,7 @@ function EdgeWrapper<EdgeType extends Edge = Edge>({
           sourceHandle: edge.sourceHandle || null,
           targetHandle: edge.targetHandle || null,
           connectionMode: store.connectionMode,
-          onError,
+          onError: store.onError,
         });
 
         const zIndex = getElevatedEdgeZIndex({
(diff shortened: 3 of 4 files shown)


moklick:
Thanks for bringing this up! We forgot to pass the `onError` handler to the dependency array.

PR #5938, opened 2026-08-05
fix(react): correct offset math in Background pattern

Feyyus:
The scaledOffset calculation used an errant "|| 1" fallback (likely copy-pasted from the scaledGap line above, where it makes sense to floor a zero gap). Applied to offset this caused two bugs:

- With the default offset={0}, offset*zoom is 0/falsy, so the fallback kicked in and added an extra +1px on top of the intended half-cell centering (translate(-11,-11) instead of (-10,-10) for gap={20}, zoom=1).
- With any non-zero offset, offset*zoom is truthy, so the "||" short- circuits and the entire "+ patternDimensions/2" centering term is dropped, not just miscalculated (e.g. gap={100}, offset={2} produced translate(-2,-2) instead of the correct (-52,-52)).

Removing the fallback and just adding the two terms fixes both cases.

### DIFF
--- a/.changeset/yellow-bushes-make.md
+++ b/.changeset/yellow-bushes-make.md
@@ -1,0 +1,5 @@
+---
+"@xyflow/react": patch
+---
+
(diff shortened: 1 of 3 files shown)


hayleigh-dot-dev:
Thanks for uconvering this @Feyyus, looks good to me! 💕

PR #5955, opened 2026-08-12
fix(react): dont call useCallback conditionally in MiniMap

Jian-Zhang08:
`MiniMap` calls `useCallback` inside a ternary, so the hook only runs when `onNodeClick` is set:

```tsx
// packages/react/src/additional-components/MiniMap/MiniMap.tsx
const onSvgNodeClick = onNodeClick
  ? useCallback((event: MouseEvent, nodeId: string) => {
      const node: NodeType = store.getState().nodeLookup.get(nodeId)!.internals.userNode;
      onNodeClick(event, node);
    }, [])
  : undefined;
```

That causes two separate problems.

**1. Toggling `onNodeClick` crashes the subtree.** The hook count changes between renders, so React throws. Reproduced by rendering the same component shape with `onNodeClick: undefined` and then with a function:

```
Warning: React has detected a change in the order of Hooks called by MiniMapLike.
   Previous render            Next render
   ------------------------------------------------------
6. undefined                  useCallback
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Error: Rendered more hooks than during the previous render.
```

This hits anything like `<MiniMap onNodeClick={enabled ? handleClick : undefined} />`, or wiring the handler up only after data has loaded.

**2. The handler is pinned to the first render.** With `[]` deps, `onNodeClick` is captured once. Rendering three times with a fresh handler each time and then clicking a minimap node calls the **first** one:

```
click after 3 renders -> ["v1 received node-A"]     // expected ["v3 received node-A"]
```

So any handler closing over props or state silently uses stale values (`setPicked([...picked, node.id])` never accumulates, etc.).

Nothing downstream compensates: `MiniMapNodes` → `NodeComponentWrapper` (memo) → `MiniMapNode` (memo) invokes exactly this function. The `[]` was presumably deliberate to keep the memoized children stable — but it was paired with a direct prop read instead of a ref. The repo's ESLint config doesn't include `eslint-plugin-react-hooks`, so neither `rules-of-hooks` nor `exhaustive-deps` flagged it.

## The fix

Call the hook unconditionally and read the handler through a ref. `onSvgNodeClick` is still `undefined` when no handler is passed.

Verified against the same reproduction:

| check | before | after |
|---|---|---|
| toggle `onNodeClick` undefined → function | `Error: Rendered more hooks…` | no error |
| click after 3 renders with a new handler each time | `v1` (stale) | `v3` (current) |
| callback identity across 3 renders | 1 (stable) | 1 (stable) — memoization preserved |
| `onSvgNodeClick` with no handler passed | `undefined` | `undefined` — unchanged |

So behavior only changes in the two broken cases; the stable identity the memoized nodes rely on is kept.

`tsc --noEmit` and `eslint` pass.

### DIFF
--- a/.changeset/minimap-onnodeclick-conditional-hook.md
+++ b/.changeset/minimap-onnodeclick-conditional-hook.md
@@ -1,0 +1,5 @@
+---
+"@xyflow/react": patch
+---
+
+Fix `MiniMap` calling `useCallback` conditionally for `onNodeClick`, which crashed with "Rendered more hooks than during the previous render" when the prop was toggled on or off, and pinned the handler to its first-render value.
--- a/.changeset/minimap-onnodeclick-conditional-hook.md
+++ b/.changeset/minimap-onnodeclick-conditional-hook.md
@@ -2,4 +2,4 @@
 "@xyflow/react": patch
 ---
 
-Fix `MiniMap` calling `useCallback` conditionally for `onNodeClick`, which crashed with "Rendered more hooks than during the previous render" when the prop was toggled on or off, and pinned the handler to its first-render value.
+Fix `MiniMap` calling `useCallback` conditionally for `onNodeClick`.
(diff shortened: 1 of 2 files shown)


moklick:
Thanks!

PR #5960, opened 2026-08-17
feat(react and svelte): add `deselectOnSelection` prop

### LINKED ISSUE #2782: Starting a new box selection clears previously selected elements
### This could be a feature request, but reporting as a bug to get started

When creating a new box selection, the existing selected elements get deselected. We should be able to persist those selections via prop or some method.
It should be noted that clicking to multi-select does not clear previously selected nodes, but that is more time cumbersome than box selection if a large number of elements are to be selected from different parts of the canvas. 

_No response_

1. Open the canvas
2. Select a few elements
3. Start a new box selection
4. The previously selected elements get deselected

### Expected behavior

I expect previously selected elements to remain selected when starting a new box selection

### Screenshots or Videos

https://user-images.githubusercontent.com/36529897/215408214-bee93f0b-c600-4c86-b1e1-e6031e85c27b.mov

### Platform

- OS: macOS
- Browser: Chrome
- Version: 11.4.0

### Additional context

_No response_

jundejz:
## Summary
Adds a `deselectOnSelection` prop to React and Svelte. If set to `false`, then initiating a box drag keeps already selected elements (e.g. nodes and edges) still selected.

## Changes Made
- Adds a `deselectOnSelection` prop to React and Svelte. If `true` same behaviour as before, so `true` is default. Else, we union the already selected nodes and edges with the ones selected by the box drag.
- Adds E2E tests

## Testing
- ✅ All existing tests pass
- ✅ New `deselect-on-selection.ts` test fixture and `Pane deselectOnSelection` test added
- ✅ Backward compatibility verified

## Example Usage
1. Set `deselectOnSelection=false` on any existing example with multiple nodes
2. Run `pnpm dev`
3. Select a node
4. Box drag around a different node
5. See that the originally selected node is still highlighted

Closes #2782

### DIFF
--- a/.changeset/deselectOnSelection-prop.md
+++ b/.changeset/deselectOnSelection-prop.md
@@ -1,0 +1,6 @@
+---
+"@xyflow/react": minor
+"@xyflow/svelte": minor
+---
+
+Add `deselectOnSelection` prop to `<ReactFlow />` and `<SvelteFlow />`. When set to `false`, starting a box selection no longer clears previously selected elements - previously selected nodes and edges stay selected while the box adds new ones.
--- a/examples/react/src/generic-tests/pane/deselect-on-selection.ts
+++ b/examples/react/src/generic-tests/pane/deselect-on-selection.ts
@@ -1,0 +1,40 @@
+export default {
+  flowProps: {
+    minZoom: 0.25,
+    maxZoom: 4,
+    fitView: true,
+    panOnDrag: false,
+    selectionOnDrag: true,
+    deselectOnSelection: false,
+    nodes: [
+      {
+        id: '1',
+        data: { label: '1' },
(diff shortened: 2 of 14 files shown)


jzhou-sl:
Hi @moklick, would you like me to close this and push the PR to `renovate`, given our discussion on #5975!

moklick:
Hey! I am still not sure if this is a prop we want to add.. however if we add it, we would do it only in the next major. I will come back to you when we decided on this!

jundejz:
> Hey! I am still not sure if this is a prop we want to add.. however if we add it, we would do it only in the next major. I will come back to you when we decided on this!

No worries, thanks for getting back to me! 

I forgot to switch accounts, so sorry for the confusion with comments 🤦

PR #5962, opened 2026-08-17
chore(attr): log

moklick:


packages/react/src/components/Attribution/index.tsx
```diff
@@ -11,7 +12,41 @@ const link = `https://reactflow.dev${
...
+  if (style.opacity === '0') return false;
+
+  const rect = el.getBoundingClientRect();
+  if (rect.width === 0 && rect.height === 0) return false;
+
+  return true;
+}
+
 export function Attribution({ proOptions, position = 'bottom-right' }: AttributionProps) {
+  const warned = useRef(false);
```
peterkogo: is this enough to prevent spamming during hot reload?

PR #5972, opened 2026-08-20
Remove updateAbsolutePositions

peterkogo:
Instead of calling `updateAbsolutePositions` after updateNodeInternal do the work directly inside `updateNodeInternals`.

This lets us get rid of updateAbsolutePosition and we don't need to iterate on the nodeLookup afterwards. Even for singular updates we iterated over the nodeLookup once the internals have been updated. 

Now, if you call `updateNodeInternals` on a single node we just update that node and maybe walk its children.
The only reason we called `updateAbsolutePositions` anyway, was to have an updated `nodeLookup` so that `fitView` can take the correct child positions into account immediately.
 
I also extracted `updateChildXYZ` from `updateChildNode`, since we can skip updating the z-indexes and the `parentLookup` at this stage.

Also this uncovered a potential bug: We update the `parentLookup` before we have calculated the actual child positions. This way the node inside the parentLookup is not the same node as in the nodeLookup.

### Before
image

### After
image

Caveat: It was pretty fast before already, and in reality this functions takes way longer because of `getBoundingClientRect()` which is just mocked in this benchmark

### DIFF
--- a/packages/system/src/utils/store.ts
+++ b/packages/system/src/utils/store.ts
@@ -47,10 +47,6 @@ const defaultOptions = {
   elevateNodesOnSelect: true,
   zIndexMode: 'basic' as ZIndexMode,
   defaults: {},
-};
-
-const adoptUserNodesDefaultOptions = {
-  ...defaultOptions,
   checkEquality: true,
 };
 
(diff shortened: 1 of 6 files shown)


packages/system/src/utils/store.ts
```diff
@@ -479,87 +477,128 @@ export function updateNodeInternals<NodeType extends InternalNodeBase>(
...
+  node: InternalNodeBase<NodeType>,
+  context: UpdateInternalsContext<NodeType>
+) {
+  const { parentLookup, updates } = context;
+  const children = parentLookup.get(node.id);
+  if (children) {
+    for (const child of children.values()) {
+      const childUpdate = updates.get(child.id);
+      if (!childUpdate) {
+        return;
```
bcakmakoglu: Shouldn't this be a `continue` instead since `return` would abort for *all* siblings?

packages/system/src/utils/store.ts
```diff
@@ -479,87 +477,128 @@ export function updateNodeInternals<NodeType extends InternalNodeBase>(
...
+      }
+      updateInternals(child, childUpdate, context);
+      updateChildXYZ(child, node, context);
```
bcakmakoglu: Wouldn't `updateInternals` write a *new* node but then we pass the *old* one to `updateChildXYZ`?

moklick:
Very good idea 🙏

PR #5975, opened 2026-08-21
fix(react): change default for `elevateEdgesOnSelect`

### LINKED ISSUE #3840: Allow Selection of Edges for Updatable Edge when repointing

When multiple edges are pointing to the same node using updatable edges, one  should be able to drag out the edges by selecting on the particular edge.

This feature works, however one can only drag the edges out using `Last In First Out` method, the edge that was connected last to the node is the one that can be dragged out. 

The ability to select a node and make it draggable or the ability to define an order for the updatable edges for repointing is required. 

This example is reproducible on the documentation examples itself and I have also attached a video reproducing it. 

https://github.com/xyflow/xyflow/assets/144752715/dd3517a5-cef0-42d3-b14d-d43d8e0b137a

jundejz:
## Summary

Changes the default for the `elevateEdgesOnSelect` prop default for React to be true. This now lines up with the Svelte behaviour and allows users to select which reconnect edge to drag. Without this, on React we have LIFO behaviour.

## Changes Made

- Changes the default for the `elevateEdgesOnSelect` prop default for React to be true. 
- Adds E2E tests

## Testing

- ✅ All existing tests pass
- ✅ New `reconnect.ts` test fixture and `elevate edges on select for overlapping reconnect anchors` test added

## Example Usage

1. Connect multiple reconnect edges into one handle 
2. Select the edge you want to drag
3. Drag the edge, then this edge should be dragged 

Closes #3840

### DIFF
--- a/.changeset/cozy-wings-slide.md
+++ b/.changeset/cozy-wings-slide.md
@@ -1,0 +1,6 @@
+---
+"@xyflow/react": patch
+---
+
+Minimap still works after ReactFlow remounts
+  
--- a/.changeset/elevate-edges-on-select-default.md
+++ b/.changeset/elevate-edges-on-select-default.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': minor
+---
+
+Elevate selected edges by default (`elevateEdgesOnSelect` now defaults to `true`), enabling reconnection of overlapping edges by selecting them first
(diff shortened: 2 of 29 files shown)


moklick:
When I remember it correctly we set it to false because we didn't want to change the behaviour. We could definitely use this as a new default for the next major.

moklick:
Could you create a new pr against "renovate"?

jundejz:
> Could you create a new pr against "renovate"?

Sure no worries!

PR #5993, opened 2026-08-31
feat(xy-theme.css): add theme file for all libraries

printerscanner:
Adds a `xy-theme.css` for all three libraries that is based off of our xy-theme.css file. 

Removed the themed buttons and inputs from the css file. We will need to re-implement this somewhere on web after merging


packages/react/src/styles/xy-theme.css
```diff
@@ -0,0 +1,12 @@
+/* this gets exported as xy-theme.css: the full default theme on top of style.css */
```
moklick, line 1: Could you name the files xy-theme.css?

moklick (changes requested):

.changeset/shared-theme-stylesheet.md
```diff
@@ -0,0 +1,7 @@
+---
+'@xyflow/react': minor
+'@xyflow/svelte': minor
+'@xyflow/vue': minor
+---
+
+Add an optional `xy-theme.css` stylesheet (`@xyflow/<lib>/xy-theme.css`) that ships the xyflow default theme on top of `style.css`.
```
moklick: No need to say that it's optional

moklick (changes requested):

packages/react/src/styles/xy-theme.css
```diff
@@ -0,0 +1,12 @@
...
+
+@import './style.css';
```
moklick, line 3: Do we really want to build it on top of styles.css? Maybe that's not necessary ?! Not sure

moklick (changes requested):

packages/react/src/styles/xy-theme.css
printerscanner: Right now, we need to import the style.css file for it to render correctly. We could add a few more classes so it doesn't need to, but I'm not sure which approach I prefer

moklick: What happens if you only import base.css?

printerscanner: Yeah if we only imported the base styles we would need to add a few more classes. Looks like specifically handles are inherited from style.css, maybe a few other small things if we look closer. 

Screenshot 2026-09-07 at 10 01 55

PR #5995, opened 2026-09-01
perf(react) simple pubsub system to prevent store fan-out

peterkogo:


peterkogo:
Closing in favor of #6025

Issue #4792, opened 2024-11-07
ERROR: ResizeObserver loop completed with undelivered notifications.


bsvetlik:
### What platform were you using when you found the bug?

- React Flow / Svelte Flow version: @xyflow-12.3.4
- Browser and version: Latest Chrome, Edge, FireFox
- OS and version: Windows 10

### Describe the Bug

When our browser application loads multiple flows simultaneously, we occasionally encounter a Webpack DevServer black screen with the error "ResizeObserver loop completed with undelivered notifications.".
This error seems to be specific to React Flow, as it's the only library in our application that utilizes the ResizeObserver API.
Despite the error, React Flow diagrams are displayed correctly. However, we're concerned that this error might potentially impact the performance or functionality of the diagrams.

### Steps to reproduce the bug or issue

Load multiple Flows simultaneously.

### Expected behavior

No errors appear.

### Screenshots or Videos

![error-ResizeObserver loop completed with undelivered notifications](https://github.com/user-attachments/assets/0cce4916-8958-4605-a1f9-0e0ecb40a1d8)

peterkogo:
I think you can safely ignore the warnings. We add a ResizeObserver for each node and with hot reloading might trigger this warning.

But just to make sure are you accessing `node.measured` inside your node and maybe changing it's dimensions based on it?

bsvetlik:
Good to know we can safely disregard this error!
A few observations:
1. The error seems to occur only after a full page refresh, not during hot reloads.
2. Our code doesn't utilize the `node.measured` property

nullromo:
I can **safely** ignore the warnings, but I cannot **sanely** ignore them. It gets annoying having to close the error every time.

I hit this message when trying to resize a node component based on its children.

```ts
// custom node component
export const MyNode = (props: NodeProps<Node<MyNodeData>>) => {
    // get a reference to the whole diagram
    const flowDiagram = useReactFlow<AnyNode, AnyEdge>();

    // find all children nodes of this node
    const children = flowDiagram.getNodes().filter((node) => {
        return node.parentId === props.id;
    });

    // size the content based on how many children are in it
    return (
        <div style={{ height: `${children.length * 30}px` }}>
            { /* content here */ }
        </div>
    );
};
```

Removing the `height: ...` part of the style makes the error stop occurring. Using a static number like `height: '10px'` also causes no errors.

### Question
Is there a "correct" way to resize a component based on its children?

nullromo:
[...]

Issue #3076, opened 2023-05-20
ResizeObserver loop limit exceeded


RainShas:
### Describe the Bug

import React, { useState } from 'react';
import { Handle } from 'reactflow';
import { Select } from 'antd';
import './index.css';
import { shallow } from 'zustand/shallow';
import useStore from '../../store';

const selector = (state) => ({
  nodes: state.nodes,
  edges: state.edges,
  onNodesChange: state.onNodesChange,
  onEdgesChange: state.onEdgesChange,
  onConnect: state.onConnect,
  onAddNodes: state.onAddNodes
});

export default function Formation (props) {
  const { id, isConnectable = true } = props;
  const { nodes, onAddNodes } = useStore(selector, shallow);
  const [Topoffset, setTopoffset] = useState([]);
  const handleChange = (value) => {
    let lengths = nodes.map(item => item.id == id);
    // nodes嵌套对象
    const AddNode = {
      id: `${new Date().getTime()}`,
      data: {
        label: value
      },
      type: 'input',
      position: {
        x: 15,
        y: lengths.length * 50
      },
      draggable: false,
      parentNode: id,
      extent: 'parent',
      targetPosition: 'top'
    }
    // 为reactflow添加嵌套子节点
    onAddNodes([...nodes, AddNode]);
    // 更新该父节点下的所有子节点，从而根据length属性更新dom的marginTop间距
    setTopoffset(lengths);
  }

  return (
    <div className='FormationNode'>
      <h4>xxx</h4>
      <div className='FormationNodeContent' style={{ marginTop: Topoffset.length * 100 }}>
        <Select
          defaultValue="lucy"
          style={{ width: 120 }}
          onChange={handleChange}
          options={[
            { value: 'jack', label: 'Jack' },
            { value: 'lucy', label: 'Lucy' },
            { value: 'Yiminghe', label: 'yiminghe' },
            { value: 'disabled', label: 'Disabled', disabled: true }
          ]}
          className="nodrag"
        />
        <button onClick={() => { handleChange('sss') }}>asd</button>

      </div>
      <Handle type="target" position="top" isConnectable={isConnectable} />
      <Handle type="source" position="bottom" isConnectable={isConnectable} />
    </div>
  )
}

这个文件是一个自定义节点，我希望在选择选项时添加新的嵌套节点，通过zustand新增node，这成功了，然后通过nodes的长度改变div的外边距头部，但是就在这里动态改变margintop时报错了。我已经被折磨一天了。

 

[...]

PR #5999, opened 2026-09-03
fix(react): defer ResizeObserver store updates

fpaupier:
## Problem

The React node and pane `ResizeObserver` callbacks update the store synchronously. Those updates can cause React to commit layout changes before the current observer delivery cycle finishes, which makes browsers report:

`ResizeObserver loop completed with undelivered notifications.`

The warning has also been reported in #4792 and #3076.

## Fix

- Collect node measurements across observer callbacks and flush one `updateNodeInternals` call in the next animation frame.
- Schedule at most one pane measurement per animation frame.
- Skip pane store writes when the measured dimensions have not changed.
- Cancel pending animation frames during cleanup.

The initial pane measurement and the window resize listener remain synchronous. Public APIs do not change. The animation-frame scheduling follows the existing pattern in `useUpdateNodeInternals`.

## Validation

- `pnpm build`
- `pnpm --filter @xyflow/react typecheck`
- ESLint and Prettier checks on the changed source files
- React Playwright suite in Chromium and Firefox: 97 passed and 2 skipped. The remaining Firefox `autoPanOnNodeDrag` assertion also fails on an unchanged `main` checkout.
- Headed Chromium and Firefox checks with the `NodeResizer` example: resized a selected node, resized the viewport, and used zoom controls with zero console errors.

### AI disclosure

- [x] AI-assisted - Drafted with help of OpenAI Codex (GPT-5)

### DIFF
--- a/.changeset/defer-react-resize-observer-updates.md
+++ b/.changeset/defer-react-resize-observer-updates.md
@@ -1,0 +1,5 @@
+---
+'@xyflow/react': patch
+---
+
+Prevent `ResizeObserver` loop warnings when node or pane dimensions change.
--- a/packages/react/src/container/NodeRenderer/useResizeObserver.ts
+++ b/packages/react/src/container/NodeRenderer/useResizeObserver.ts
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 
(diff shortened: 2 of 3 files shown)


fpaupier:
happy to provide more context / reproduce steps for this one as needed, let me know

PR #6017, opened 2026-09-14
perf(react): Separate stores

peterkogo:
We separate the store into Store, NodesStore, EdgesStore, ConnectionStore and SelectionStore.
Some of the work might be superseded by #6020


packages/react/src/store/index.ts
```diff
@@ -442,16 +412,25 @@ const createStore = ({
...
+        connectionStore.setState(initialState.connectionStore);
+        nodesStore.setState(initialState.nodesStore);
+        edgesStore.setState(initialState.edgesStore);
+        selectionStore.setState(initialState.selectionStore);
+        set(initialState.store);
+      },
     };
   });
 
+  return { store, viewportStore, connectionStore, nodesStore, edgesStore, selectionStore };
```
moklick, line 433: nice! could you rename store to optionsStore or something similar?

moklick (changes requested):

PR #6018, opened 2026-09-14
Fix initial fitView firing early when useUpdateNodeInternals is called on mount

MancTwonk:
I have created this as it was something that happened when we upgraded. But if the intended behaviour of how to use the `fitView` prop on `ReactFlow` within a `ReactFlowProvider` has changed since #4550 and I have simple missed it feel free to close this. I thought I would raise this just in case it was overlooked.

## Problem

This issue for me happened when upgrading from v11 to v12, we wrap our component that has ReactFlow in it with `ReactFlowProvider` and we used `fitView` in order to have the nodes fit when loading our flows. But if you use the `fitView` prop and a custom node calls `updateNodeInternals` when it mounts, the initial fit only covers that one node. The other nodes haven't been measured yet, so they get skipped and end up off screen.

## Our Temp Fix

We have added a temp workaround within our app of
```tsx
...
  const fitted = useRef(false);

  useEffect(() => {
    if (!nodesInitialised || fitted.current) return;
    fitted.current = true;
    if (nodeIdFromQueryString) return;
    flow.fitView();
  }, [nodesInitialised, nodeIdFromQueryString, flow]);

...

<ReactFlow
  fitView
...
```

## Cause

I think the cause is `useUpdateNodeInternals` passes `{ triggerFitView: false }` to the store. That was added in #4550 (12.1.0) so only the ResizeObserver runs the initial fitView, after every node has a size.

#5067 (12.5.0) reworked fitView and moved the initial fit to `fitViewQueued`. The queued fit gets resolved in `updateNodeInternals` (db4c019fe), but that check never looks at `params.triggerFitView`. The store stopped reading the param at all. The type still has it and the hook still passes it, it just doesn't do anything anymore.

So the hook's `requestAnimationFrame` can run before the ResizeObserver has measured the rest of the nodes. It uses up the queued fit, and `getFitViewNodes` leaves out every node without a size.

I wanted to make sure #5067 didn't change how this was meant to work before calling it a bug. I don't think it did. The `fitView` prop isn't deprecated, #5067 doesn't mention `useUpdateNodeInternals`, and the hook still passed `triggerFitView: false` after #5067 and the follow up merge in #5187. Looks like the check just got lost in the refactor.

## Fix

`updateNodeInternals` reads `params` again and only resolves the queued fitView when `triggerFitView` is true. It still defaults to true, so the ResizeObserver path works like it did before.

## Repro

1. Wrap the flow in `ReactFlowProvider`
2. Custom node that calls `updateNodeInternals(id)` in a `useEffect` on mount
3. Another node far away, like `{ x: 2000, y: 1200 }`
4. `<ReactFlow fitView ...>`

Before: zoomed in on the custom node, other node off screen.
After: both nodes fit.

## Tests

Added `tests/playwright/e2e/fit-view.spec.ts` and a generic test page at `generic-tests/fit-view/update-node-internals.ts`. The page has `fitView`, a node that calls `updateNodeInternals` on mount, and a node far away. The test checks both nodes are on screen.

### Why `withProvider`

My first version of the test passed even without the fix. The bug only shows up when the flow is inside a `ReactFlowProvider`. When `ReactFlow` makes its own store, the nodes are already in it on the first render and the ResizeObserver measures them before the hook's `requestAnimationFrame` runs, so the fit comes out fine. With an outer provider, the store starts empty and `StoreUpdater` adds the nodes in an effect after mount. Then the manual update runs first and uses up the fit.

A lot of apps wrap in `ReactFlowProvider` so they can use `useReactFlow`, so that's the setup people will actually hit. The generic test `Flow` had no way to add a provider, so I added an optional `withProvider` to `FlowConfig` in `app.d.ts`. `Flow.tsx` wraps in `ReactFlowProvider` when it's set. Only the new page turns it on, so the other generic tests don't change.

The tests fail on chromium, firefox and webkit without the fix, passes with it. Rest of the React e2e suite still passes.


peterkogo:
Interesting edge case. Why would you call updateNodeInternals right after a node mounts though?

MancTwonk:
> Interesting edge case. Why would you call updateNodeInternals right after a node mounts though?

fair question. Our custom nodes derive there handles from the node data (the ports come from the config which our users edit in a side panel, so handles get added / removed at runtime without the node remounting).

So there is 2 places we have calls to `useEffect(() => updateNodeInternals(id), [id, ports])` to get the handle bounds when that happens. the effect also fires on first render as its keyed on ports. I could add a guard that skips the first one but that felt like something i would be added to workaround the library and something that worked in v11. What also made this appear was we load nodes async and render under the `ReactFlowProvider` so the first node mounts and then calls the `updateNodeInternals` you end up with the screen zoomed into one node instead of it acting like it did before the update which was to fit all the nodes to the screen.

I think the hook passing `triggerFitView: false` and the store igoring it is the actual bug. the fix just puts that check back and keeps the default as true so the ResizeObserver path is unchanged.

i can keep my workaround to this in place if you dont think it should change. But would mean in my case it would be to not use the built in fitView and use `useNodesInitialized` plus a manual `fitView()`

Thanks @peterkogo

PR #6019, opened 2026-09-15
perf(react): React Compiler

peterkogo:
This is just the first pass on enabling the React Compiler and cleaning the component code. Some more work that improves performance is in #6020 and there is probably more subsequent work tbd.

Removing all `useCallback` and `useMemo` degrades performance at this state, so it has been left in.
Should this stack be merged in its current form the rerendering dynamics will change a lot and possible removal should be reinvestigated.


packages/react/src/additional-components/Background/Background.tsx
```diff
@@ -143,4 +143,4 @@ BackgroundComponent.displayName = 'Background';
  * When combining multiple <Background /> components it’s important to give each of them a unique id prop!
  *
  */
-export const Background = memo(BackgroundComponent);
+export const Background = BackgroundComponent;
```
moklick, line 146: let's add the memos for components again to prevent calling the render functions.

moklick (changes requested):

moklick (commented):
Do we need the `useCallback` s used for `useSyncExternalStore` in `useConnectionStateForHandle` and other hooks?

PR #6021, opened 2026-09-15
Fix edges remaining visible when connected nodes are hidden

### LINKED ISSUE #6001: React Flow keeps rendering edges connected to a hidden node (only edge.hidden is checked), unlike Svelte Flow after #5977

- React Flow / Svelte Flow version: `@xyflow/react` 12.11.6
- Browser and version: Chrome 128
- OS and version: any

### Describe the Bug

An edge whose source or target node has `hidden: true` keeps rendering in React Flow. The only thing that suppresses an edge is `edge.hidden`; the hidden state of the nodes it connects is never consulted.

I traced it in current source:

- `packages/react/src/hooks/useVisibleEdgeIds.ts` returns every edge id when `onlyRenderVisibleElements` is off, and when it is on it filters with `sourceNode && targetNode && isEdgeVisible(...)`. Neither branch checks `sourceNode.hidden` / `targetNode.hidden`.
- `packages/react/src/components/EdgeWrapper/index.tsx` bails out with `return null` only for `edge.hidden` or a null position, and the position is null only when the node is missing from `nodeLookup` (`nullPosition`). A hidden node is still in `nodeLookup` with its coordinates (`adoptUserNodes` / `updateNodeInternals` in `packages/system/src/utils/store.ts` keep the entry and only clear `handleBounds`), so `getEdgePosition` resolves and the edge draws to where the hidden node used to be.

Svelte Flow just changed this in the opposite direction. #5977 (merged, `@xyflow/svelte`) made `getLayoutedEdges` in `packages/svelte/src/lib/store/visibleElements.ts` skip an edge when `sourceNode.hidden || targetNode.hidden`, closing [SvelteFlow] #3387. That change did not reach the React package, so the two renderers now disagree on what `node.hidden` means for connected edges.

1. Render a React Flow with nodes `a` and `b` and an edge `a -> b`.
2. Update node `b` to `hidden: true` (leave the edge untouched, no `edge.hidden`).
3. Node `b` disappears, but the `a -> b` edge stays on the canvas as a stub pointing at `b`'s last position.

Same setup in Svelte Flow after #5977: the edge is hidden together with the node.

### Expected behavior

Either React Flow matches the Svelte Flow behavior from #5977 (a hidden source or target node also hides the edge), or, if the difference is intentional and users are expected to set `edge.hidden` themselves, the two packages should be documented as diverging on this point. Right now `node.hidden` silently means different things in the two renderers.

### Additional context

Happy to open a PR adding the `sourceNode.hidden || targetNode.hidden` check to `useVisibleEdgeIds` (and an early `return null` in `EdgeWrapper` for the non-`onlyRenderVisibleElements` path) if you want the behaviors aligned.

naveedsajjdawan:
## Summary

- Hide React Flow edges when their source or target node is hidden
- Apply the behavior with and without `onlyRenderVisibleElements`
- Add a Playwright regression test using the existing Update Node example
- Add a patch changeset for `@xyflow/react`

This aligns React Flow with the Svelte Flow behavior introduced in #5977.

Closes #6001

## Verification

- `pnpm --filter @xyflow/system build`
- `pnpm --filter @xyflow/react build`
- `pnpm --filter @xyflow/react typecheck`
- ESLint passed for the changed React hook
- Prettier passed for all changed files

The Playwright regression test is included for CI. Local browser execution was unavailable because the runner could not download the Chromium binary.


peterkogo:
Not a fix really and we will change the behavior around this in the future so I will close this.

PR #6025, opened 2026-09-17
perf(react): Fine grained reactivity for nodes, edges and connections

peterkogo:
Different approach but same performance optimization as : https://github.com/xyflow/xyflow/pull/5828, https://github.com/xyflow/xyflow/pull/5829, https://github.com/xyflow/xyflow/pull/5830, https://github.com/xyflow/xyflow/pull/5831

closes https://github.com/xyflow/internal/issues/350


packages/react/src/store/pubsub.ts
```diff
@@ -0,0 +1,127 @@
...
+  publishConnection(previous: ConnectionState, connection: ConnectionState) {
+    const notifyHandles = new Set<string>();
+
+    if (!sameHandle(previous.fromHandle, connection.fromHandle)) {
+      this.notifyAll(this.handleListeners);
+      return;
+    }
+
+    // We notify all handles that have matched or will match directly
+    notifyHandles.add(getHandleKey(previous.fromHandle));
```
peterkogo: previous is same as new from handle so we don't need to notify both

packages/react/src/components/Handle/index.tsx
```diff
@@ -111,7 +121,8 @@ function HandleComponent(
     connectionInProcess,
     clickConnectionInProcess,
     valid,
-  } = useConnectionStore(useShallow(connectingSelector(nodeId, handleId, type, connectionMode)));
+  } = selector(connection);
```
moklick, line 124: this looks a bit complicated. I think you can pass the connection to the connectingSelector and drop selector.

moklick (changes requested):

